PR #183, opened 2015-01-21
Add Norwegian Bokmål localization

danmichaelo:
At last I got back to this! I've set up a test installation at http://li148-205.members.linode.com/nb/, and I'm planning to move it to http://data.ub.uio.no .

Btw. I'm curious what your workflow is for updating the message definitions in the .po files. Something like running `xgettext` to extract them and `msgmerge` to merge with the existing files? I think it would be nice to have the steps needed documented somewhere, perhaps in the translation script or the readme. And then in the next round automate it :)


danmichaelo:
Renamed "concept counts" to "terms counts" in the message definitions, since that seems to be more correct. Didn't attempt to update the Finnish translation :)

osma:
Hi @danmichaelo ! Thanks for the translations!

Re: the workflow, it's documented in the wiki: https://github.com/NatLibFi/Skosmos/wiki/Translation

danmichaelo:
ah, didn't see that :)

PR #210, opened 2015-03-03
Rest-controller: Add charset in content-type header

danmichaelo:


osma:
Thanks! And sorry for taking so long, we've had a bit of code churn going on.

PR #226, opened 2015-04-07
Updated nb translation

danmichaelo:
I haven't fully grasped the difference between changeNote and historyNote, but here we go :)

In general, "noter" is used more commonly than "bemerkninger"


osma:
Thanks!

PR #271, opened 2015-08-17
Fix base href if SERVER_NAME is not set

danmichaelo:
If SERVER_NAME is not set, it should generate e.g. "/" rather than "http:///" :)


osma:
Right. Base href is specified to contain an absolute URL, at least in [HTML 4.01](http://www.w3.org/TR/html4/struct/links.html#h-12.4 (not sure about HTML5, browsing the spec I couldn't tell whether it's required to be absolute or not). A relative URL in there would violate the spec and it's not surprising if IE chokes on that.

Of course this is just a default/fallback mechanism. You can always specify the real base URL in configuration.

danmichaelo:
Yeah, probably better to leave it like it is.

PR #369, opened 2015-12-02
Add support for RDF collections

danmichaelo:
## Background

Our vocabulary is postcoordinated, so we have concepts like `Vadefugler : Europa` made up from (any number of) other concepts (`Vadefugler` and `Europa`) having their own URIs. When modelling the complex concepts as RDF, we need to express the relationship to the components in some way. And the component list has to be *ordered*… So I turned to RDF collections – feeling a bit uneasy about it since it involves bnodes, but the nice Turtle syntax calmed me down a bit. I also found that MADS already had a property ready for me to use (`mads:componentList`).

## The issue

Given the following data:

```turtle
realfagstermer:c015805
  mads:componentList ( realfagstermer:c005287 realfagstermer:c030200 ) ;
  skos:prefLabel "Vadefugler : Europa"@nb ;
  a mads:ComplexSubject, skos:Concept .
```

, the Skosmos `/data` method returns:

```turtle
realfagstermer:c015805
  mads:componentList [ ] ;
  skos:prefLabel "Vadefugler : Europa"@nb ;
  a mads:ComplexSubject, skos:Concept .
```

not including the collection, just the head blank node from the list. So I've tried to put together a fix for this. Let me know what you think.

Test: https://skosmos.biblionaut.net/rest/v1/realfagstermer/data?uri=http%3A%2F%2Fdata.ub.uio.no%2Frealfagstermer%2Fc018315&format=text/turtle

### DIFF
--- a/model/Model.php
+++ b/model/Model.php
@@ -12,6 +12,7 @@ EasyRdf_Namespace::set('skosmos', 'http://purl.org/net/skosmos#');
 EasyRdf_Namespace::set('void', 'http://rdfs.org/ns/void#');
 EasyRdf_Namespace::set('skosext', 'http://purl.org/finnonto/schema/skosext#');
 EasyRdf_Namespace::set('isothes', 'http://purl.org/iso25964/skos-thes#');
+EasyRdf_Namespace::set('mads', 'http://www.loc.gov/mads/rdf/v1#');
 
 /**
  * Model provides access to the data.
--- a/model/sparql/GenericSparql.php
+++ b/model/sparql/GenericSparql.php
@@ -288,6 +288,10 @@ CONSTRUCT {
  ?directgroup skos:member ?uri .
  ?parent skos:member ?group .
  ?group skos:prefLabel ?grouplabel .
+ ?b1 rdf:first ?item .
+ ?b1 rdf:rest ?b2 .
+ ?item rdf:type ?it .
+ ?item skos:prefLabel ?il .
  ?group rdf:type ?grouptype . $construct
 } WHERE {
  $gc {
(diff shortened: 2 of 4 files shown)


osma:
Thank you for the PR. Makes sense to me. Basically this is not only about (ordered) collections, but about supporting RDF lists as the object and returning the whole list instead of a blank node.

Any chance of writing a unit test for this?

danmichaelo:
Sure, I'll try to do that. Just need to wrap up what I'm currently working at, so it might take a few days

danmichaelo:
I just amended it to the original commit. Not exactly a *unit* test, but at least a test :)

I also found the need for having the `rdf:type` and `skos:prefLabel` for the list items, so I added that to the query.

And I added the MADS namespace. If you would rather not have this, I can remove it.

One issue is that the JSON-LD serialization doesn't produce `@list`s, but just list up the bnodes with their `rdf:first`, `rdf:next` relations. All the data is there, but it makes it harder to consume. So I tried to do a quick look into the ml/json-ld library. There is a `createListObjects` method in `Processor.php` which `"Reconstruct @list arrays from linked list structures"`, but it's private and seems to only be invoked from the `fromRdf` method... so the only way I've found to produce `@list`s so far is the silly

    $result = \ML\JsonLD\JsonLD::toRdf($result);
    $result = \ML\JsonLD\JsonLD::fromRdf($result);

I'm not very familiar with JSON-LD though. Is this perhaps something that can be fixed with framing?

osma:
Thanks for the test, looks enough like a unit test to me :)

Also +1 for adding the MADS namespace. Do you need more MADS support? For example, it would be possible to add class/property qnames to view/extra-msgids.twig to make them translatable. Then we could have support for showing the MADS construct, with localized names, on the concept pages.

Regarding the JSON-LD serialization, the way we produce it for concepts is already rather silly. We first create an initial serialization with this:
https://github.com/NatLibFi/Skosmos/blob/master/model/Model.php#L193
But then we further compact it in RestController:
https://github.com/NatLibFi/Skosmos/blob/master/controller/RestController.php#L544

In practice, it is serialized, then parsed and compacted, then reserialized. If you can hook into this already convoluted process something that produces proper JSON lists, a PR is more than welcome :)

danmichaelo:
Good, then I'll probably send a PR on JSON lists, and either a PR or an issue on MADS soon :)

PR #388, opened 2015-12-12
Move phpunit to dev dependencies

### LINKED ISSUE #386: New requirement PHP version 5.6?
When I tried to upgrade skosmos from (b3fa430) to the latest version (d300eec), I got a dependency error
```
[root@ite-srv11 skosmos]# composer.phar update --no-dev
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - phpunit/phpunit 5.1.3 requires php >=5.6 -> your PHP version (5.3.3) or "config.platform.php" value does not satisfy that requirement.
    - phpunit/phpunit 5.1.2 requires php >=5.6 -> your PHP version (5.3.3) or "config.platform.php" value does not satisfy that requirement.
    - phpunit/phpunit 5.1.1 requires php >=5.6 -> your PHP version (5.3.3) or "config.platform.php" value does not satisfy that requirement.
    - phpunit/phpunit 5.1.0 requires php >=5.6 -> your PHP version (5.3.3) or "config.platform.php" value does not satisfy that requirement.
    - Installation request for phpunit/phpunit 5.1.* -> satisfiable by phpunit/phpunit[5.1.0, 5.1.1, 5.1.2, 5.1.3].
```
My OS is Centos 5, where PHP 5.3.3 is the most recent version available by the standard repositories (same for Centos 6, sadly).

If phpunit is the only dependency which requires php 5.6, it would be great if that could be changed (or phpunit ommited from the `--no-dev` install).

jneubert:
Should fix #386

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,13 +41,13 @@
     "twitter/typeahead.js": "v0.10.5",
     "willdurand/negotiation": "1.3.*",
     "vakata/jstree": "3.2.*",
-    "phpunit/phpunit": "5.1.*",
     "punic/punic": "1.4.0",
     "ml/json-ld": "1.*",
     "medialize/uri.js": "1.15.1",
     "ext-gettext": "*"
   },
   "require-dev": {
+    "phpunit/phpunit": "5.1.*",
     "umpirsky/twig-gettext-extractor": "1.1.*",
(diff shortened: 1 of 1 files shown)


henriyli:
Good catch. Thanks!

PR #419, opened 2015-12-20
Filter out blank nodes as subject

danmichaelo:
Fixes a side effect of #369 . 

Given

    test:ta124 mads:componentList ( test:ta125 test:ta126 ) .

, when querying for info about `test:ta125` or `test:ta126`, the result would include a blank node `rdf:first` statement:

```
test:ta125
  a mads:Topic, skos:Concept ;
  skos:prefLabel "Vadefugler"@nb .

[] rdf:first test:ta125 .
````

### DIFF
--- a/model/sparql/GenericSparql.php
+++ b/model/sparql/GenericSparql.php
@@ -298,7 +298,10 @@ CONSTRUCT {
  ?group a ?grouptype . $construct
 } WHERE {
  $gc {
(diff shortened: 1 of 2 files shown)


osma:
Thanks for the PR! I will get back to this in January after holidays.

osma:
Merged, thanks!

Issue #567, opened 2016-12-07
empty FROM clause causes errors on Stardog


semanticfire:
My SKOS dataset is stored in the default graph.

I have no skosmos:sparqlGraph defined and this resolution results in NULL

The checks in GenericSparql however do not check for null, hence causing issues

osma:
Thanks for reporting. This probably broke with the FROM refactoring in version 1.8 (#534).

osma:
Fixed by merging #568.

PR #568, opened 2016-12-07
NULL for graphS should work

semanticfire:
Fix not checking for NULL related to issue #567 

NULL is returned for situations where no graphs are configured, this should cause a empty from clause

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,16 +27,31 @@
           "reference": "v0.9.13"
         }
       }
-    } 
+    } ,
+	 "netage-easyrdf": {
+        "type":"package",
+        "package": {
+          "name": "netage/easyrdf",
+          "version":"0.9.1",
+          "source": {
+              "url": "https://github.com/netage/easyrdf.git",
+              "type": "git",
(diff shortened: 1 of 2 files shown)


osma:
Thanks for the PR. The NULL fixes look good. A unit test would be good to have, but with the current test framework that might be a bit difficult since it assumes that vocabularies are loaded into specific named graphs.

Generally we apply fixes on the master branch first, then reapply the most important ones to the most recent maintenance branch. So ideally PRs should target the master branch, but we can sort that out.

Was the "Use our fixed easyrdf" commit also intended for this PR? It also seems to change Twig versions. How is your easyrdf fork different from the original?

semanticfire:
our EasyRDF fork does a couple of things

1. set the accept header for describe to turle or jsonld

osma:
Merged the NULL part (master: e1f1b97702f35329a196398bf1323183e6661332, v1.8-maintenance: 3793486ea81de811b25b690fb108f4a083bce988) but not the composer.json changes as they didn't seem relevant to this PR.

Issue #594, opened 2017-03-03
Improve spam blocking in feedback form


osma:
The feedback form seems to pass through more spam than it used to be. The trick we are using to protect the form (hidden field etc) is not working so well anymore. We should try to come up with better measures. A CAPTCHA would be the ultimate, but I would prefer to avoid that if we can as it's annoying for users.

See [discussion on skosmos-users](https://groups.google.com/d/msg/skosmos-users/velzz0XjinQ/EU0-MQiRDgAJ)

kinow:
I found this field https://github.com/NatLibFi/Skosmos/blob/0b2ba5db4c526b0c921d1f02d39d100043028431/view/feedback.twig#L39

I also prefer honeypots to captchas. The one I've been using in some web sites is this one for Laravel: https://github.com/msurguy/Honeypot

I normally add 4, 5 fields, sometimes required, some times not. Further customization is possible to add randomness to name and value requirements.

Besides checking the fields, the library also discards requests that happen too fast, under a certain threshold that depends on the kind of form.

Perhaps we could try a similar approach?

Hope that helps,
Bruno

osma:
@kinow Thanks a lot for the excellent tips!

osma:
As always, PRs are very welcome...

kinow:
Haven't written any PHP in a while. Sounds like a good exercise (unless someone else beats me :)

osma:
@kinow So, what are your plans? Are you interested in doing this? A contribution would be very welcome! :)

kinow:
@osma, yup. Just releasing some Jenkins stuff, then jump back to Jena and can work on Skosmos in parallel (after all I haven't used it with the latest version of Jena).

osma:
@kinow Great! If you use Skosmos with latest Jena, you should be aware of #562 - some things (concept pages and search result page) may be a bit slow. It's nowhere near as bad now (on the Skosmos `master` i.e. development branch) as it was originally, but still Skosmos is SPARQL queries are not performing as they should. This was apparently caused by changes to ARQ in Jena 3.1.1.

kinow:
Done @osma!

henriyli:
Thank you @kinow!

PR #604, opened 2017-03-25
issue #594 add a honeypot

kinow:
This pull request adds a honey pot to Skosmos. See issue #594 for more. It adds two configurations too. Would be useful for users to later add a section in the web site about these new parameters, and instruct them to tweak the values (adding more parameters, or changing the default time threshold).

The honey pot is based on [msurguy/Honeypot](https://github.com/msurguy/Honeypot) module for Laravel, also licensed under the MIT Licence. The main difference being that instead of using Laravel's Crypt class, it simply base64 the server's current time.

There are two types of honey pots. In the first honey pot, a certain value must be empty. Otherwise the request is considered invalid. The user will receive a message of OK, but no e-mail will be sent (as with the previous trap field).

The second honey pot, called also a honey time, is based on time. This field contains the server base64 encoded time. The default threshold is 5 seconds. If any feedback form is submitted in less than 5 seconds, the request is then discarded.

The reason for using base64 was for simplicity. I think we could also use mcrypt, or some other hash algorithm, but preferred to keep the pull request simple, but it can be changed later (feel free to update this pull request, I'm checking "Allow edits from maintainers").

Simple unit test included. I could not find another test for a controller, so had to include a few libraries under Composer's require-dev entry. One for mocking certain methods (like sendFeedback), and another library for parsing HTML. The feedback page's HTML contain a `footer` tag that PHP's default DOMDocument fails to parse.

Hope that helps.

Cheers
Bruno

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,9 @@
     "phpunit/phpunit": "4.8 - 5.1",
     "umpirsky/twig-gettext-extractor": "1.1.*",
     "phpdocumentor/phpdocumentor": "2.*",
-    "codeclimate/php-test-reporter": "0.3.2"
+    "codeclimate/php-test-reporter": "0.3.2",
+    "symfony/dom-crawler": "3.2.6",
+    "mockery/mockery": "1.0.0-alpha1"
   },
   "autoload": {
     "classmap": ["controller/", "model/", "model/sparql/"]
--- a/config.inc.dist
+++ b/config.inc.dist
@@ -60,3 +60,9 @@ define("CUSTOM_CSS", "resource/css/stylesheet.css");
 
 // whether to display the ui language selection as a dropdown (useful for cases where there are more than 3 languages) 
 define("UI_LANGUAGE_DROPDOWN", FALSE);
+
+// whether to enable the spam honey pot or not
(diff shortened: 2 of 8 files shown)


osma:
Wonderful, thanks @kinow! I'll look at this more closely on Monday. Seems like you've done a lot of work to implement this, even unit tests for the Controller which we previously didn't have at all!

kinow:
Thanks @osma. Take your time. Probably may take some minutes to review it, and there is still the question on the encryption of server time. If we add the symfony-crawler and mockery dev dependencies, we can even cover more of the code in the controllers later :)

osma:
Code looks good! I'm unsure how much this will improve spam blocking compared to the current, simple honeypot implementation, but we'll just have to try it to find out. In any case this is should be easy to extend later on if we want more sophisticated honeypot functionality.

kinow:
I replaced a captcha (not Google's "are you a robot" one) by the honeypot library used as reference, and it reduced the spam (to zero for the past year so far). I think the trick is to use meaninful names, that would suggest that need to be filled, and tweak the minimum time for users to write a message (45 seconds in my case).

PR #618, opened 2017-07-12
[WIP] Add options to query Jena Text endpoints with a collation function

kinow:
We need to wait until the new version of Jena is released for testing and publishing a new version of Skosmos with this feature.

But for now we can start reviewing the code :-) I have 2 tests failing, one regarding time in UTC, and which I assume works fine  in Travis-CI.

The other test that fails, is the new test added in this PR for the order of elements returned using a collation. However, this test passes when I run with `phpunit --filter ...`. Only when executed with the other tests it fail.

I couldn't find the reason tonight, but will revisit this later. Will be necessary to update the shell script that downloads fuseki1, or maybe use the new embedded server instead.

PR #621, opened 2017-08-21
Fix ShellCheck issues, and use Jena 3.4.0 for tests

kinow:
Hello,

This pull request addresses all issues found with `shellcheck` in `init_fuseki.sh`.

Also updates Fuseki1 to Jena 3.4.0 version. Fuseki1 is not being distributed with Jena any more. However, it is still being uploaded to Maven central repository.

Hence this change that grabs the `-distribution.zip` from Maven instead of Apache archives.

Ideally, the best approach would - I guess- update the tests to use Fuseki2, however, I am not sure what would be necessary for that. And I suspect we can address it later.

Cheers
Bruno

### DIFF
--- a/tests/init_fuseki.sh
+++ b/tests/init_fuseki.sh
@@ -1,33 +1,34 @@
+#!/bin/bash
 # Note: This script must be sourced from within bash, e.g. ". init_fuseki.sh"
 
-FUSEKI_VERSION=${FUSEKI_VERSION:-1.3.0}
+FUSEKI_VERSION=${FUSEKI_VERSION:-3.4.0}
 
 if [ "$FUSEKI_VERSION" = "SNAPSHOT" ]; then
 	# find out the latest snapshot version and its download URL by parsing Apache directory listings
 	snapshotdir="https://repository.apache.org/content/repositories/snapshots/org/apache/jena/jena-fuseki1/"
-	latestdir=`wget -q -O- "$snapshotdir" | grep 'a href=' | cut -d '"' -f 2 | grep SNAPSHOT | tail -n 1`
-	FUSEKI_VERSION=`basename $latestdir`
-	fusekiurl=`wget -q -O- "$latestdir" | grep 'a href=' | cut -d '"' -f 2 | grep 'distribution\.tar\.gz$' | tail -n 1`
+	latestdir=$(wget -q -O- "$snapshotdir" | grep 'a href=' | cut -d '"' -f 2 | grep SNAPSHOT | tail -n 1)
+	FUSEKI_VERSION=$(basename "$latestdir")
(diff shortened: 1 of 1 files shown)


osma:
Thanks @kinow, this is excellent! And indeed we need to do this to be able to run tests for #618. Would it make sense to merge this into #618? Or is it better if I just merge this into master now and then you can rebase #618 on top of that?

Re Fuseki2, you're right, ideally we would update the tests to run with that, but if this works, it's fine for now.

kinow:
>Or is it better if I just merge this into master now and then you can rebase #618 on top of that?

I think this would be better.

That will give us an idea of what happens in Travis CI. Then when I rebase the PR I **think** Travis may trigger another build. And if so we can compare the output of the tests.

How does that sound?

>Re Fuseki2, you're right, ideally we would update the tests to run with that, but if this works, it's fine for now.

+1

osma:
Merged, thanks!

kinow:
Thank you! :tada:

osma:
Wait a sec...this doesn't actually run the Travis tests with 3.4.0, since in Travis we override the FUSEKI_VERSION environment variable using `env` settings in `.travis.yml`. I will try to fix that soon in `master`...

osma:
Fixed in b8bd6f7078e0ed1521ff8aa660ea8842c1a492f8, tests are now run with 3.4.0 and 3.5.0-SNAPSHOT

kinow:
>Wait a sec...this doesn't actually run the Travis tests with 3.4.0, since in Travis we override the FUSEKI_VERSION environment variable using env settings in .travis.yml. I will try to fix that soon in master...

Oh, and almost forgot. I think .travis.yml is calling `sh`. Any reason for not using `bash`? Or is it not available in Travis agents?

osma:
I think both should be available, as the Travis environment is a fairly standard Ubuntu. Why does it matter?

kinow:
Some times the syntax for some commands can be different, but mostly when using Mac OS... as we are using Travis Linux nodes (must be either RHEL based or Debian based distro) normally `sh` can either be a link to `bash` or `dash`, etc. So it should be fine I guess.

But things like the `$()` and `[[ test ]]` I suspect are more likely to work on `bash` than POSIX `sh`? https://unix.stackexchange.com/questions/182869/use-bin-sh-or-bin-bash-for-ubuntu-osx-compatibility-and-ease-of-use-posi

But really not important, sorry for digressing :)

osma:
If you want it changed, just make a PR with justifications. In my understanding, both `sh` (nowadays often `dash`) and `bash` are OK - problems start to appear if you try to execute `bash`-specific commands under `sh`, which will work on some systems (that symlink `sh` to `/bin/bash`) and fail on others that use e.g. `dash`. But Ubuntu already defaults to using `dash` for `sh` so these problems would quickly get noticed.

kinow:
Agreed :-) sorry again for digressing. Let's focus on the collation issue now :+1:

PR #637, opened 2017-10-19
Handle domains level for ConceptSchemes

### LINKED ISSUE #634: Display hierarchy of domains and micro-thesaurus
We are working on an improvement in Skosmos to handle data structured in the following way :

```
@prefix skos: <http://www.w3.org/2004/02/skos/core#>
@prefix ex: <http://exemple.fr/>
@prefix dcterms: <http://purl.org/dc/terms/>

ex:thesaurus a skos:ConceptScheme ;
	skos:prefLabel "The Thesaurus"@en .

ex:mt1 a skos:ConceptScheme ;
	skos:prefLabel "Micro-Thesaurus 1"@en ;
	dcterms:subject ex:d1 .

ex:mt2 a skos:ConceptScheme ;
	skos:prefLabel "Micro-Thesaurus 2"@en ;
	dcterms:subject ex:d1 .

ex:mt3 a skos:ConceptScheme ;
	skos:prefLabel "Micro-Thesaurus 3"@en ;
	dcterms:subject ex:d2 .

### Begin Domains

ex:domains a skos:ConceptScheme ;
	skos:prefLabel "Special 'Domains' Concept Scheme"@en .

ex:d1 a skos:Concept ;
	skos:inScheme ex:domains ;
	skos:topConceptOf ex:domains ;
	skos:prefLabel "Domain 1"@en .

ex:d2 a skos:Concept ;
	skos:inScheme ex:domains ;
	skos:topConceptOf ex:domains ;
	skos:prefLabel "Domain 2"@en .

#### End Domains

ex:c1 a skos:Concept ;
	skos:prefLabel "Concept 1"@en ;
	skos:inScheme ex:thesaurus , ex:mt1 ;
	skos:topConceptOf ex:thesaurus , ex:mt1 ;
	skos:narrower ex:c1.1 .

ex:c1.1 a skos:Concept ;
	skos:prefLabel "Concept 1.1"@en ;
	skos:inScheme ex:thesaurus , ex:mt1 ;
	skos:broader ex:c1 .

ex:c2 a skos:Concept ;
	skos:prefLabel "Concept 2"@en ;
	skos:inScheme ex:thesaurus , ex:mt2 ;
	skos:topConceptOf ex:thesaurus , ex:mt2 .

ex:c3 a skos:Concept ;
	skos:prefLabel "Concept 3"@en ;
	skos:inScheme ex:thesaurus , ex:mt3 ;
	skos:topConceptOf ex:thesaurus , ex:mt3 .
```

The "domains" are top-level entries in the thesaurus. They are declared as skos:Concept used to "index" the micro-thesaurus with a dcterms:subject. This is typically how large thesaurus like Eurovoc are structured. See http://eurovoc.europa.eu/drupal/?q=download/subject_oriented&cl=en.
Using skos:Collections to represent this is also possible, but less optimal, mostly because :
1. This ConceptScheme-oriented approach is the one preferred in VocBench, so making Skosmos compatible with this would avoid transforming the data between VB and Skosmos
1. The Collection modelling does not allow to browse the full hierarchy starting from top-collections down to concept hierarchy (the concepts are listed flat under each collections)
1. Micro-thesaurus, as their name suggest, _are_ really small thesaurus, so theoretically best modelled as ConceptSchemes.

So, in order to do this, we will propose the following improvements :
1. Extend the REST API that returns the list of ConceptSchemes to also read optional "dcterms:subject" metadata, and include that in the JSON response;
1. Modify the way the tree is generated to generate top-level nodes for domains under which the micro-thesaurus will be inserted;
1. ConceptSchemes without subject information will be inserted as root of the hierarchy, as usual.

Also, we note that the hierarchy REST API make the assumption that a concept is topConceptOf one ConceptScheme only. We need to extend this to be able to properly open the navigation tree for every ConceptScheme in which a concept is declared. I will open a separate issue for this.

### LINKED ISSUE #636: Make top ConceptSchemes in REST API an Array and not a single value
As described in  #634, the query in `GenericSparql->generateQueryConceptSchemesQuery()` makes the assumption that a concept can have at most only one `skos:topConceptOf`.
This needs to be changed if Skosmos is used to display concepts that do belong to multiple concept schemes at the same time.
Note that this will also impact the API format : "top" will hold an array and not a single value anymore.

tfrancart:
Fix #636 
Fix #634


osma:
Are you finished with this work for now or is it still ongoing? There are 14 commits already in this PR and more seem to be added every day. Can you indicate when you would like it to be reviewed?

It is a little hard to digest this since there are many changes. It would be better to break this up to smaller pieces if possible. But I understand it can be hard to do afterwards. However, at least squashing some of the commits could make this easier to review. Now the commit history is quite messy.

Unit tests are broken so this cannot be merged as is. Existing unit tests should never break (if a test is bad it should be broken or disabled) and new unit tests should be added for new features.

tfrancart:
Sorry Osma if we are messing with git branches. We are trying to clean things up and we will propose you something you can review in a single commit.
Essentially the latest commits were done when we decided to include a new "tops" key in the JSON results instead of modifying the existing "top" key.

osma:
@tfrancart OK I see. I will ignore this PR for now, maybe you can open a new one once you are ready to contribute something.

I would suggest that you try to break it up a little. For example the "tops" key could perhaps be contributed on its own?

tfrancart:
Closing this PR and will open a new one when ready.
(The modifications are not that complicated, you'll see ;-) )

PR #647, opened 2017-10-27
Fix Virtuoso problems by reordering operations on literal values

tfrancart:


osma:
Can you elaborate on what the problem is you're addressing here? Is Virtuoso not honoring `DISTINCT` with plain literals? I find that a little surprising..

The unit tests are erroring. See the Travis build details. So this code isn't currently working on Fuseki and cannot be merged as is.

I believe the second argument to `STRDT` should be a IRI, not a literal. Simplest would be to give it as `xsd:string`. I think EasyRdf already defines the `xsd` prefix so no need to add that.

tfrancart:
I was as surprised as you.

1. Go to http://data.legilux.public.lu/sparql
2. Copy/paste and execute following query :

```
SELECT DISTINCT (substr(ucase(str(?label)), 1, 1) as ?l) FROM <http://data.legilux.public.lu/resource/authority/legal-subject-theme/graph> WHERE {
  ?c skos:prefLabel ?label .
  ?c a ?type
  FILTER(langMatches(lang(?label), 'fr'))
  VALUES (?type) { (<http://www.w3.org/2004/02/skos/core#Concept>) }
}
```
3. Note how some letters are duplicated in the result
4. Copy/paste and execute following query :

```
SELECT DISTINCT (STRDT(substr(ucase(str(?label)), 1, 1), xsd:string) as ?l) FROM <http://data.legilux.public.lu/resource/authority/legal-subject-theme/graph> WHERE {
  ?c skos:prefLabel ?label .
  ?c a ?type
  FILTER(langMatches(lang(?label), 'fr'))
  VALUES (?type) { (<http://www.w3.org/2004/02/skos/core#Concept>) }
}
```

5. Note how the letters are now not duplicated.

tfrancart:
Of course the second argument to STRDT must be an IRI and not a string (but Virtuoso accepted it as a string !). Apologies. Will fix this.

tfrancart:
As a way to improve, could you (roughly) describe what is necessary to run the unit tests ourselves before committing ?

osma:
@tfrancart Sure! It looks like we forgot to document how to run tests. See here: https://github.com/NatLibFi/Skosmos/wiki/Unit-and-integration-tests

The unit tests seem to pass now after your IRI fix.

osma:
Possibly related to this Virtuoso issue https://github.com/openlink/virtuoso-opensource/issues/282 but not sure, since they mention underlying issues in the data

I tried the queries on your endpoint, and the results indeed contain duplicates. To me this looks like a bug in Virtuoso, I can't imagine why it doesn't seem to honor the DISTINCT keyword. Even the result set says `distinct="false"`

osma:
This seems to work as well and is IMO a bit cleaner than forcing the datatype to `xsd:string`:

    (ucase(str(substr(?label, 1, 1))) as ?l)

tfrancart:
OK for this, I have integrated it. Note there is also extra datatypes inserted in the "order by" clause of "generateAlphabeticalListQuery", otherwise Virtuoso does not properly handle the ordering.

osma:
Right. I wonder whether a similar workaround is possible there? For example, switch the nesting order of LCASE() and STR() to see if that helps? I'm a bit hesitant to add extra datatype casts to the queries, because they seem like unnecessary clutter and in the worst case might break some other triple store.

I think it would be good to report these issues back to Virtuoso / OpenLink Software, they look very much like bugs.

tfrancart:
`ORDER BY IF(BOUND(?alabel), LCASE(?alabel), LCASE(?label))` without the STR function seems to work. Probably the "STR" function is confusing Virtuoso. Do you see any potential side effect of removing this function from this ORDER BY clause ?

osma:
The reason for using `STR()` is that it strips the language tags. Literals with different language tags do not have a defined order according to SPARQL 1.1. By removing the language tags, we can ensure a consistent ordering of the label (based on Unicode codepoints) even though it may not match language-specific collation order.

I don't think this particular query is ever used without targeting a specific language, so there should never be a mix of different language tags in the result set. So it may be safe to remove the STR() but I'm not 100% sure. In some other queries it's certainly not safe to remove it.

The `IF(BOUND ...)` clause could be replaced by something like `COALESCE(?alabel, ?label)`.

How about `STR(LCASE(COALESCE(?alabel, ?label)))` ?

tfrancart:
̀ORDER BY STR(LCASE(COALESCE(?alabel, ?label)))` works with Virtuoso and yields the proper result. I have updated the PR accordingly.

osma:
Excellent! Waiting for Travis test results.

osma:
I think the trick to work around the Virtuoso bug is to use `STR()` as the outermost function. So if there are further problems like this, it would probably help.

`COALESCE` itself is just a shortcut that is equivalent to the `IF(BOUND(...)...)` construct. I don't think it makes a difference here.

PR #653, opened 2017-11-06
Ability to show deprecated concepts / skosmos:showDeprecated

tfrancart:
In some situations we would like to show deprecated concepts in the interface :
- in alphabetical listings
- when searching
- when listing concepts in a group

We have added a "skosmos:showDeprecated" configuration property in vocabularies.ttl that drives whether the `FILTER NOT EXISTS { ?s owl:deprecated true }` clause is inserted in the SPARQL queries. The default is always to hide deprecated concepts.

Unit tests are included.


osma:
Thanks for the PR. We will consider this. It needs documentation in the wiki, but that should be easy.

However, the tests fail on Travis with this error:

    Trying to @cover or @use not existing method "VocabularyConfig::showDeprecated".

I think it should be `getShowDeprecated` to match the code.

osma:
Looks good to me. @henriyli want to take a look?

PR #654, opened 2017-11-06
Fixes issue 652

tfrancart:


tfrancart:
Sorry about this unnecessary change (I actually though about passing only the URI as a parameter to the method, but changed my mind half-way through, and didn't realized I left thing in an unstable state). This should be fixed now.

PR #657, opened 2017-11-17
Fixes Issue 639

### LINKED ISSUE #639: The "filter by subvocabulary" dropdown in advanced search screen shows URIs instead of labels

http://vocabs.sparna.fr/skosmos/irstea-conceptSchemes/fr/search?clang=fr&q=

1. Go to provided URL
2. Open the first drop-down "Par sous-vocabulaire" ("By subvocabulary") : the dropdown shows the URI of ConceptSchemes

The dropdown contains the URI of ConceptSchemes, I would expect to see their labels instead.

Firefox

tfrancart:
Fix #639

### DIFF
--- a/view/light.twig
+++ b/view/light.twig
@@ -93,7 +93,7 @@
             <select id="scheme-limit" value="{{ limit-scheme }}" class="form-control multiselect" multiple="multiple">
               <option></option>
               {% for uri,scheme in conceptSchemes %}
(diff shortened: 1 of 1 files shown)


osma:
Thanks for the PR!
Twig has an `elseif` statement (see [docs](https://twig.symfony.com/doc/2.x/tags/if.html)) which would be slightly cleaner, you could perhaps switch to that. Other than that looks good! It's a bit unfortunate that the template has to try so many methods...

osma:
Thanks a lot, PR merged and fix also applied to v1.9-maintenance branch.

PR #662, opened 2017-11-21
2 typos + a minor correction in German translation

hauschke:
1. 2 typos 
2. Changed msgstr "Nicht zu einem speziellen Vokabular" (line 221) a bit to make it better fitting in the German feedback form.

### DIFF
--- a/resource/translations/skosmos_de.po
+++ b/resource/translations/skosmos_de.po
@@ -164,7 +164,7 @@ msgstr "Feedback"
 
 #: /tmp/cache/587c83c09abaa/4a/4a8733799c4f4aa865ece4715881ffcbef92985ed08edc2acd445e7b0c129fab.php:64
 msgid "Feedback has been sent!"
-msgstr "Ihr Feedback wurde versended!"
+msgstr "Ihr Feedback wurde versendet!"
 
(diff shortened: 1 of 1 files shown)


osma:
Thanks for the PR.

However, we use Transifex to manage contributed translations, so the "master data" lives there and is just pulled to this git repository when necessary. Can you please do these changes via Transifex instead? See the [wiki page](https://github.com/NatLibFi/Skosmos/wiki/Translation) for instructions.

hauschke:
Thanks for your pointer. I made these changes (and some others, too) at Transifex.

osma:
This is great! The German translation has languished for quite some time. If you find any other problems, feel free to correct them on Transifex.

I pulled the updated German translations from Transifex into the `master` branch.

hauschke:
If you could help me with a real world example of the following to statements, I could complete the translation:
https://www.transifex.com/national-library-of-finland/skosmos/translate/#de/skosmos_enpo/50899943?translated=no

osma:
Did you look at the translator help texts in Transifex? I've added hints for most messages that have been difficult to translate, but they are not very prominent in the Transifex UI, unfortunately.

PR #663, opened 2017-11-22
Multiple top concepts / dct:subject of concept schemes

### LINKED ISSUE #636: Make top ConceptSchemes in REST API an Array and not a single value

tfrancart:
fix #636


osma:
Thanks for the PR. Looks much cleaner than the previous one :)

tfrancart:
Just a gentle reminder on this PR. @henriyli  were you able to review it ?
(I am asking because I have at least one other PR depending on this one to propose to handle hierarchy properly on the javascript side)
Thanks

osma:
We reviewed this together with @henriyli and decided to merge it, looks good although we don't need this functionality in our own vocabularies

tfrancart:
Thanks - yes, it should be harmless for data that don't use this structure.

Issue #658, opened 2017-11-17
In advanced search screen, when selecting more than 1 subvocabulary, the query fails


tfrancart:
## At which URL did you encounter the problem?

http://finto.fi/yso/en/search?clang=en&q=agriculture&vocabs=

## What steps will reproduce the problem?
1. Go to above URL. Note that the search yields 2 results from YSO vocabulary
2. In the left dropdown "By subvocabulary", select vocabulary "YSO - General Finnish Ontology", and click on "limit search"
3. The list is refreshed
4. Now select a second subvocabulary from the list (http://finto.fi/yso/en/search?clang=en&q=agriculture&vocabs=), and click on "limit search"
5. The query fails and return no results

## What is the expected output? What do you see instead?

Expected output : the same 2 concepts in the search results.
What happens instead : no results found.

The SPARQL query is wrong, the URI of the 2 graphs are not properly concatenated (exemple taken from another Skosmos install with different graph URIs). Note the "+" signs between the 2 URIs :

```
...
  {
   ?s a ?type .
    ?s skos:inScheme <http://data.education.fr/voc/scolomfr/scolomfr-voc-022+http://data.education.fr/voc/scolomfr/scolomfr-voc-023> .
  }`
```

osma:
Thanks for reporting!

osma:
This was fixed by PR #664

PR #664, opened 2017-11-24
Issue 658

tfrancart:
1. decode parameter before exploding it on whitespace
2. use a UNION in SPARQL query for the schemes conditions


osma:
Thanks for the PR.

Would it be possible to have a unit test for the issue (#658) this is fixing?

I'd do the `$schemecond` construction a bit differently - put the individual conditions in an array, then `implode()` the array using ` UNION ` as the glue string. No need for the `substr` trickery then.

tfrancart:
Added unit test. Used implode.

osma:
Excellent! Waiting for Travis and other tests to complete.

osma:
Merged after squashing into a single commit 4b131ad6873c927f1f6acb045db14319df3ea4dc

PR #667, opened 2017-11-30
Issue 666 in a single commit

### LINKED ISSUE #666: Prefer mapping property in the same Vocabulary rather than in other vocabularies

http://erato.sparna.fr/scolomfr/scolomfr-voc-000-3-2/fr/page/scolomfr-voc-016-num-00720

1. Go to the above URL
2. Look at the bottom section in the mapped concepts
3. Note how concepts are found in other vocabularies "ScolomFr 3.0", "ScolomFr 3.1", at random

We have the following situation :

1. We have a SKOS file containing multiple concept schemes. The file contains SKOS mapping properties between concepts of different concept schemes in this file
2. We have different _versions_ of the same SKOS file that we want to publish in Skosmos. We have setup one vocabulary per version (see http://erato.sparna.fr/scolomfr/fr/). The URIs of the Concept in these vocabularies are the same.

So, when displaying the mapping properties of a Concept, I expect to see links to concepts _in the same Skosmos vocabulary_ (here "ScolomFr 3.2"), and not to concepts in another vocabulary.

I think the issue lies here : https://github.com/NatLibFi/Skosmos/blob/master/model/Concept.php#L253

I suggest adding an extra optional parameter to `Model:guessVocabularyFromURI`, that would the current vocabulary. When trying to guess the vocabulary of a URI, if this parameter was set, the method `guessVocabularyFromURI` could first test if the URI belongs to the current vocabulary.

tfrancart:
Fix #666


osma:
Merged, thanks!

PR #676, opened 2017-12-19
Domain hierarchy display

### LINKED ISSUE #675: Ability to display concept schemes domains in the tree
Following #634 and its integration in #663, we know want to display the "dct:subject" of concept schemes as first-level nodes in the hierarchy tree.

This is actually already visible at https://vocabulaires.irstea.fr/thesaurus (click on "hiérarchie" tab).

This requires the following modification :
  1. When creating the tree from concept schemes, create first-level nodes for each concept scheme domain, and add concept schemes underneath; concept schemes without a domain will be inserted at the first level.
  2. When hooking top concepts to concept scheme nodes in the tree, also search at the second level in the tree (if not found a the first level)
  3. Sorts the tree nodes to make domain nodes appear before concept schemes nodes without domains.

PR to follow.

tfrancart:
Fix #675


osma:
Closing due to lack of response. Please open a new PR if you still want this to be included. It would be better to contribute new features in small pieces that are easy to review.

Issue #665, opened 2017-11-30
SERVICE_LOGO is not used


osma:
There is a constant SERVICE_LOGO defined in config.inc dist, plus infrastructure in GlobalConfig to query for its value and pass it to templates, plus unit tests. But it is not actually used in the templates. This was [asked about](https://groups.google.com/d/msg/skosmos-users/ziIZGkRTdJs/Meb5Hu5oAQAJ) on skosmos-users.

The supported way of adding a logo is via a custom stylesheet (see #629 for a similar issue). I think we should just remove the dead code related to SERVICE_LOGO.

osma:
Fixed by #682

PR #682, opened 2018-01-09
Remove SERVICE_LOGO old constant and related code

kinow:
Fix for issue Issue #665.

Looks like the WebController was even setting the variable in the twig template, but it was just never used.

Removed the constant, tests, and the code in the controller. phpunit tests passing fine in my local computer.

Cheers
Bruno


osma:
Great, thanks for the PR!

PR #683, opened 2018-01-09
Add Jena Fuseki to gitignore

kinow:
Just to avoid someone (probably me :-) )including the Apache Jena Fuseki used for tests in a commit.

(actually did that, then removed from the staged changes)

Cheers
B

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ report
 components
 composer.phar
 composer.lock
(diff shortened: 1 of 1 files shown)


osma:
Makes sense. I've also noticed that the Fuseki directory clutters `git status` output, but never bothered to add it to `.gitignore`. Thanks!

Issue #686, opened 2018-01-13
Missing skosxl:literalForm causes crash


semanticfire:
When a skosxl:literalForm property is not present on the label resource the code will crash due to calling a member on a non object

PR will follow soon

henriyli:
backported to v1.10-maintenance

PR #687, opened 2018-01-13
Fix hard crash on missng skosxl:literalForm

semanticfire:
If skosxl:literalForm is missing on a label resource the $label variable remains uninitialzed and causes a hard crash
related to #686


henriyli:
Thanks!

Issue #673, opened 2017-12-14
Embed JSON-LD data on concept page


osma:
We could embed the graph with the concept information in HTML pages generated by Skosmos. This way it would be available both to search engines (related to #583) as well as Skosmos plugins.

henriyli:
This is now implemented but we should properly document the feature. @kouralex

PR #689, opened 2018-01-23
Embed concept and externalMappingProperty resource data as JSON-LD to page source #673 #674

kouralex:
Fixes issues #673 #674


osma:
Agree with @henriyli. All long URIs should be changed to short CURIEs and the necessary additional prefix declarations added to Model.php

PR #703, opened 2018-02-07
Some css fixes and updated default language name

ismakutl:
CSS updates for about, feedback and topbar
- Deleted some unnecessary lines of css
- More responsive (mobile)
- Better working topbar
- Sidebar fixed

Twig - topbar.twig
- Fixed default language name


osma:
Thanks for the PR!

In general it's better if you can make separate PRs for separate enhancements, but this looks pretty clear.

henriyli:
This is how the sidebar renders for me without the margins:
screenshot from 2018-02-08 12-25-33

ismakutl:
Can you check it now? Sidebar should be fixed now

henriyli:
It seems a lot better now! But there still seems to be some issues with the page width. @ismakutl 
screen shot 2018-02-13 at 12 35 40

ismakutl:
screenshot
I don't have any problems with it, there's probably a conflict with your css or maybe old cached css?

osma:
Can both of you try a concept page, toying with different browser window widths?

henriyli:
I tried this again and it would seem that the vocabulary home page does indeed work well. But the concept pages seem to have resizing issues when the viewport is wider than 1090ish pixels.

ismakutl:
I had this problem 1 time and it's gone, but I found it.

in your styles.css you have something like this line of code
`.main-container { margin: 15px; }`

and it should be changed to
`.main-container { margin: 15px auto; }`
you can change 15px to whatever you want, but the second argument should be auto to center it

resource/css/styles.css
```diff
@@ -796,16 +804,22 @@ span.xl-pref-label > img {
...
   border-top: 2px solid #394554;
-  width: 500px;
 }
 
 /* sidebar styles
  ***************************/
 
+.sidebar-wrapper {
+  position: relative;
+  margin-right: 15px;
```
osma: This should be `margin-right: 25px` to better match the current layout.

resource/css/styles.css
```diff
@@ -953,10 +968,13 @@ div.input-group > .input-group-addon {
...
 }
 
-.alphabetical-search-results li, .group-container > p, .group-container > ul {
+.alphabetical-search-results li,
+.group-container > p,
+.group-container > ul {
   font-size: 14px;
   line-height: 15px;
   list-style-type: none;
+  line-height: 2em;
```
osma: This overrides the setting two lines up. Better modify the original one...

I think the original setting is better. The alphabetical index layout should be quite tight because it may contain a huge number of entries. We tested and found that 16px could be a good value.

resource/css/styles.css
```diff
@@ -1017,7 +1035,8 @@ th, td {
 }
 
 tr {
-  border-image: '../pics/stripe.png' 20 20 repeat;
+  display: flex;
```
osma: This causes layout problems on the vocabulary front page:

skosmos-css-problems

Are you sure it's a good idea to change all tables to flex layout? Will it work with older browsers?

view/topbar.twig
```diff
@@ -15,7 +15,7 @@
 <div id="language"><span class="navigation-font">|</span>
   {% for langcode, langdata in languages %}
   {% if request.lang != langcode %}
-  <a id="language-{{ langcode}}" class="navigation-font" href="{{ request.langurl|replace({("#{request.lang}/"): "#{langcode}/"}) }}"> {{langdata.name}}</a>
+  <a id="language-{{ langcode}}" class="navigation-font" href="{{ request.langurl|replace({("#{request.lang}/"): "#{langcode}/"}) }}"> {{langdata.lemma}}</a>
```
osma: The problem you are seeing (with **in_this_language** displayed) is caused by missing locales on your system. See the [FAQ](https://github.com/NatLibFi/Skosmos/wiki/FAQ-and-Troubleshooting). Please revert this part.

osma:
Hi @ismakutl , is it possible for you to address the remaining comments so that we could merge this for Skosmos 2.0? Especially the flex layout is a bit scary. The other issues I commented on are easy to fix while merging but sorting out that one takes a bit more work...

osma:
Closing due to lack of response to change requests from original contributor. Please reopen, or open a new PR, if you still want to contribute this work.

PR #700, opened 2018-02-04
Add Dockerfile and a docker-compose to run Skosmos with Fuseki

kinow:
Not sure if others would be interested. I started a simple Docker environment as I am on a new notebook and didn't want to install PHP + Apache + Fuseki etc just for developing SKOSMOS again.

To get everything up and running, once in the project folder:

```
$ docker-compose up
$ curl -I -u admin:admin -XPOST --data "dbName=skosmos&dbType=tdb" -G http://localhost:3030/$/datasets/
$ curl -I -X POST -H Content-Type:text/turtle -T vocabularies.ttl -G http://localhost:3030/skosmos/data
```

And with that SKOSMOS should be running on http://localhost:8000.

Feel free to close if that's not interesting at this point :-) just in case someone else was thinking about doing something along these lines. It can be easily modified to also support nginx.

PR #704, opened 2018-02-09
Updated language replace

JorgWieme:
http://localhost:3030/something_ending_on_en/en/ becomes http://localhost:3030/something_ending_on_nl/nl/  when trying to change to change languages. The new replace has this fixed.

### DIFF
--- a/view/topbar.twig
+++ b/view/topbar.twig
@@ -6,7 +6,7 @@
   <ul class="dropdown-menu dropdown-menu-right">
   {% for langcode, langdata in languages %}
   {% if request.lang != langcode %}
-  <li><a id="language-{{ langcode }}" class="versal" href="{{ request.langurl|replace({("#{request.lang}/"): "#{langcode}/"}) }}"> {{langdata.lemma}}</a></li>
+  <li><a id="language-{{ langcode }}" class="versal" href="{{ request.langurl|replace({("/#{request.lang}/"): "/#{langcode}/"}) }}"> {{langdata.lemma}}</a></li>
   {% endif %}
   {% endfor %}
   </ul>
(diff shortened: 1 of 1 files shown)


view/topbar.twig
```diff
@@ -15,7 +15,7 @@
 <div id="language"><span class="navigation-font">|</span>
   {% for langcode, langdata in languages %}
   {% if request.lang != langcode %}
-  <a id="language-{{ langcode}}" class="navigation-font" href="{{ request.langurl|replace({("#{request.lang}/"): "#{langcode}/"}) }}"> {{langdata.name}}</a>
+  <a id="language-{{ langcode}}" class="navigation-font" href="{{ request.langurl|replace({("/#{request.lang}/"): "/#{langcode}/"}) }}"> {{langdata.lemma}}</a>
```
osma: Is it intentional that you replace `langdata.name` with `langdata.lemma` here?

JorgWieme: Yes, before changing it, all languages were displayed with /in_this_language**

kinow: Interesting that when running with Docker (see pull #700), I get the same, i.e. **in_this_language**. But when I get the changes in this pull request, at least that text is fixed, and displays now "svenka" and "English" instead. So at least it seems to fix this issue for my environment.

kinow: Fixes the text, but now I noticed that the links are not working. If I change in the URL from /fi/ to /en/, everything works. Clicking on the links doesn't seem to be working for me 👍

osma:
Exactly, the `in_this_language` is a special gettext id that should be translated to a phrase in the target language. If you see it in the UI the most likely cause is that you have a language defined in `config.inc` but haven't installed the corresponding locale. Not sure how locales work inside Docker containers, but you should make sure all locales mentioned in `config.inc` are available also from inside the container.

kinow:
Aaahh the locale packages. Got to update the pull request for docker then, and include a few default locales. Thanks Osma!!!

osma:
As discussed above, the **in_this_language** problem is caused by missing locales on the system and is not relevant. `langdata.name` should be left intact, not changed to `langdata.lemma`.

The original problem is real - if the vocabulary name ends with a language tag, it will be changed by the language change links. But the suggested fix is also broken, as pointed out by @henriyli: the language change links stop working e.g. on the About page. So we cannot merge this as is.

joelit:
It seems that this is a limitation of twig's replace() function, which does not have proper regex capability. The language replacement will have to be handled somewhere else than topbar.twig. Closing this issue and creating a new one.

Issue #661, opened 2017-11-20
Add a copy to clipboard button


henriyli:
## At which URL did you encounter the problem?
Any concept page.

## What is the expected output? What do you see instead?
It would be nice to be able to copy preferred labels or URIs with a single click. There is now a single click highlight functionality, but the user still needs to manually copy the highlighted content to the clipboard.

This could be solved by using something like: https://clipboardjs.com/ or http://www.jstips.co/en/javascript/copy-to-clipboard/

kinow:
Tested on dev.finto.fi/geo/fi/page/GTK2015ID5, and the clipboard icon is working fine on Firefox and Chrome. Tested using Ubuntu Linux LTS. I think this can be closed? 🎉

henriyli:
I agree. Thanks @kinow !

PR #709, opened 2018-02-13
Issue #661 add a copy to clipboard button

kinow:
Just wanted to learn how it was done in JS. First tried zeroclipboard as I thought that's what GitHub was using, but the flash hidden component was failing. Then tried the simplest example provided in the issue #661.

Here's the result:

image

It re-uses the function to make a selection. So the example code provided is simplified, but we get the same effect in the end.

Not completely happy with final button, so happy to take any suggestions for it 👍 

Bruno

### DIFF
--- a/resource/js/docready.js
+++ b/resource/js/docready.js
@@ -94,8 +94,8 @@ $(function() { // DOCUMENT READY
   countAndSetOffset();
 
   // Make a selection of an element for copy pasting.
-  function makeSelection() {
-    var $clicked = $(this);
+  function makeSelection(e, elem) {
+    var $clicked = elem || $(this);
     var text = $clicked[0];
     var range;
     if (document.body.createTextRange) { // ms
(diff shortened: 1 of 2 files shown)


henriyli:
Great timing! We were talking about this feature earlier on the same day when you made the pull request. I think the button would be better without the border. Maybe something like this?

screenshot-2018-2-15 fintoto yso cat

kinow:
Looks so much better! I'm in the airport now, disconnected until Monday. But if you'd like, feel free to update the pull request or even create a new one if necessary :-) otherwise I will update it with the your markup next week. Thanks!

henriyli:
Thank you @kinow ! This is now merged to master. I added the css in the master branch, since I couldn't figure out the proper GitHub workflow to do it.

kinow:
Not a problem. Thanks a lot @henriyli !

Issue #712, opened 2018-02-21
No concepts displayed in Alphabetical, if no concept starts with A


Kordishal:
## At which URL did you encounter the problem?

http://bartoc-skosmos.unibas.ch/aap/en/

## What steps will reproduce the problem?
1. Click on the link.

## What is the expected output? What do you see instead?
Should show the concepts of the first letter available. The same problem persists for languages with non-Latin characters. 

The initial Alphabetical List is empty giving the impression that there are no concepts under the first letter.

Once you click on one of the letters it works as expected.

## What browser did you use? (eg. Firefox, Chrome, Safari, Internet explorer)
Firefox

PR #716, opened 2018-02-23
WIP Issue #712: display the concepts of the first letter available

kinow:
For issue #712. If no concepts exist with letter `'A'`, then Skosmos will list the letters, but the concept tree will be empty.

This happens cause the `Request` by default assumes the informed letter is `A` when none given.

Instead, we can assume it is empty. Initial tests indicate everything worked with no issues. I still have to confirm the queries are not working by accident, more specifically in `Vocabulary#searchConceptsAlphabetical`.

But happy to take some reviews before digging deeper into this.

Cheers
Bruno

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,13 +19,12 @@ before_script:
 after_script:
   - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT
   - pkill -f 'java -Xmx1200M -jar'
-addons:
-  code_climate:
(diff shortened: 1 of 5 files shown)


osma:
Thanks for the PR. 

We tested this, but it doesn't work quite right. The initial alphabetical index shows the first concepts from the vocabulary (in the case of YSO, starting with numbers), but scrolling down the list doesn't work since none of the letter links is active and there is a JS error.

Trying to figure out next what would be the proper solution...

osma:
Got it working with an additional fix that picks the first letter from the available letters. Thanks!

kinow:
:tada: thanks for reviewing and fixing it!

Issue #626, opened 2017-09-25
REST api function for retrieving all concepts present in a vocabulary


d0rg0ld:
Hi,

As far as I understand the current FINTO/SKOSMOS API doc, I can retrieve information about all the vocabularies present in the collection and can then retrieve top level concepts from or search for specific concepts within each of them. What I am missing is the possibility to directly harvest all declared concepts from each vocabulary, the only currently available option in this regard seems to be an API triggered RDF dump of the whole vocabulary. Probably I did not understand the API documentation well enough and it is already possible, in this case I'd be glad to learn how to do it, otherwise I'd like to suggest this as future enhancement.

Kind regards,
Doron Goldfarb

osma:
You're right, getting a list of all the concepts from a vocabulary with the current API is quite awkward. It's not really intended for that purpose originally. As you mentioned, asking for the full vocabulary is probably the easiest way.

Can you elaborate on what kind of use case you have and what information you would need for each concept in the response?

d0rg0ld:
Hi,

thanks for your reply! In the context of the EUDAT project we are currently working on a prototype for aggregating the contents from multiple semantic repositories, i.e. ontology hubs such as BioPortal etc., for providing centralized concept search and overlap analysis functionality - For more info I can offer a link to a recent workshop paper on our endeavor: http://ceur-ws.org/Vol-1933/paper-7.pdf.  

Coming from the environmental domain where we have a number of SKOS based resources, I'd really also like to include SKOS related repositories such as FINTO/SKOSMOS in our aggregator, because we consider re-using SKOSMOS for our own resources and it would be great if we could then directly harvest all stored concepts for aggregation. 

Regarding the information for each concept, the basics would include concept IRI, preferred label, synonyms and description, information we could use to provide lookup services to find the right concept for e.g. semantic annotation. 

I hope that this use case is of interest to you, I am looking forward to learning your opinion on this!

Thanks and kind regards,
Doron

d0rg0ld:
Hi, 

I played around with the code a bit and extended the API with a new call "allConcepts". 

As far as my use case was concerned, it works fine, returning ID, prefLabel, altLabel(s) and Description for each concept present in a vocabulary. The only thing I did not add yet is paging, which might be necessary for larger vocabularies.

In case you are interested, these are the files were I made changes, you can see the diff in a Skosmos fork on https://github.com/d0rg0ld/Skosmos, I forked it today from the current version and merged my changes there.

[...]

PR #739, opened 2018-03-12
REST api function for retrieving all concepts present in a vocabulary  #626 

d0rg0ld:


d0rg0ld:
Please ignore changes made to index.php, thanks

osma:
Thanks for the PR.
Can you please remove the irrelevant parts? E.g. `.gitignore`, `index.php` and the date manipulations.

I'm worried about the approach of creating a JSON string within SPARQL. There is no escaping, so for example if a definition contains a double quote, will it break?

d0rg0ld:
Hi, I removed the irrelevant parts. Thanks for your comment re JSON Strings from SPARQL, you are right that an unescaped double quote char would indeed break this. I had always thought that quotes must be escaped in RDF, but https://www.w3.org/TR/rdf11-concepts/#dfn-literal-value doesnt mention it - thanks for pointing this out. I suspect, however, that all RDF serializations nevertheless require it. But of course it could happen. I only tested the code with 3,6K concepts from our EnvThes thesaurus, which did not cause any problems.

osma:
Thanks for cleaning up the PR! Now it's a lot easier to review.

The fact that RDF serializations require escaping is not relevant here, since the SPARQL endpoint always sees the literals as-is. When you perform string operations with SPARQL, you are dealing with the raw strings, not escaped ones.

I would like to see a unit test that shows that the Vocabulary.getAllConcepts() method works even in extreme cases. Skosmos is used with a lot of vocabularies and some of them have truly strange data...
For example the resource `test:ta118` in the [test.ttl](https://github.com/NatLibFi/Skosmos/blob/master/tests/test-vocab-data/test.ttl#L79) has a deliberately tricky prefLabel. Will it break your JSON serialization?

There are also a couple of SPARQL queries already in Skosmos that perform a similar trick by creating CSV strings with SPARQL string functions: [formatPropertyCsvClause](https://github.com/NatLibFi/Skosmos/blob/master/model/sparql/GenericSparql.php#L758) and [formatPrefLabelCsvClause](https://github.com/NatLibFi/Skosmos/blob/master/model/sparql/GenericSparql.php#L774). I think encoding of special characters in CSV is easier than in JSON (fewer characters to escape), so you may consider switching to CSV.

osma:
@d0rg0ld Are you still working on this? I'm afraid that this cannot be merged in its current form as the concerns detailed above have not been addressed.

d0rg0ld:
@osma unfortunately, I haven't had the time yet, still planning to look into this bit cant promise when this will be the case. Would it be ok to leave it open for now? Thanks

osma:
@d0rg0ld Thanks for the update. No problem keeping it open. I hope you can find the time eventually, it would be a good feature :)

osma:
I'm closing this for now due to inactivity. Please reopen (or open a new PR) if you still want to contribute this work.

Issue #737, opened 2018-03-12
Invalid sort order for numeric notation codes in PTVL


osma:
## At which URL did you encounter the problem?
http://dev.finto.fi/ptvl/fi/page/?uri=http://urn.fi/URN:NBN:fi:au:ptvl:

## What steps will reproduce the problem?
1. Go to concept scheme page of PTVL (above URL)
2. Look at the list of top concepts

## What is the expected output? What do you see instead?

Expected to see top concepts sorted by notation code in natural (numeric) sort order: P1, P2, P3 ...

Instead they are ordered P10, P11, P12, ... P1, P20, P21 ...

This is a regression caused by the fix to #679.

I think the solution would be to use natural sort order (`strnatcasecmp`) when sorting by notation code, but `strcoll` sorting for labels as was done in #679 / #680 .

PR #741, opened 2018-03-20
Issue #737 sorting concept property values correctly

henriyli:
Using natural case sort for sorting concept property values when sortByNotation is true, fixes issue #737.

This PR should undo the damage done in the previous bug fix. I had to add an extra parameter for the ConceptProperty constructor. Is this ok or should the vocabulary configuration value be passed some other way?


osma:
There's a lot of clutter in the commit history. I will do a manual rebase before merging to verify that only the intended changes are included.

osma:
Merged after manual rebase. Thanks!

Issue #731, opened 2018-03-07
Front page content lang


kouralex:
## At which URL did you encounter the problem?
http://finto.fi/en/?clang=fi
## What steps will reproduce the problem?
1. Go to link above

## What is the expected output? What do you see instead?
Content of the RDF in "Available vocabularies and ontologies" is always in the language of the UI, even if one has content language set in the address bar (contradicting the behavior on vocabulary and concept pages). The question is if this is the desired functionality?

jarmosaarikko:
This problem is related to the UI links on the top right hand corner of the UI. Clicking the link, the interface should change.  what is the difference in behaviour between these links:
http://finto.fi/fi/?clang=en
http://finto.fi/sv/?clang=en
http://finto.fi/en/?clang=en
should the interface be in English, but the titles of the vocabularies be in the selected language?
compared to these:
http://finto.fi/sv/?clang=fi
http://finto.fi/sv/?clang=sv
http://finto.fi/sv/?clang=en
Now the Vocabulary titles should be in Swedish in all three, but interface in the selected language?
is this correct?

osma:
I think the vocabulary titles showed on the front page should follow the selected content language, as suggested by @kouralex .

PR #747, opened 2018-03-21
Issue #731 - showing front page vocabulary titles with the chosen content language

henriyli:
Fixes issue #731.


osma:
I think the fallback, in case the vocabulary has no title in the content language, should be the default language of the vocabulary. Not the UI language as suggested by @kouralex .

PR #748, opened 2018-03-23
Add SKOS relations to JSON-LD serialization

danmichaelo:
Realized earlier today that `skos:relatedMatch` is returned as `related:Match` in the JSON-LD serialization.. [example](http://data.ub.uio.no/skosmos/rest/v1/data?uri=http%3A%2F%2Fdata.ub.uio.no%2Frealfagstermer%2Fc012920&format=application/json)  🌵  Ouch!

As a start, I added a failing test case to document it.

I think the easiest fix is to add the various SKOS mapping properties (exactMatch, closematch, etc.) to the default JSON-LD context). Let me know what you think.

### DIFF
--- a/controller/RestController.php
+++ b/controller/RestController.php
@@ -112,27 +112,43 @@ class RestController extends Controller
         return $parameters;
     }
 
-    private function transformSearchResults($request, $results)
+    private function transformSearchResults($request, $results, $parameters)
     {
         // before serializing to JSON, get rid of the Vocabulary object that came with each resource
         foreach ($results as &$res) {
             unset($res['voc']);
         }
-        $ret = array(
-            '@context' => array(
-                'skos' => 'http://www.w3.org/2004/02/skos/core#',
-                'isothes' => 'http://purl.org/iso25964/skos-thes#',
-                'onki' => 'http://schema.onki.fi/onki#',
-                'uri' => '@id',
-                'type' => '@type',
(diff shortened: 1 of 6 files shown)


osma:
Ouch indeed! Thanks for spotting! 

A failing test is a great way to report an issue :)

The proposed solution sounds good. I think this will only affect `skos:relatedMatch` since it's the only property whose URI starts with another SKOS property. But defining all the SKOS mapping properties is better for clarity and consistency.

danmichaelo:
Ok, added the mapping relations to the context for the `data` endpoint.

I also wanted the output from the `search` endpoint to be more consistent with the output from the `data` endpoint, so I added any additional fields being requested using the `fields`parameter to the JSON-LD context there. Note that this change is technically breaking (if users just rely on the `'skos:'` prefix being present in the keys in the resulting JSON without taking the JSON-LD context into account), but I guess the impact should be low since it only affects additional fields requested using the `fields` parameter.

osma:
Excellent work! I see you even added a RestControllerTest class, which we didn't have before.

The only problem I see is that the Travis tests failed due to an unrelated PHP version compatibility issue. That should be fixed in master now. I will try merging this first into a separate branch to make sure it doesn't actually break the tests.

osma:
Tests OK after merge. Thanks a lot!

Issue #774, opened 2018-06-07
Add documentation for Plugins


kinow:
Part of this is from discussion at #771 . What about a page with the following title & structure?

--- snip ---

# Extending Skosmos

Besides changing Skosmos configuration, and modifying the code directly, another option to extend and customize Skosmos is through **plugins**.

## Creating Plugins

A plugin consists of a folder, with a unique name. The **folder name** is also used as the **plugin name**. Within the folder, there should be a file **`plugin.json`**, which provides information about each plugin's resources. For example:

```json
{
    "js": ["widget.js"],
    "css": ["stylesheet.css", "anotherfile.css"],
    "templates": ["template.html", "anothertemplate.html"],
    "callback": ["my_callback"]
}
```

The **js** resources are used in the Skosmos template file. These JavaScript files are loaded for every page, and are located near the bottom of the page (i.e. near the `</body>` HTML end tag).

The **css** resources are loaded also for each page by Skosmos template file. The CSS stylesheet files are put within the HTML `<head>` section.

The **templates** resources are loaded for JavaScript, using [Handlebars](https://handlebarsjs.com/), and the syntax looks like:

```javascript
<script id="{{ id }}" type="text/x-handlebars-template">
{{ template|raw }}
</script>
```

The template ID is automatically generated, based on the plugin name, and on the template file name minus the extension. If your plugin name is "*acrobata*" and the template file name is "*privacy.html*", the template ID will be "*acrobata-privacy*".

This is important when you need to use the template in JavaScript. It is also important to note that the extension is extracted simply by creating a substring of the template file name from the position `0` until the first appearance of the `.` symbol. So if your template file name contains multiple `.` characters, you may have unwanted template ID's.

The **callbacks** resources are a called for each page loaded.

## Installing Plugins

When you install Skosmos, there should be a folder in the root application folder, called "*plugins*". Inside this folder you can put your plugin folders. As explained in the previous section, the folder names are used as plugin names as well.

For each request, Skosmos loads the content of each **enabled plugins'** `plugin.json`, and utilises its resources, as explained in the previous section.

Plugins are enabled in the configuration file, `config.ttl`. There are two ways to activate plugins. This is explained in the next section.

## Activating Plugins

The first way to activate plugins in Skosmos is globally. This is done via the `skosmos:globalPlugins` property in the `config.ttl` configuration file.

```turtle
:config a skosmos:Configuration ;
  # ...
[...]

PR #775, opened 2018-06-07
Replace PHP explode function call by a Turtle list for globalPlugins

kinow:
Tested locally in Skosmos + Docker, and also ran `phpunit`. No issues or changes in tests, as we are still returning an array. But in the configuration, instead of using spaces to separate plugin names, now we have a turtle list.

Also updated the docs for a method while reading the code and writing the Wiki page for #774 issue.

### DIFF
--- a/config.ttl.dist
+++ b/config.ttl.dist
@@ -64,7 +64,7 @@
     # default time a user must wait before submitting a form
     skosmos:uiHoneypotTime 5 ;
     # plugins to activate for the whole installation (including all vocabularies)
-    skosmos:globalPlugins "" .
+    skosmos:globalPlugins () .
 
 # Skosmos vocabularies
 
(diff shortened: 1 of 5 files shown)


osma:
Thanks! Should the example `config.ttl` be updated as well? What about `testconfig.ttl` used by unit tests?

I'm just leaving for a trip so I don't dare to merge this right now as I might break our installations that follow the `master` branch. But I will get back to it when I return!

kinow:
Logged off already. Will check other files tomorrow, well remembered.

Take your time, and enjoy your trip ! :tada:

kinow:
>Should the example config.ttl be updated as well?

Done!

>What about testconfig.ttl used by unit tests?

It did not have a `globalPlugins`. There were no tests before the that either. So I've added an example `globalPlugins` with a list value, and a simple unit test as well.

osma:
Finally got around to merging this. Thanks!

PR #781, opened 2018-08-22
Display concept types for search

### LINKED ISSUE #776: Concept type not mentioned in enter search results
Concept type is not mentioned in enter search results. In ontologies where there are several concept types this information would be useful. In domain ontologies which are linked to general ontologies (such as JUHO, which is linked to YSO) this information is crucial. At the moment, enter search does not show whether a given concept is a domain ontology concept or a general ontology concept. This information may be inferred from the concept's URI, but not in all cases. And as far as I know, URI name space isn't exactly meant for indicating concept types.

See also: https://github.com/NatLibFi/Skosmos/issues/546

kinow:
Add concept types for search results. Icon edited from https://feathericons.com/, but happy to edit the commit and replace by another icon if necessary (due to colour, size, license, etc).

Closes #776


osma:
Thanks, looks very promising! But a couple of issues:

1. I tried this with YSO, which has two rdf:types for some concepts, and for some reason one of them shows up empty:

skosmos-type

The other one shows up as skos:Concept, which is not ideal, it would be better to translate it to "Concept" (in English, or the corresponding term in the current UI language).

Replacing `{{ concepttype }}` with `{{ concepttype | trans }}` in search-results.twig should handle the translation, but currently the results are a bit funny for the first type which is returned as the empty string, resulting in a translation with POT file metadata. So the empty type problem has to be fixed first, then the translations should work just by adding `| trans`.

2. Regarding the icon - the feather icons are nice, and the MIT license they use is of course very liberal, but still using them would add another dependency, i.e. a new source that must be credited. So it would be better to just draw an icon from scratch, or use a CC0 icon (for example one from Iconmonstr.com) that doesn't require attribution.

osma:
Sorry, I mixed up things, Iconmonstr isn't actually CC0. Their license prohibits e.g. redistribution. So maybe not ideal here.

kinow:
Will take a look with more calm tomorrow. But a quick analysis today (last thing before going to bed, so might have missed a few things), I've found that:

* The other concept is actually a `NULL`, while the one being displayed is a valid string... no idea why the query returned a `NULL`... I checked the yso.ttl, and the types Concept & Hierarchy are correctly set up.

* Not displaying skos:Concept and showing the right translation instead sounds like a good idea. But both `{% trans concepttype %}` and `{{ concepttype | trans }}` resulted in:

image

Which honestly makes no sense right now :confused: 

* The icons from iconmonstr are amazing! So that's where you get the icons from? Bookmarked, and will update the PR again before the weekend.

Thanks!!!

kinow:
Hi @osma ,

Sorry, reading again I understood - finally - that I can't use those icons. Such a shame, because they look good IMO! :)

>I think it would be best to draw a custom icon. I can get someone to do that if necessary. You can concentrate on the code :)

That'd be great!

>By the way, the concept page already shows the type of concept (unless it's just skos:Concept). Maybe you can reuse that code?

Thanks for that! Now I can confirm my suspicion. The query doesn't have the prefix for `ysa-meta`, so it returns a resource with URI (i.e. instead of ysa-meta:Hierarchy, it's something like <http://...ysa-meta/Hierarchy> or so).

The code in this pull request worked with skos:Concept by luck. The "events and action" concept is of type skos:Concept and <ysa-meta:Hierarchy>.

Looks like the concept page has some checks for that already in place. I copied the part that prints the type, and it works out of the box, just with the incorrect layout.

Let me see if I can use the code from that page here then. Great idea @osma (as always).

Thanks!

kinow:
image

kinow:
Tried to simplify the code from the concept page. Hopefully didn't simplify it too much. It should display the same text from the concepts page.

joelit:
Yeah, I thought it was simpler to create a new PR than to push the changes to your branch.

kinow:
Not a problem for me @joelit , happy to close this one once #827 gets merged :-)

PR #785, opened 2018-08-27
Minor improvements

kinow:
Mentioned in another ticket that I had a few minor improvements to submit. All of these were found with PhpStorm, which I have been using for another project (normally use Eclipse). Simply downloaded a copy with my apache e-mail (free :tada: ), imported the project, and then did an "Inspect Code".

I think it will be easier to review each commit, as I tried to group the enhancements as they appeared in PhpStorm.

Some other interesting entries **not included in this commit** as I think it would require further analysis:

- unnecessary `\.` in JS regex
- Inconsistent return points: several missing return statement (some functions have a return, but if there is an error, it doesn't return anything, nor throws any exception)
- Duplicate array keys: In RestController, around line 228, an array with values is created to be returned. But if you look closer, and search for `title`, you will see that it first is set to `rdfs:label`, and then set again in the array to `dct:title` (@osma not in this commit, as I am not sure which one is correct)
- Void function result used: in controllers, we are using something like `return $this->returnError(....)`, but actually, the returnError does not return. It sets headers and echoes. I _think_ PHP stops the code execution and returns too, being lenient here. But it could change in the future. Will put another ticket for it someday

If we started using a bit more of type setting (e.g. `function blabla(): bool`, also phpdocs annotations, and those nice `@var $ret array`, etc) I think we could spot more probable bugs, and make the code simpler. I am working with a senior symfony guy, who is teaching me a few more tricks while helping on a project he recently set up. Might come back later with a few more ideas later :-) :running_man:

### DIFF
--- a/model/ConceptPropertyValueLiteral.php
+++ b/model/ConceptPropertyValueLiteral.php
@@ -67,7 +67,6 @@ class ConceptPropertyValueLiteral extends VocabularyDataObject
 
     public function hasXlProperties()
     {
-        $ret = array();
         $graph = $this->resource->getGraph();
         $resources = $graph->resourcesMatching('skosxl:literalForm', $this->literal);
         return !empty($resources);
--- a/model/LabelSkosXL.php
+++ b/model/LabelSkosXL.php
@@ -9,7 +9,7 @@ class LabelSkosXL extends DataObject
     }
 
     public function getPrefLabel() {
-        $label;
+        $label = null;
         $labels = $this->resource->allResources('skosxl:prefLabel');
         foreach($labels as $labres) {
             $label = $labres->getLiteral('skosxl:literalForm');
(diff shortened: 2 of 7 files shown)


kinow:
Will wait to see what are the test results in Travis CI. In my environment 5 tests failed, but I suspect due to my PHP 7.2. They failed with "Cannot use stdClass as Object because 'Object' is a special class name", and it looks like it's being fixed right now in JsonLD -> https://github.com/lanthaler/JsonLD/pull/79

kinow:
Tests that failed in my environment:

- ConceptTest::testExternalResourceReifications
- ConceptTest::testProcessExternalResource
- ModelTest::testGetRDFWithVocidAndURIasJSON
- ModelTest::testGetResourceFromUri
- RestControllerTest::testDataAsJson

PR #790, opened 2018-08-29
Vagrant virtual environment for deploying and running Skosmos

mikipernu:


ansible/playbook.yml
```diff
@@ -0,0 +1,16 @@
...
+- hosts: all
+  become: True
+  gather_facts: False
+  roles:
+    - base
+    - openjdk
+    - fuseki
+    - skosmos
+  pre_tasks:
+    - name: Install Python 2.7 for Ansible
```
osma: I wonder why this is needed?

ansible/playbook.yml
```diff
@@ -0,0 +1,16 @@
+---
+- hosts: all
+  become: True
+  gather_facts: False
+  roles:
+    - base
+    - openjdk
+    - fuseki
```
osma: Fuseki is installed but not actually used by Skosmos here (it is configured to use the Finto dev SPARQL endpoint). I think it would be better to start with a Skosmos-only Vagrant install. Later we can change to include Fuseki as well, with some vocabularies loaded and actually used by Skosmos

ansible/roles/fuseki/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
+
+- name: Download Fuseki tarball
+  get_url:
+    url: https://archive.apache.org/dist/jena/binaries/apache-jena-fuseki-2.3.0.tar.gz
```
osma: The version number is hardcoded in a number of places. It would be better to define it in only one place.
Version 2.3.0 is pretty old, 3.8.0 is the latest version of Fuseki1 (and final, since new development is done only on Fuseki2).

Issue #792, opened 2018-08-31
Not all new-style configuration settings for the skosmos:Configuration have an effect


rwalkerands:
`config.ttl.dist` has:

```
:config a skosmos:Configuration ;
    # ... several lines omitted ...
    # sparql-query extension, or "Generic" for plain SPARQL 1.1
    # set to "JenaText" instead if you use Fuseki with jena-text index
    skosmos:sparqlDialect "Generic" ;
```

But setting the default SPARQL dialect at this top level doesn't have any effect. To confirm: change this setting for the `skosmos:Configuration` to 
```
    skosmos:sparqlDialect "JenaText" ;
```
and observe that the queries _remain unchanged_.

The method ``Vocabulary::getSparql()`` seems to be correct:
```
$dialect = $this->resource->get('skosmos:sparqlDialect');
$dialect = $dialect ? $dialect->getValue() : $this->model->getConfig()->getDefaultSparqlDialect();
```

But it also seems that commit 6d28f04 did not update `model/GlobalConfig.php` completely to use the new configuration style; it still says:

```
public function getDefaultSparqlDialect()
{
    return $this->getLiteral('DEFAULT_SPARQL_DIALECT', 'Generic');
}
```

Perhaps it should be:
```
public function getDefaultSparqlDialect()
{
    return $this->getLiteral('skosmos:sparqlDialect', 'Generic');
}
```

And note also the surrounding methods in `GlobalConfig`: `getTemplateCache()`, `getFeedbackAddress()`, etc., which seem to have the same issue.

kinow:
Oh, mea culpa @rwalkerands ! Will probably have some time this weekend to re-sync the repository and investigate it. But thanks already for always including plenty information in the bug reports. It makes fixing the code so much easier.

osma:
Thanks a million for reporting and great analysis @rwalkerands ! This actually explains a couple of mysterious issues issues we've had with Skosmos 2.0 - we've been unable to deploy it to Finto.fi due to performance issues but didn't yet have time to investigate in depth. Getting the SPARQL dialect wrong is probably a major factor in that.

kinow:
Oh, my reply didn't go to GitHub, sorry. Here's my message that goes before @osma's comment above:

>Super sorry for the regressions pre and post the 2.0 release Osma.
>Going to find some time to work on the issues reported, and try to increase the test coverage :(
>Cheers
>Bruno

[...]

PR #793, opened 2018-08-31
Fix config

kinow:
Fix for #792 . Re-writes `GlobalConfigTest` to test values other than the default ones. Also includes more tests for `VocabularyConfig`.

Tests passing locally, with exception of those JSON-LD ones, that were failing in another PR.

Happy to update the pull request, so feel free to suggest enhancements/fixes, please.

Thank you! And sorry for the regression!
Bruno

### DIFF
--- a/model/GlobalConfig.php
+++ b/model/GlobalConfig.php
@@ -84,6 +84,7 @@ class GlobalConfig extends BaseConfig {
     /**
      * Parses configuration from the config.ttl file
      * @param string $filename path to config.ttl file
+     * @throws \EasyRdf\Exception
      */
     private function parseConfig($filename)
     {
@@ -95,7 +96,7 @@ class GlobalConfig extends BaseConfig {
 
     /**
      * Returns the graph created after parsing the configuration file.
-     * @return Graph
+     * @return \EasyRdf\Graph
      */
     public function getGraph()
     {
@@ -125,7 +126,9 @@ class GlobalConfig extends BaseConfig {
         if (!is_null($languageResources) && !empty($languageResources)) {
             $languages = array();
             foreach ($languageResources as $languageResource) {
+                /** @var \EasyRdf\Literal $languageName */
                 $languageName = $languageResource->getLiteral('rdfs:label');
+                /** @var \EasyRdf\Literal $languageValue */
                 $languageValue = $languageResource->getLiteral('rdf:value');
                 if ($languageName && $languageValue) {
                     $languages[$languageName->getValue()] = $languageValue->getValue();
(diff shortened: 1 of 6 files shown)


kinow:
Waiting for Travis. Testing locally with Docker. The service name is working as expected. Am looking at the queries being executed with dialect `Generic`, and then looking at the queries again after I switch to `JenaText`.

Appears to be working OK

```
1c1
< [2018-08-31 13:06:15] INFO [qid 10447] SPARQL query:
---
> [2018-08-31 13:06:49] INFO [qid 27514] SPARQL query:
3a4
> PREFIX text: <http://jena.apache.org/text#>
16c17
<      ?s ?prop ?match . FILTER (STRSTARTS(LCASE(STR(?match)), 'wave') && LANGMATCHES(lang(?match), 'en'))
---
>      (?s ?score ?match) text:query (?prop 'wave*' 'lang:en*' 100000) .
51c52
< [2018-08-31 13:06:21] INFO [qid 10447] result: 5 rows returned in 5895 ms
---
> [2018-08-31 13:06:53] INFO [qid 27514] result: 5 rows returned in 3995 ms
53c54
< [2018-08-31 13:06:21] INFO [qid 95451] SPARQL query:
---
> [2018-08-31 13:06:53] INFO [qid 81368] SPARQL query:
146c147
< [2018-08-31 13:06:23] INFO [qid 95451] result: 252 triples returned in 1283 ms
---
> [2018-08-31 13:06:54] INFO [qid 81368] result: 252 triples returned in 1303 ms

```

kinow:
All done @osma, @rwalkerands . Going to log off again, but can update the pull request throughout this weekend and the next week. Bruno

osma:
Looks excellent! Especially the expanded unit test coverage is really good!

Also waiting for Travis now. I'm a bit worried that the changes to testconfig.ttl may have an effect on other tests, but you did run the tests locally I assume?

One thing we currently don't do is to run the tests both with JenaText and Generic dialects. This could perhaps be an idea for the future, since we can't now be sure that both versions of the SPARQL queries (when they differ) actually work.

kinow:
Most tests passed. Got 5 errors, but they are - as far as I can tell - unrelated. They are failing due to my ubuntu running php 7.2 and jsonld is using the stdClass somewhere it should not think...

>One thing we currently don't do is to run the tests both with JenaText and Generic dialects. This could perhaps be an idea for the future, since we can't now be sure that both versions of the SPARQL queries (when they differ) actually work.

Great idea! Even when coverage reaches 100%, some variations of the settings may affect other covered tests... so now I found useful focusing on both coverage and some scenarios for the tests.

PR #795, opened 2018-09-04
using fuseki2 for unit tests, fixes #791

### LINKED ISSUE #791: Use Fuseki2 for unit tests
Currently the tests set up a Fuseki1 instance. But 3.8.0 is the last full release of Fuseki1. We should switch the tests to use Fuseki2 instead. Snapshot builds were already disabled in #783 because new Fuseki1 distribution snapshots are no longer built.

henriyli:
Updated the bash script init_fuseki.sh to use fuseki2. This should fix #791

### DIFF
--- a/tests/init_fuseki.sh
+++ b/tests/init_fuseki.sh
@@ -1,25 +1,16 @@
 #!/bin/bash
 # Note: This script must be sourced from within bash, e.g. ". init_fuseki.sh"
 
-FUSEKI_VERSION=${FUSEKI_VERSION:-3.4.0}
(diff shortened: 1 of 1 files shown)


osma:
Great, thanks a lot! Unfortunately this PR has conflicts (probably based on an earlier version of init_fuseki.sh), but I don't think those should be difficult to resolve.

henriyli:
Oops, I'll have a look at it later this week!

henriyli:
Resolving the conflicts didn't get me much further. My fork seems to be completely broken somehow. I'll make a new pr when I get it working again.

PR #797, opened 2018-09-13
Vagrant Virtual Machine with Fuseki

mikipernu:
Add Fuseki to the Vagrant setup, and configure the virtual environment to use locally-hosted STW and Unesco vocabularies


osma:
Thanks, and sorry for the long delay. I haven't really tested this myself but I trust this is an improvement over the current Vagrant setup, which didn't include Fuseki at all, so I'm merging it as is.

PR #802, opened 2018-10-01
 Fix config migration of LOG_FILE_NAME=null

### LINKED ISSUE #801: Configuration migration script issue
In my old `config.inc`, I had the default value for `LOG_FILE_NAME`:

```
define ("LOG_FILE_NAME", NULL);
```

This was migrated into

```
skosmos:logFileName ""
```

which caused the following error:

```
Fatal error: Uncaught LogicException: Missing stream url, the stream can not be opened. This may be caused by a premature call to close(). in /var/www/skosmos/vendor/monolog/monolog/src/Monolog/Handler/StreamHandler.php:95 Stack trace: #0 /var/www/skosmos/vendor/monolog/monolog/src/Monolog/Handler/AbstractProcessingHandler.php(37): Monolog\Handler\StreamHandler->write(Array) #1 /var/www/skosmos/vendor/monolog/monolog/src/Monolog/Logger.php(337): Monolog\Handler\AbstractProcessingHandler->handle(Array) #2 /var/www/skosmos/vendor/monolog/monolog/src/Monolog/Logger.php(546): Monolog\Logger->addRecord(200, '[qid 32579] SPA...', Array) #3 /var/www/skosmos/model/sparql/GenericSparql.php(90): Monolog\Logger->info('[qid 32579] SPA...') #4 /var/www/skosmos/model/sparql/GenericSparql.php(607): GenericSparql->query('CONSTRUCT {\n <...') #5 /var/www/skosmos/model/Vocabulary.php(143): GenericSparql->queryConceptScheme('http://data.ub....') #6 /var/www/skosmos/vendor/twig/twig/lib/Twig/Extension/Core.php(1605): Vocabulary->getInfo('nb in /var/www/skosmos/view/vocab-shared.twig on line 9
```

Fixed easily by commenting out the line:

```
# skosmos:logFileName ""
```

But would be good if the migration script automatically did this.

danmichaelo:
Fixes #801


kinow:
Simple fix, looks good. I had to check the docs to check what was the result for `empty`. We used `is_null` in another part of that code, for `GLOBAL_PLUGINS`, but looks like `empty` returns the same result as `empty` for `NULL` (https://www.virendrachandak.com/techtalk/php-isset-vs-empty-vs-is_null/).

So +1 LGTM and thanks for the fix @danmichaelo !
Bruno

PR #812, opened 2018-11-15
#fixes #791; also update fuseki to 3.9.0

### LINKED ISSUE #791: Use Fuseki2 for unit tests

kouralex:


osma:
I noticed that the Travis tests were still being executed using Fuseki 3.8.0, not 3.9.0. But at least they are using Fuseki2, not Fuseki1.

PR #814, opened 2018-11-16
Added a check for empty arrays to avoid the situation in #813

### LINKED ISSUE #813: PHP Notice: Undefined index -- querying labels for properties which don't have one defined
https://finto.fi/kauno/fi/page/p1604?clang=sv

1. Open a concept in Skosmos with a property that doesn't have a label defined
2. Look at apache error log

This is a minor bug that doesn't affect the output for the concept, but it does add unnecessary entries to the error log. For example, <http://www.yso.fi/onto/kauno-meta/facet> is not a property which would be shown for the user anyway.

joelit:
Fixes #813 by evaluating the result array for null or zero-length to get rid of undefined index errors for the array. Instead, if there's no label found for a property it returns null instead.

### DIFF
--- a/model/Concept.php
+++ b/model/Concept.php
@@ -514,7 +514,9 @@ class Concept extends VocabularyDataObject
                 // note that this imply that the property has an rdf:type declared for the query to work
                 if(!$proplabel) {
                     $envLangLabels = $this->model->getDefaultSparql()->queryLabel($longUri, $this->getEnvLang());
-                    $proplabel = ($envLangLabels)?$envLangLabels[$this->getEnvLang()]:$this->model->getDefaultSparql()->queryLabel($longUri, '')[''];
+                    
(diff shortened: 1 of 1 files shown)


kouralex:
Hey Joeli,
 
the proposed pull request introduces a viable resolution to the issue mentioned, however, I'd suggest you to change variable name  `$defaultGraphPropLabel` to `$defaultPropLabel` since it more accurately describes the variable.

Also, in case you haven't already noticed, the Travis build failed due to some unit tests testing the output of the method you changed. I'd appreciate if you could change the unit tests to follow the output of your changed method so that the tests would not fail (and after of which I could merge your request).
See [here](https://github.com/NatLibFi/Skosmos/wiki/Unit-and-integration-tests) for more information about unit/integration testing with your code and setup before committing.

Issue #836, opened 2019-01-09
Search result page generation is slow due to too many queries


osma:
## At which URL did you encounter the problem?
http://localhost/Skosmos/ysa/fi/search?q=kar&clang=fi

(and many other search result pages when there are more than a couple results)

## What steps will reproduce the problem?
1. Perform a search that matches at least 20 concepts, e.g. search for "kar" in YSA
2. Wait for the search results page

## What is the expected output? What do you see instead?

Expected to get the first result page in a reasonable time, say 1-2 seconds at most.
Instead it can take 8-10 seconds or sometimes more.

The problem seems to be that generating the page causes a large number of SPARQL queries to be submitted to the endpoint. Some of the queries are also repeated. For the YSA `kar` example, there more than 500 queries are performed before the page is ready.

A lot of these queries are lookups for property labels and subClassOf relationships of properties. These were added when fixing #669 (implemented in PR #671) which added lookups for label information in the default endpoint. These are performed unconditionally for every property of every concept shown on the search result page - even very common properties such as SKOS, DC and `rdf:type`. The results are not even cached, so when many concepts use the same property, there are more, identical queries being sent to the endpoint.

A few strategies for addressing the problem:

1. Avoid performing label and subClassOf relationship lookups for well-known properties, e.g. SKOS and DC/DCTerms properties. For most of these we already have labels available via gettext and custom subClassOf relationships should be unlikely.

2. Avoid repeating the same lookups/queries at least during the same page load. For example, property labels looked up from the default graph could be cached.

3. Cache the results of all SPARQL queries for a short period, for example using APCu cache. This could give an additional boost even when a SPARQL/HTTP cache such as Varnish is being used, as the HTTP queries don't have to be done at all. But the TTL should be short (e.g. 60 seconds, or even just a single page load) so it doesn't cause problems with stale data.

I think that fix 1 and possibly also 2 could be done as a simple bugfix that can be backported to the maintenance branch, but 3 needs additional testing and should wait for the next release. Solutions 2 and 3 are somewhat overlapping, perhaps both are not needed though 3 could have additional benefits beyond helping with this issue.

osma:
I implemented solutions 1 and 3 and tested their effect.
The test case is the URL above, i.e. searching YSA for `kar`.
I selected the fastest of 3+ runs.

Current `master`: 5.2 seconds (though often more like 6.5 seconds)
Solution 1 (skip well known): 2.9 seconds
[...]

PR #838, opened 2019-01-09
Cache SPARQL query results for 1 minute. 

osma:
Implements Solution 3 for #836


osma:
TBH I'm not sure this is a great idea after all. I think the other optimizations should be done first (some are already merged, #851 still needs to be sorted out) and then evaluate whether this approach is still necessary. Caching is always tricky to get right...

osma:
I think the other performance fixes are good enough so that this rather questionable fix isn't necessary. Closing the PR.

Issue #575, opened 2017-01-16
Support 304 Not Modified response


osma:
Once we have Last-Modified headers (see #574), we could also add support for the 304 Not Modified response which should make concept pages (and RDF dumps) very quick to serve when the client (e.g. a bot) already has a cached copy and uses the If-Modified-Since header.

osma:
This is now implemented for concept pages, thanks to #839. Opened #864 for RDF responses. Closing this one.

Issue #574, opened 2017-01-16
Support Last-Modified header


osma:
Skosmos does not currently set the Last-Modified HTTP header on responses. The data comes directly from the triple store and Skosmos is generally unaware of when it has last changed.

However, some vocabularies include dct:modified timestamps, either for the whole vocabulary (set on the skos:ConceptScheme resource) and/or individual concepts. Skosmos could make use of these timestamps to set Last-Modified headers, at least for concept pages (and RDF data for concepts) but possibly also other responses. For a concept page, the Last-Modified timestamp would primarily be based on the concept-level timestamp, but if not set, the concept scheme level timestamp could be used instead.

This would perhaps have to be enabled in a per-vocabulary setting. Not sure if there are any dangers in enabling it for all vocabularies. There would be some additional SPARQL queries, but they should be fast and easily cached.

kinow:
I this one, #574, and #575, might be done together.

I have a branch for the performance work around mapping properties, and thought it would be actually simple to try something on this one using that branch (because I had Docker + cache + config.ttl properly setup, etc).

Here's what I did:

1. Added a vocabulary config entry called "useModifiedDate"... not happy with the value... I guess someone could come up with a better name for this setting? I have the feeling that it might be either confusing or cause trouble in the future if we add more settings?
2. Added a method to `Concept` to return the modified date
3. Added some logic to the `WebController` where it returns concepts, to search for the vocab config setting (1.), and the concept's modified date (2.), then also check if the browser (or any other client, tested curl, and it worked fine) set the `If-Modified-Since` header.
4. Finally, if the header is present, and if the content has not changed, then return an empty 304. But if the header is not present, then (and this is important) set the `Last-Modified` header. When browsers receive the `Last-Modified` header, they send it back as `if-Modified-Since` in the next request, unless the user marked to disable cache or forced the reload of the page...

Screenshots below show the initial request receiving a 200 with the Last-Modified, then a subsequent request/response 304, being correctly displayed by the browser.

![screenshot from 2019-01-03 00-14-04](https://user-images.githubusercontent.com/304786/50590051-1ba03400-0eee-11e9-8037-b20da278bd1e.png)
![screenshot from 2019-01-03 00-16-52](https://user-images.githubusercontent.com/304786/50590050-1ba03400-0eee-11e9-808a-c13563e9b87a.png)
[...]

PR #839, opened 2019-01-10
Support last modified header

kinow:
For #575 and #574. Adds support to last modified header. Only works if the vocabulary configuration, in config.ttl, contains the `skosmos:useModifiedDate "true";`.

It first tries to load the modified from the concept. If this date is not found, then it tries to use the main concept scheme's modified.

If the modified date was found, then it tries to set the HTTP 304 header. This only happens when the `$_SERVER` variable has been populated by PHP&Browser. Otherwise it will simply put the date in the `Last-Modified` header. This gets picked up by the browser for the next requests.

WIP for now. Missing steps:

- There's one TODO right after it retrieves the modified date (either from Concept or ConceptScheme), to validate the date. That will be a simple pattern match against YYYY-mm-ddTHH:MM:SS or so (WDYT @osma ? Should we validate more, checking for the year or so?)
- It is clearly missing unit tests :grimacing: but they are coming. First had to do some development with local tests to refresh how the 304 header worked with browsers, and also to craft the code around EasyRDF classes.

Final note, it is being created for PHP 7.2+, with code hints everywhere. Which removes the need to test for things like null, instanceof, etc, in some cases. But happy to revert that part if necessary.

Cheers
Bruno

### DIFF
--- a/controller/Controller.php
+++ b/controller/Controller.php
@@ -141,8 +141,19 @@ class Controller
                 header('HTTP/1.0 304 Not Modified');
                 return true;
             }
-            header("Last-Modified: " . $modifiedDate->format('Y-m-d H:i:s'));
+            $this->sendHeader("Last-Modified: " . $modifiedDate->format('Y-m-d H:i:s'));
         }
         return false;
+    }
+
+    /**
+     * Sends HTTP headers. Simply calls PHP built-in header function. But being
+     * a function here, it can easily be tested/mocked.
+     *
+     * @param $header string header to be sent
+     */
+    protected function sendHeader($header): void
+    {
+        header($header);
     }
 }
--- a/controller/Controller.php
+++ b/controller/Controller.php
@@ -121,7 +121,7 @@ class Controller
 
     /**
      * If the $modifiedDate is a valid DateTime, and if the $_SERVER variable contains the right info, and
-     * if the $modifiedDate is more recent than the latest value in $_SERVER, then this function sets the
+     * if the $modifiedDate is not more recent than the latest value in $_SERVER, then this function sets the
      * HTTP 304 not modified and returns true..
      *
      * If the $modifiedDate is still valid, then it sets the Last-Modified header, to be used by the browser for
--- a/controller/Controller.php
+++ b/controller/Controller.php
@@ -136,14 +136,22 @@ class Controller
     protected function sendNotModifiedHeader($modifiedDate): bool
     {
         if ($modifiedDate) {
-            if (isset($_SERVER['HTTP_IF_MODIFIED_SINCE']) &&
-                strtotime($_SERVER['HTTP_IF_MODIFIED_SINCE']) >= $modifiedDate) {
-                header('HTTP/1.0 304 Not Modified');
+            $ifModifiedSince = $this->getIfModifiedSince();
+            $this->sendHeader("Last-Modified: " . $modifiedDate->format('Y-m-d H:i:s'));
+            if (!is_null($ifModifiedSince) && $ifModifiedSince >= $modifiedDate) {
+                $this->sendHeader("HTTP/1.0 304 Not Modified");
                 return true;
             }
-            $this->sendHeader("Last-Modified: " . $modifiedDate->format('Y-m-d H:i:s'));
         }
         return false;
+    }
+
+    /**
+     * @return DateTime|null a DateTime object if the value exists in the $_SERVER variable, null otherwise
+     */
+    protected function getIfModifiedSince()
+    {
+        return isset($_SERVER["HTTP_IF_MODIFIED_SINCE"]) ? strtotime($_SERVER["HTTP_IF_MODIFIED_SINCE"]) : null;
     }
 
     /**
@@ -152,7 +160,7 @@ class Controller
      *
      * @param $header string header to be sent
      */
-    protected function sendHeader($header): void
+    protected function sendHeader($header)
     {
         header($header);
     }
--- a/controller/Controller.php
+++ b/controller/Controller.php
@@ -118,4 +118,31 @@ class Controller
         header("Content-type: text/plain; charset=utf-8");
         echo "$code $status : $message";
     }
+
+    /**
+     * If the $modifiedDate is a valid DateTime, and if the $_SERVER variable contains the right info, and
+     * if the $modifiedDate is more recent than the latest value in $_SERVER, then this function sets the
+     * HTTP 304 not modified and returns true..
+     *
+     * If the $modifiedDate is still valid, then it sets the Last-Modified header, to be used by the browser for
+     * subsequent requests, and returns false.
+     *
+     * Otherwise, it returns false.
+     *
+     * @param DateTime $modifiedDate the last modified date to be compared against server's modified since information
+     * @return bool whether it sent the HTTP 304 not modified headers or not (useful for sending the response without
+     *              further actions)
+     */
+    protected function sendNotModifiedHeader($modifiedDate): bool
+    {
+        if ($modifiedDate) {
+            if (isset($_SERVER['HTTP_IF_MODIFIED_SINCE']) &&
+                strtotime($_SERVER['HTTP_IF_MODIFIED_SINCE']) >= $modifiedDate) {
+                header('HTTP/1.0 304 Not Modified');
+                return true;
+            }
+            header("Last-Modified: " . $modifiedDate->format('Y-m-d H:i:s'));
+        }
+        return false;
+    }
 }
--- a/controller/WebController.php
+++ b/controller/WebController.php
@@ -264,7 +264,6 @@ class WebController extends Controller
                 }
             }
         }
-        // TODO: validate date before sending it!
         return $modifiedDate;
     }
 
--- a/controller/WebController.php
+++ b/controller/WebController.php
@@ -248,6 +248,7 @@ class WebController extends Controller
 
     /**
      * @param Concept $concept
+     * @param Vocabulary $vocab
      * @return DateTime|null
      */
     protected function getModifiedDate(Concept $concept, Vocabulary $vocab)
--- a/controller/WebController.php
+++ b/controller/WebController.php
@@ -223,6 +223,14 @@ class WebController extends Controller
             $this->invokeGenericErrorPage($request);
             return;
         }
+        $useModifiedDate = $vocab->getConfig()->getUseModifiedDate();
+        if ($useModifiedDate) {
+            $modifiedDate = $this->getModifiedDate($results[0], $vocab);
+            // return if the controller sends the not modified header
+            if ($this->sendNotModifiedHeader($modifiedDate)) {
+                return;
+            }
+        }
         /** @var \Twig\Template $template */
         $template = (in_array('skos:Concept', $results[0]->getType()) || in_array('skos:ConceptScheme', $results[0]->getType())) ? $this->twig->loadTemplate('concept-info.twig') : $this->twig->loadTemplate('group-contents.twig');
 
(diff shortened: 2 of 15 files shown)


osma:
For date validation, I'm mainly concerned about the cases I know we already have in some data, e.g. wrongly formatted datestamps like `20181022` (no dashes) and nonsensical dates like `2018-00-00`. If your validation catches those it's fine. Also different precision levels should be supported - at least date only (`2018-10-22`) and datetime with second level precision (`2018-10-22T12:34:45`). You're parsing the date with the DateTime class which is probably fine for this purpose. Unit tests can be used to check important cases here.

As for the code hints, currently the unit tests pass with PHP 7.1 and 7.2 but fail on 7.0. I think 7.0 brought the type annotations and 7.1 added the nullable types - with the question mark - and those are a problem for 7.0. I'd like to still keep support for 7.0 as it's for example the PHP version that comes with Ubuntu 16.04. So it's fine to use basic type declarations, but I'd like to avoid the nullable types if possible so that the code still works on 7.0. Does this make sense from your perspective?

Your diff seems to include `config.inc` and `vocabularies.ttl` files, these were probably not intended?

kinow:
>For date validation, I'm mainly concerned about the cases I know we already have in some data, e.g. wrongly formatted datestamps like 20181022 (no dashes) and nonsensical dates like 2018-00-00. If your validation catches those it's fine. Also different precision levels should be supported - at least date only (2018-10-22) and datetime with second level precision (2018-10-22T12:34:45). You're parsing the date with the DateTime class which is probably fine for this purpose. Unit tests can be used to check important cases here.

I was more concerned if you were thinking about some validation for a date that is "too old", like 1800, or 1970 - which I don't know how I would test for that.

I wasn't concerned about the formatting, because it's mostly handled by EasyRDF. The only requirement, is the `dc:modified` to be a datetime literal (e.g. `dc:modified "20181022"^^xsd:dateTime`).

Assuming the data contains `xsd:dateTime`'s, EasyRDF appears to do a good job parsing the dates.

```turtle
test:cat a skos:Concept ;
	skos:prefLabel "Cat"@en;
	dc:modified "2018-12-13T06:28:14+00:00"^^xsd:dateTime .

test:dog a skos:Concept ;
    skos:prefLabel "Dog"@en .

test:owl a skos:Concept ;
	skos:prefLabel "Owl"@en;
	dc:modified "20181022"^^xsd:dateTime .

test:lizard a skos:Concept ;
	skos:prefLabel "Lizard"@en;
	dc:modified "2018-00-00"^^xsd:dateTime .

test:parrot a skos:Concept ;
	skos:prefLabel "Parrot"@en;
	dc:modified "2018-10-22"^^xsd:dateTime .

test:macaw a skos:Concept ;
	skos:prefLabel "Macaw"@en;
	dc:modified "2018-10-22T12:34:45"^^xsd:dateTime .

test:sloth a skos:Concept ;
	skos:prefLabel "Sloth"@en;
	dc:modified "2018-10-22T12:34:45+05:30"^^xsd:dateTime .
```

In order:

```
# cat
class DateTime#149 (3) {
  public $date =>
  string(26) "2018-12-13 06:28:14.000000"
  public $timezone_type =>
  int(1)
  public $timezone =>
  string(6) "+00:00"
}

# dog
NULL

# owl
class DateTime#149 (3) {
  public $date =>
  string(26) "2018-10-22 00:00:00.000000"
  public $timezone_type =>
  int(3)
  public $timezone =>
  string(3) "UTC"
}

# lizard
class DateTime#149 (3) {
  public $date =>
  string(26) "2017-11-30 00:00:00.000000"
  public $timezone_type =>
  int(3)
  public $timezone =>
  string(3) "UTC"
}

# parrot
class DateTime#149 (3) {
  public $date =>
  string(26) "2018-10-22 00:00:00.000000"
  public $timezone_type =>
  int(3)
  public $timezone =>
  string(3) "UTC"
}

# macaw
class DateTime#149 (3) {
  public $date =>
  string(26) "2018-10-22 12:34:45.000000"
  public $timezone_type =>
  int(3)
  public $timezone =>
  string(3) "UTC"
}

# sloth
class DateTime#149 (3) {
  public $date =>
  string(26) "2018-10-22 12:34:45.000000"
  public $timezone_type =>
  int(1)
  public $timezone =>
  string(6) "+05:30"
}
```

So I am assuming even though the date format may not be consistent, users will have passed the data through some parser/validator that will have said that the date time was indeed OK according to the [XML Schema](https://www.w3.org/TR/xmlschema-2/#dateTime).

I added a validator Friday (my last commit of the day, which also included the two extra files), but then now realized it won't be necessary.

I tried the date `2018-45-45`. PHP's `DateTime:createFromFormat` accepts without error/warnings, but the developer has to remember to call `DateTime::getLastErrors()`. However, I tried the following reading from the data.

```
test:mink a skos:Concept ;
	skos:prefLabel "Mink"@en;
	dc:modified "2018-45-45"^^xsd:dateTime .
```

And EasyRDF somewhere validates it and throws an error (same error if I try `20184545`, so the error is not due to the `-`):

```
Exception : DateTime::__construct(): Failed to parse time string (2018-45-45) at position 6 (5): Unexpected character
 /home/kinow/Development/php/workspace/Skosmos/vendor/easyrdf/easyrdf/lib/Literal/Date.php:99
 /home/kinow/Development/php/workspace/Skosmos/model/Concept.php:668
 /home/kinow/Development/php/workspace/Skosmos/tests/ConceptTest.php:489
```

So I think it's safe for now to leave validation to EasyRDF, and add more later if necessary? What do you think?

One thing that still concerns me are the time zones. I am not sure how to handle them here, nor how to test. But will try to simulate a case of server in UTC, data in some time zone, and client in another time zone (not sure if a unit test will handle it, might use Docker and test locally).

>As for the code hints, currently the unit tests pass with PHP 7.1 and 7.2 but fail on 7.0. I think 7.0 brought the type annotations and 7.1 added the nullable types - with the question mark - and those are a problem for 7.0. I'd like to still keep support for 7.0 as it's for example the PHP version that comes with Ubuntu 16.04. So it's fine to use basic type declarations, but I'd like to avoid the nullable types if possible so that the code still works on 7.0. Does this make sense from your perspective?

That makes sense. I've edited the commits to revert the changes. Also set the IDE compatibility to 7.0, so hopefully won't have further issues with that :slightly_smiling_face: 

>Your diff seems to include config.inc and vocabularies.ttl files, these were probably not intended?

Oh, you are right. Last commit was from another computer that still had these files. Just amended the commit.

Thanks!

osma:
Oh one more thing: it would make sense to apply the same Last-Modified and If-Modified-Since / 304 logic to the REST API concept-specific `data` method, because it returns the same information that was used to produce the equivalent concept page. This way also the RDF Linked Data exposed by Skosmos would be cacheable!

kinow:
> Good point. How about:
>
>    If the timestamp has timezone information, use that.
>    Otherwise, interpret the timestamp in the data assuming the time zone of the server.
>
>In both cases, the actual generated timestamp could always be sent as UTC to the client.

Works for me. Will implement it that way and try to test it.

>Oh one more thing: it would make sense to apply the same Last-Modified and If-Modified-Since / 304 logic to the REST API concept-specific data method, because it returns the same information that was used to produce the equivalent concept page. This way also the RDF Linked Data exposed by Skosmos would be cacheable!

Agreed! And speed up even more the page loading. What about other REST clients (e.g. `curl`, `wget`, https://github.com/NatLibFi/Skosmos-client)? Wouldn't it limit, or at least require that they also implement caching? It's quite easy to implement it now in the RESTController, but just not too sure if it wouldn't cause issues to clients not going through a browser. What do you think @osma ?

kinow:
Rebased, then also did a backward rebase, where I dropped the commit with validation.

Added a few more commits, separated more or less by feature (so that they can be edited/dropped later if necessary as well).

So far unit tests for Model/Vocabulary/VocabularyConfig, will complete coverage I believe.

Final part is to add some unit tests or functional tests for the controllers, and also cache results for RESTController. Once this is done, this pull request should be ready for review! :tada:

kinow:
Waiting for Scrutinizer and Travis. If there are no errors, it should be almost ready for the review @osma .

I got stuck with the last part, about the `RESTController`... I started implementing the same logic over its `data` method... but the problem is that in `RESTController`, I do not have the `Concept` object... instead it does get a `Concept` object, but that happens in another method (`getRDF`). And that method serializes and return only the string back.

https://github.com/NatLibFi/Skosmos/blob/9de5b86a7322c8d787c8062f0a2a428c9fac1a2c/controller/RestController.php#L618

I then tried to move the `if/else` that retrieves the `$serializer` to another function. This way I would be able to use the `queryConceptInfoGraph` in the controller, then call the `$model->getSerializer()`, but only when returning the XML... for the 304 HTTP response, I would inspect the graph and try to return the date.

Issue is that for a `$format`, it creates the right serializer **and** also creates a string with the format name, used when serializing... so it was starting to look dangerous the amount of changes I started doing for that.

Would you have any suggestions??? :-) the logic for the HTTP 304 is done, and is not very hard to reproduce it... issue is only how to get the modified date from the RDF... I thought about parsing it again into a Concept... but it sounds wrong :disappointed: 

Cheers
Bruno

kinow:
Oh, this pull request now includes:

- Change for 304
- Changes to add a few more unit tests to either existing or new classes
- Changes to add php type hinting (up to PHP 7.0)
- Changes to allow tests running from command line (i.e. `phpunit`, or `phpunit --filter=...`), and also from the IDE (e.g. PyCharm... it uses the phpunit loader, but for some reason the relative directories used were causing issues).
- A few phpdocs that were missing

osma:
As the name implies it's a dev server, so not used very heavily. It's probably difficult to estimate how much work is saved by this but we can try to deduce something. In any case I will merge this now and let's see what happens!

osma:
Enabled this for YSA, YSO, YSO Places, Lapponica, SLM, IPTC. This should cover all the cases listed in https://github.com/NatLibFi/Skosmos/issues/574#issuecomment-452316697

The good news:

* Last-Modified headers seem to be added to all concept pages! And my browser is now sending If-Modified-Since requests accordingly.

The bad news:

1. I couldn't find any evidence of 304 responses. My browser is now sending If-Modified-Since requests but the server seems to be responding with normal 200 responses. This is complicated a bit because there's a Varnish cache in between, but I tried to do manual cache invalidation etc. to prevent it from interfering and still I couldn't see a single 304 response in the Apache access.log (except for static content like images where Apache itself gives 304s). Are 304 responses working for you @kinow?
2. It broke for IPTC (which doesn't have any timestamps at all, not even on the concept scheme). The concept pages do not display at all and I got this in the Apache error log:

[Fri Jan 18 14:04:36 2019] [error] [client 128.214.190.131] PHP Fatal error:  Uncaught Error: Call to a member function getValue() on null in /var/www/dev.finto.fi/controller/WebController.php:263\nStack trace:\n#0 /var/www/dev.finto.fi/controller/WebController.php(228): WebController->getModifiedDate(Object(Concept), Object(Vocabulary))\n#1 /var/www/dev.finto.fi/index.php(87): WebController->invokeVocabularyConcept(Object(Request))\n#2 {main}\n  thrown in /var/www/dev.finto.fi/controller/WebController.php on line 263

I disabled the useModifiedDate setting for IPTC for now.

kinow:
@osma !

Hooray for the good news!!!

Boo boo for the bad news :disappointed_relieved: , time to investigate ! (And apologize beforehand).

>I couldn't find any evidence of 304 responses. My browser is now sending If-Modified-Since requests but the server seems to be responding with normal 200 responses. This is complicated a bit because there's a Varnish cache in between, but I tried to do manual cache invalidation etc. to prevent it from interfering and still I couldn't see a single 304 response in the Apache access.log (except for static content like images where Apache itself gives 304s). Are 304 responses working for you @kinow?

I just tested dev.finto.fi, using "abbots" concept, from YSO. The first call gave me a 200. Then the next call returned 304.

image

To test that, I loaded the main page of dev.finto.fi on Firefox. Opened the YSO vocabulary, and then right clicked on the screen and Inspect Element. Then opened the Network tab. Cleared all previous requests, and selected to show only HTML. Finally, clicked on abbort in the hierarchy side bar.

It appeared as a 200 HTML response. Then clicking more times, it always appeared as 304. Could you check if it makes sense, and if it appears to be working OK indeed, please @osma ?

In the meantime let me check this second issue...

kinow:
>It broke for IPTC (which doesn't have any timestamps at all, not even on the concept scheme). The concept pages do not display at all and I got this in the Apache error log:

Addressed this one in #840

PR #841, opened 2019-01-20
Remove duplicated array index

kinow:
This had been bothering, but I never found time to check which value was actually used. I suspected the last value set, but thought it was still better to check.

Screenshot from debugger attached to the docker container running latest version. It's possible to see that rdfs:label is not used, so it should be safe to remove that entry, and leave the `dct:title` only.

screenshot from 2019-01-20 16-59-13

Just to avoid possible issues when other developers could try to update these keys/values.

Cheers
Bruno


osma:
Good catch, thanks!

PR #846, opened 2019-01-31
Fix performance issue when displaying concept mapping properties

kinow:
Call concept mapping endpoint via Ajax to populate the page instead of using PHP to create greedily the complete page content.

### DIFF
--- a/controller/RestController.php
+++ b/controller/RestController.php
@@ -3,7 +3,7 @@
 /**
  * RestController is responsible for handling all the requests directed to the /rest address.
  */
-class RestController extends Controller
+class RestController extends WebController
 {
     /* supported MIME types that can be used to return RDF data */
     const SUPPORTED_FORMATS = 'application/rdf+xml text/turtle application/ld+json application/json application/marcxml+xml';
@@ -657,7 +657,8 @@ class RestController extends Controller
         $ret = [];
         foreach ($concept->getMappingProperties() as $mappingProperty) {
             foreach ($mappingProperty->getValues() as $mappingPropertyValue) {
-                $ret[] = $mappingPropertyValue->asJskos($queryExVocabs, $request->getLang(), $request->getContentLang());
+                $hrefLink = $this->linkUrlFilter($mappingPropertyValue->getUri(), $mappingPropertyValue->getExVocab(), $request->getLang(), 'page', $request->getContentLang());
+                $ret[] = $mappingPropertyValue->asJskos($queryExVocabs, $request->getLang(), $hrefLink);
             }
         }
 
--- a/controller/RestController.php
+++ b/controller/RestController.php
@@ -640,9 +640,8 @@ class RestController extends Controller
     {
         $vocab = $request->getVocab();
 
-        if ($request->getUri()) {
-            $uri = $request->getUri();
-        } else {
+        $uri = $request->getUri();
+        if (!$uri) {
             return $this->returnError(400, 'Bad Request', "uri parameter missing");
         }
 
@@ -658,7 +657,7 @@ class RestController extends Controller
         $ret = [];
         foreach ($concept->getMappingProperties() as $mappingProperty) {
             foreach ($mappingProperty->getValues() as $mappingPropertyValue) {
-                $ret[] = $mappingPropertyValue->asJskos($queryExVocabs);
+                $ret[] = $mappingPropertyValue->asJskos($queryExVocabs, $request->getLang(), $request->getContentLang());
             }
         }
 
--- a/controller/RestController.php
+++ b/controller/RestController.php
@@ -654,13 +654,18 @@ class RestController extends WebController
 
         $concept = $results[0];
 
-        $ret = [];
+        $mappings = [];
         foreach ($concept->getMappingProperties() as $mappingProperty) {
             foreach ($mappingProperty->getValues() as $mappingPropertyValue) {
                 $hrefLink = $this->linkUrlFilter($mappingPropertyValue->getUri(), $mappingPropertyValue->getExVocab(), $request->getLang(), 'page', $request->getContentLang());
-                $ret[] = $mappingPropertyValue->asJskos($queryExVocabs, $request->getLang(), $hrefLink);
+                $mappings[] = $mappingPropertyValue->asJskos($queryExVocabs, $request->getLang(), $hrefLink);
             }
         }
+
+        $ret = array(
+            'mappings' => $mappings,
+            'graph' => $concept->dumpJsonLd()
+        );
 
         return $this->returnJson($ret);
     }
--- a/controller/RestController.php
+++ b/controller/RestController.php
@@ -639,6 +639,7 @@ class RestController extends Controller
      */
     public function mappings(Request $request)
     {
+        $this->setLanguageProperties($request->getLang());
         $vocab = $request->getVocab();
 
         $uri = $request->getUri();
--- a/controller/RestController.php
+++ b/controller/RestController.php
@@ -3,7 +3,7 @@
 /**
  * RestController is responsible for handling all the requests directed to the /rest address.
  */
-class RestController extends WebController
+class RestController extends Controller
 {
     /* supported MIME types that can be used to return RDF data */
     const SUPPORTED_FORMATS = 'application/rdf+xml text/turtle application/ld+json application/json application/marcxml+xml';
@@ -635,6 +635,7 @@ class RestController extends WebController
      * Get the mappings associated with a concept, enriched with labels and notations.
      * Returns a JSKOS-compatible JSON object.
      * @param Request $request
+     * @throws Exception if the vocabulary ID is not found in configuration
      */
     public function mappings(Request $request)
     {
--- a/controller/Controller.php
+++ b/controller/Controller.php
@@ -151,7 +151,7 @@ class Controller
         $localname = $vocab->getLocalName($uri);
         if ($localname !== $uri && $localname === urlencode($localname)) {
             // check that the prefix stripping worked, and there are no problematic chars in localname
-            $paramstr = sizeof($params) > 0 ? '?' . http_build_query($params) : '';
+            $paramstr = count($params) > 0 ? '?' . http_build_query($params) : '';
             if ($type && $type !== '' && $type !== 'vocab' && !($localname === '' && $type === 'page')) {
                 return "$vocid/$lang/$type/$localname" . $paramstr;
             }
--- a/controller/Controller.php
+++ b/controller/Controller.php
@@ -116,6 +116,56 @@ class Controller
     }
 
     /**
+     * Creates Skosmos links from uris.
+     * @param string $uri
+     * @param Vocabulary $vocab
+     * @param string $lang
+     * @param string $type
+     * @param string $clang content
+     * @param string $term
+     * @throws Exception if the vocabulary ID is not found in configuration
+     * @return string containing the Skosmos link
+     */
+    public function linkUrlFilter($uri, $vocab, $lang, $type = 'page', $clang = null, $term = null) {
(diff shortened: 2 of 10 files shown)


kinow:
Rebased onto master and fixed conflicts. Hopefully Travis-CI will be still OK with the change :+1:

controller/RestController.php
```diff
@@ -3,7 +3,7 @@
 /**
  * RestController is responsible for handling all the requests directed to the /rest address.
  */
-class RestController extends Controller
+class RestController extends WebController
```
osma: Why this change? If there is something in WebController that RestController needs, it can be moved to the Controller base class, but I wouldn't make one inherit the other, they are intended to be sibling classes with different responsibilities.

osma:
Travis tests were failing like this:

```
There was 1 failure:
1) ConceptMappingPropertyValueTest::testAsJskos
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     'type' => Array (...)
     'toScheme' => Array (...)
     'from' => Array (...)
     'to' => Array (...)
+    'uri' => 'http://www.skosmos.skos/mapping/m1'
+    'notation' => null
+    'prefLabel' => 'Eel'
+    'description' => 'Exactly matching concepts in ...ulary.'
+    'hrefLink' => null
+    'lang' => 'en'
+    'vocabName' => 'Test ontology'
+    'typeLabel' => 'Exactly matching concepts'
/home/travis/build/NatLibFi/Skosmos/tests/ConceptMappingPropertyValueTest.php:203
```

It seems that ConceptMappingPropertyValueTest just has to be updated to match what the code does.

osma:
Tested this locally. Some observations:

1. The dynamic loading works fine when you do a full page load. But if you navigate to a concept page using the sidebar (alphabetical index, hierarchy, group index or new concepts list), then the mappings are not loaded at all. I bet this happens because when using the sidebar, the new concept page is loaded using an AJAX-style request, not a full page load.
2. The field labels ("EXACTLY MATCHING CONCEPTS" etc) and their descriptions (shown when hovering/tapping) are always in English, even when I'm using Finnish UI and content languages. They should be displayed in the UI language, like the rest of the field labels.
3. On pages for concepts which don't have any mappings, there is still a grey bar (div class=concept-appendix in the HTML source) displayed under the concept information. Before this PR, nothing was shown there.

osma:
I think that's enough review for this PR for now! Great work, it's advancing well though some things need to be fixed before this can be merged.

controller/RestController.php
kinow: This was just to use the `WebController::linkUrlFilter` method. Moved `linkUrlFilter` to `Controller`. And will keep in mind the separation of concerns for both controllers and try to keep them well separated :)

kinow:
Rebased, updated, and tests passing now.

Good feedback on the Ajax responses! Hadn't looked at different ways the concept page could be used. Might be the case that the JavaScript used just needs to be loaded somewhere else.

Will test it during this (very rainy so far) weekend. Thanks!

kinow:
>The dynamic loading works fine when you do a full page load. But if you navigate to a concept page using the sidebar (alphabetical index, hierarchy, group index or new concepts list), then the mappings are not loaded at all. I bet this happens because when using the sidebar, the new concept page is loaded using an AJAX-style request, not a full page load.

Fixed in 5064253362f2a5c93f09ebaf53ddce52fc4a6493. Interesting, I thought the whole page was being loaded when you clicked on the hierarchy node. Found where that magic takes place in `docready.js` (`$(document).on('click', 'concept-hierarchy a')...`), and changed the code to match what's done when the concept page is loaded.

kinow:
>The field labels ("EXACTLY MATCHING CONCEPTS" etc) and their descriptions (shown when hovering/tapping) are always in English, even when I'm using Finnish UI and content languages. They should be displayed in the UI language, like the rest of the field labels.

That's weird. Not sure what I changed that caused this regression. But I had a look at Finto, and the hover notes also appear always in English I think? https://finto.fi/yso/en/page/?clang=fi&uri=p13015

Or is it something else that needs to appear in Finnish instead of English. I mean the text __EXACTLY MATCHING CONCEPTS__ and also what appears when I hover the mouse over that text.

kinow:
>On pages for concepts which don't have any mappings, there is still a grey bar (div class=concept-appendix in the HTML source) displayed under the concept information. Before this PR, nothing was shown there.

Oh, I see what you mean. Went to YSO groups, and clicked on 02. Philosophy and saw that spinner working its way, and then leaving the empty grey bar. In the beginning, the component is hidden

image

kinow:
Whew, addressed the points from the feedback I think. Waiting for Travis CI now, but if no issues found there, it should be ready for review again :tada:

kinow:
Rebased :+1:

osma:
Sorry @kinow, this has been waiting for another review way too long.

I fixed a minor merge conflict in `concept-shared.twig` and re-tested. 

The AJAX style loading and extra grey bar issues are now solved, great!

There are still some issues with language support.

1. It appears that the `lang` parameter given to the `mappings` REST method is based on the UI language, not the content language. I think the content language should be used instead, so that the returned labels will be in the same language as the labels in the vocabulary. For example, if I'm browsing YSO Places with the UI language set to Finnish and content language set to English, and browse to the concept Africa (http://localhost/Skosmos/yso-paikat/fi/page/?clang=en&uri=p94080), the Wikidata link will be shown as "Afrikka" (in Finnish) instead of English as it should be.

2. The property labels "CLOSELY MATCHING CONCEPTS" etc. are still always displayed in English. The `gettext` calls made in ConceptMappingPropertyValue.php have no effect, probably because when they are called starting from RestController, the setLanguageProperties method is never called and so the locale settings are not in place. Even if they were, it's not clear how the `mappings` method would know what language to use.

I discussed the second problem with @kouralex and we came to the conclusion that it would be better to include all the possible property labels (and help texts / descriptions) in the `property-mappings-template` at the end of `concept-shared.twig`. There are only seven of them (see `$MAPPING_PROPERTIES` in Concept.php) so it would be possible to provide all of them in the template, pre-translated on the PHP/Twig side, and just pick the right one based on the property URI. This way some of the extra fields added to the JSKOS response could be dropped as well. What do you think @kinow?

kinow:
Hi @osma ! Thanks for fixing the merge conflicts.

> 1. It appears that the lang parameter given to the mappings REST method is based on the UI language, not the content language. I think the content language should be used instead, so that the returned labels will be in the same language as the labels in the vocabulary (...)

Makes sense to me. Without looking at the code, I think we already have access to the vocabulary & its configuration and related objects. So it should be easily available somewhere I think :crossed_fingers: 

> 2. The property labels "CLOSELY MATCHING CONCEPTS" etc. are still always displayed in English.  (...)
>
>I discussed the second problem with @kouralex and we came to the conclusion that it would be better to include all the possible property labels (and help texts / descriptions) in the property-mappings-template at the end of concept-shared.twig. There are only seven of them (see $MAPPING_PROPERTIES in Concept.php) so it would be possible to provide all of them in the template, pre-translated on the PHP/Twig side, and just pick the right one based on the property URI. This way some of the extra fields added to the JSKOS response could be dropped as well. What do you think @kinow?

Sounds like a good workaround. Probably not likely to increase the number of property labels, so that should do.

kinow:
Let me know you are planning on updating the PR with the workaround. Otherwise I can give it a try in the next days/weeks.

osma:
I fixed the first issue, it was a very simple fix (use `clang` instead of `lang`) though had to be applied in a few different places.

For the second one, I'd appreciate if you could take a look @kinow and update the PR.

We're planning a Skosmos 2.2 release (finally!), ideally this week, not sure if this can make it though.

osma:
@kinow have you had a chance to look at this?

kinow:
Hadn't had time yet @osma, sorry. But next Monday is last day before 2 weeks holidays :-) will have time to finally sync my git working copy and check out all the changes in Skosmos, and also finish this PR :+1:

osma:
Thanks @kinow for the update, no worries, take your time! And happy holidays!

kinow:
I think there were some merging conflicts that introduced more commits in this PR? I'm rebasing locally, and looking at the pull request. But there are now commits for ansible files in this PR too...

kinow:
Rebased again, hopefully didn't miss any commits from `master`.

kinow:
Sonar, Scrutinzer, and Travis looking good. Codacy issues IMHO are OK, as I'm using `gettext` which is used in other parts. There is one method with cyclomatic complexity of 14 while the limit is 10. Happy to split it if others prefer.

Now will spend some time refreshing memory on the issue, then re-read @osma 's recent comments and catch up on the existing work :+1:

kinow:
Sorry the delay @osma. Added `$this->setLanguageProperties($request->getLang());` to the `RestController::mappings` method. Can't tell what's written, but hoping it's the correct translation now :-)

image

osma:
Thanks @kinow! Just came back from my vacation, I will check this out soon!

osma:
I had a look. The property labels are now translated, but there are still issues - the translation seems to follow the *content language*, not the *UI language* like other property labels. So the UI is a bit inconsistent.

Consider this: (English UI, Finnish content language)
mappings-english-ui-finnish-content

Or this: (Finnish UI, English content language)
mappings-finnish-ui-english-content

In both cases the labels for the mapping properties are displayed in a different language than the rest of the property labels for the concept.

The `mappings` REST method is passed a single `lang` attribute (based on the content language of the current page), and this language is currently used to select both the labels of the matching concepts and the property labels. But to solve the above case, the method would need to know *both* the UI language and the content language, and use the UI language for the property labels and the content language for everything else. The only way I see to implement this would be to pass it two parameters: `lang` (UI language) and `clang` (content language), just like the parameters for pages invoked by WebController. WDYT?

(As an aside, I noticed that the `mappings` REST method is not documented in the [swagger.json](https://github.com/NatLibFi/Skosmos/blob/master/swagger.json) spec file. This is not the fault of this PR, as the method was added in PR #803, but in any case it would be good to add the method signature so that it becomes visible in interactive API documentation - perhaps as a separate PR after this one is done)

kinow:
Hi @osma 

Oh, sorry about that. I should have done a better testing. Implemented your suggestion and I think it's working correctly.

I tested using the Finto installation, using English as interface language and content. Then switching content language to Finnish. Then switched the interface language to Finnish, and finally switched content from Finnish to English.

I had similar output in both — the difference was mainly some words/links were different, but I suspect that's just old data (e.g. instead of "Library of Congress Subject Headings", my YSO has "id.loc.gov"). 

PR rebased as well. Do you mind having another go at testing it, please?

Thanks!

osma:
Thanks @kinow! Much better now.
There was still the problem that the `clang` parameter for the `mappings` REST query wasn't actually used. I fixed that in the above commit.

kinow:
> Thanks @kinow! Much better now.
> There was still the problem that the `clang` parameter for the `mappings` REST query wasn't actually used. I fixed that in the above commit.

Thanks @osma! It appeared to have worked before, funny.

osma:
There were some back-and-forth changes recently as we considered adding a `clang` parameter to the REST API - see #882 and #900. So if you based your fork somewhere in between the time these PRs got merged, the `clang` worked differently.

The new situation is different, as we genuinely need both `lang` and `clang` for the mappings method - `lang` for the property labels and `clang` for the labels of the mapping targets.

osma:
Merged, thanks a lot @kinow! Now we have to see how this affects the plugins/widgets and if we need to adjust the callback mechanisms in some cases. For example the Wikipedia widget used on YSO Places needs the graph of triples from Wikidata, which now arrives later than it used to.

osma:
Whoa, the Wikipedia widget seems to work, so you got the callbacks right too! Excellent!

PR #866, opened 2019-04-17
Rename docker-compose created container from "web" to "skosmos"

kinow:
There was a note in the Docker installation documentation about renaming it. I'd agree it makes more sense, so prepared this PR.

Will update the Wiki afterwards if approved :+1:


osma:
Looks good, thanks!

kinow:
Thanks @osma! Wiki updated https://github.com/NatLibFi/Skosmos/wiki/Install-Skosmos-with-Fuseki-in-Docker/_compare/f7eabdf28f62d4a77774dd2e03385a6aab39216c

Issue #862, opened 2019-04-09
Swagger/OpenAPI documentation for mappings method


joelit:
Currently, we have concept-specific methods for most relations presented in the YSO data model in the Skosmos API. We're still missing a way to query skos:exactMatches and skos:closeMatches.

kouralex:
There exists `mappings` REST method for asking all kinds of relations. It is, however, missing documentation.

osma:
Changed the title to better reflect the issue.

PR #869, opened 2019-04-24
Add 304 to rest controller

kinow:
WIP pull request! Will add another commit to update the `swagger.json` after the initial review. Notes taken during code writing below:

In methods for resources like /vocabularies, which return collections that do not have a timestamp, before the collection is retrieved are now doing the checks for git/config modified dates, and using them if request contains the right headers.

I used https://petstore.swagger.io/#/ with the swagger.json from https://raw.githubusercontent.com/NatLibFi/Skosmos/master/swagger.json, going from the top to bottom. So the comments below should reflect what I found while progressing through the methods.

The code to return the 304 header was added after validation methods. Except validation that required
performing a query (otherwise returning 304 after that would be pointless performance-wise I think).

Manual tests were performed while implementing the changes with the following commands:

```
# First request before the change, with response below (NB date is in the future!)
$ curl -I --header 'If-Modified-Since: 2020-04-29 07:39:48' http://localhost:8000/rest/v1/yso/
HTTP/1.1 200 OK
Date: Wed, 24 Apr 2019 08:59:09 GMT
Server: Apache/2.4.25 (Debian)
X-Powered-By: PHP/7.0.33
Access-Control-Allow-Origin: *
Vary: Accept
Content-Type: application/json; charset=utf-8

# Second request after implementing change
$ curl -I --header 'If-Modified-Since: 2020-04-29 07:39:48' http://localhost:8000/rest/v1/yso/
HTTP/1.0 304 Not Modified
Date: Wed, 24 Apr 2019 09:01:05 GMT
Server: Apache/2.4.25 (Debian)
Connection: close


# Last request where the date is in the past (meaning the resource was modified after that, so no 304 expected)
HTTP/1.1 200 OK
Date: Wed, 24 Apr 2019 09:01:08 GMT
Server: Apache/2.4.25 (Debian)
X-Powered-By: PHP/7.0.33
Access-Control-Allow-Origin: *
Last-Modified: 2019-04-24 07:39:48
Vary: Accept
Content-Type: application/json; charset=utf-8

```

## First modification: move methods related to modified date to Controller

The first commit moves the methods related to modified dates to the base Controller class.

## Second modification: per vocabulary setting to enable/disable modified date

In the concept method for WebController, we check if the Vocabulary has the setting `skosmos:useModifiedDate`, but there is no global setting to enable / disable the modified date globally. I thought we could have a global one and also a per vocabulary setting... but that would make troubleshooting it harder... in the end I thought the easiest approach would be to simply make `skosmos:modifiedDate` a global configuration. What do you think?

## Third modification: method signature for getModifiedDate

`getModifiedDate` was expecting a Concept. So now it allows this concept to be null. This way it can
still be used by methods with a Concept, but otherwise simply skip that value if it was not provided.

It was also expecting a Vocabulary in the method. But then I realized it was not necessary, as the
Concept contains already a getVocab() to retrieve the Concept's Vocabulary. It won't affect the existing
concept method in WebController, as the Concept is obtained in the first place by calling a method on the Vocabulary (IOW, the Concept->getVocab() will return the same as before).

## Fourth modification: realized I was always doing:

```php
        $useModifiedDate = $this->model->getConfig()->getUseModifiedDate();
        if ($useModifiedDate) {
            $modifiedDate = $this->getModifiedDate(null);
            // return if the controller sends the not modified header
            if ($this->sendNotModifiedHeader($modifiedDate)) {
                return;
            }
        }
```

So created the helper method `notModified(Concept)` in `Controller`, and now the same code is simplified to:

```php
        if ($this->notModified(null)) {
            return;
        }
```

## Methods that were not modified

I think for some methods, it may not make sense to return not modified, for some content
that the client may not have cached:

- data: it looks strange to return 304 for a url that may start the download of a file... what do you think?
- search: I have never seen a search result returning 304 I think... what do you think?

---

Q: I realized that I was not using the `Concept` in the REST methods related to concepts... but the reason is that the methods used by those REST methods are querying single string values (e.g. label), never returning a complete Concept. I could alter each query to return the `dc:modified` too, change the method to return... an array? Then use the `dc:modified` value. Is that the right direction?

### DIFF
--- a/controller/Controller.php
+++ b/controller/Controller.php
@@ -183,6 +183,17 @@ class Controller
         echo "$code $status : $message";
     }
 
+    protected function notModified(Concept $concept = null)
+    {
+        $notModified = false;
+        $useModifiedDate = $this->model->getConfig()->getUseModifiedDate();
+        if ($useModifiedDate) {
+            $modifiedDate = $this->getModifiedDate($concept);
+            $notModified = $this->sendNotModifiedHeader($modifiedDate);
+        }
+        return $notModified;
+    }
+
     /**
      * Return the modified date. First try to get the modified date from the concept. If found, return this
      * date.
--- a/controller/Controller.php
+++ b/controller/Controller.php
@@ -186,8 +186,7 @@ class Controller
     protected function notModified(Modifiable $modifiable)
     {
         $notModified = false;
-        $useModifiedDate = $this->model->getConfig()->getUseModifiedDate();
-        if ($useModifiedDate) {
+        if ($modifiable->getUseModifiedDate()) {
             $modifiedDate = $this->getModifiedDate($modifiable);
             $notModified = $this->sendNotModifiedHeader($modifiedDate);
         }
--- a/controller/Controller.php
+++ b/controller/Controller.php
@@ -6,6 +6,12 @@
 class Controller
 {
     /**
+     * How long to store retrieved disk configuration for HTTP 304 header
+     * from git information.
+     */
+    const GIT_MODIFIED_CONFIG_TTL = 600; // 10 minutes
+
+    /**
      * The controller has to know the model to access the data stored there.
      * @var Model $model contains the Model object.
      */
@@ -175,6 +181,124 @@ class Controller
         header("HTTP/1.0 $code $status");
         header("Content-type: text/plain; charset=utf-8");
         echo "$code $status : $message";
+    }
+
+    /**
+     * Return the modified date. First try to get the modified date from the concept. If found, return this
+     * date.
+     *
+     * Then try to get the modified date from the vocabulary's main concept scheme. If found, then return this
+     * date.
+     *
+     * Finally, if no date found, return null.
+     *
+     * @param Concept $concept
+     * @param Vocabulary $vocab
+     * @return DateTime|null
+     */
+    protected function getModifiedDate(Concept $concept, Vocabulary $vocab)
+    {
+        $modifiedDate = null;
+
+        $conceptModifiedDate = $this->getConceptModifiedDate($concept, $vocab);
+        $gitModifiedDate = $this->getGitModifiedDate();
+        $configModifiedDate = $this->getConfigModifiedDate();
+
+        // max with an empty list raises an error and returns bool
+        if ($conceptModifiedDate || $gitModifiedDate || $configModifiedDate) {
+            $modifiedDate = max($conceptModifiedDate, $gitModifiedDate, $configModifiedDate);
+        }
+        return $modifiedDate;
+    }
+
+    /**
+     * Get the concept modified date. Returns the concept modified date if available. Otherwise,
+     * reverts to the modified date of the default concept scheme if available. Lastly, if it could
+     * not get the modified date from the concept nor from the concept scheme, it returns null.
+     *
+     * @param Concept $concept concept used to retrieve modified date
+     * @param Vocabulary $vocab vocabulary used to retrieve modified date
+     * @return DateTime|null|string
+     */
+    protected function getConceptModifiedDate(Concept $concept, Vocabulary $vocab)
+    {
+        $modifiedDate = $concept->getModifiedDate();
+        if (!$modifiedDate) {
+            $conceptSchemeURI = $vocab->getDefaultConceptScheme();
+            if ($conceptSchemeURI) {
+                $conceptSchemeGraph = $vocab->getConceptScheme($conceptSchemeURI);
+                if (!$conceptSchemeGraph->isEmpty()) {
+                    $literal = $conceptSchemeGraph->getLiteral($conceptSchemeURI, "dc:modified");
+                    if ($literal) {
+                        $modifiedDate = $literal->getValue();
+                    }
+                }
+            }
+        }
+
+        return $modifiedDate;
+    }
+
+    /**
+     * Return the datetime of the latest commit, or null if git is not available or if the command failed
+     * to execute.
+     *
+     * @see https://stackoverflow.com/a/33986403
+     * @return DateTime|null
+     */
+    protected function getGitModifiedDate()
+    {
+        $commitDate = null;
+        $cache = $this->model->getConfig()->getCache();
+        $cacheKey = "git:modified_date";
+        $gitCommand = 'git log -1 --date=iso --pretty=format:%cd';
+        if ($cache->isAvailable()) {
+            $commitDate = $cache->fetch($cacheKey);
+            if (!$commitDate) {
+                $commitDate = $this->executeGitModifiedDateCommand($gitCommand);
+                if ($commitDate) {
+                    $cache->store($cacheKey, $commitDate, static::GIT_MODIFIED_CONFIG_TTL);
+                }
+            }
+        } else {
+            $commitDate = $this->executeGitModifiedDateCommand($gitCommand);
+        }
+        return $commitDate;
+    }
+
+    /**
+     * Execute the git command and return a parsed date time, or null if the command failed.
+     *
+     * @param string $gitCommand git command line that returns a formatted date time
+     * @return DateTime|null
+     */
+    protected function executeGitModifiedDateCommand($gitCommand)
+    {
+        $commitDate = null;
+        $commandOutput = @exec($gitCommand);
+        if ($commandOutput) {
+            $commitDate = new \DateTime(trim($commandOutput));
+            $commitDate->setTimezone(new \DateTimeZone('UTC'));
+        }
+        return $commitDate;
+    }
+
+    /**
+     * Return the datetime of the modified time of the config file. This value is read in the GlobalConfig
+     * for every request, so we simply access that value and if not null, we will return a datetime. Otherwise,
+     * we return a null value.
+     *
+     * @see http://php.net/manual/en/function.filemtime.php
+     * @return DateTime|null
+     */
+    protected function getConfigModifiedDate()
+    {
+        $dateTime = null;
+        $configModifiedTime = $this->model->getConfig()->getConfigModifiedTime();
+        if (!is_null($configModifiedTime)) {
+            $dateTime = (new DateTime())->setTimestamp($configModifiedTime);
+        }
+        return $dateTime;
     }
 
     /**
--- a/controller/Controller.php
+++ b/controller/Controller.php
@@ -202,7 +202,7 @@ class Controller
     protected function getModifiedDate(Modifiable $modifiable = null)
     {
         $modified = null;
-        $modifiedDate = $modifiable->getModifiedDate();
+        $modifiedDate = $modifiable !== null ? $modifiable->getModifiedDate() : null;
         $gitModifiedDate = $this->getGitModifiedDate();
         $configModifiedDate = $this->getConfigModifiedDate();
 
@@ -285,7 +285,7 @@ class Controller
      *
      * Otherwise, it returns false.
      *
-     * @param DateTime $modifiedDate the last modified date to be compared against server's modified since information
+     * @param DateTime|null $modifiedDate the last modified date to be compared against server's modified since information
      * @return bool whether it sent the HTTP 304 not modified headers or not (useful for sending the response without
      *              further actions)
      */
--- a/controller/Controller.php
+++ b/controller/Controller.php
@@ -183,42 +183,35 @@ class Controller
         echo "$code $status : $message";
     }
 
-    protected function notModified(Concept $concept = null)
+    protected function notModified(Modifiable $modifiable)
     {
         $notModified = false;
         $useModifiedDate = $this->model->getConfig()->getUseModifiedDate();
         if ($useModifiedDate) {
-            $modifiedDate = $this->getModifiedDate($concept);
+            $modifiedDate = $this->getModifiedDate($modifiable);
             $notModified = $this->sendNotModifiedHeader($modifiedDate);
         }
         return $notModified;
     }
 
     /**
-     * Return the modified date. First try to get the modified date from the concept. If found, return this
-     * date.
+     * Return the modified date.
      *
-     * Then try to get the modified date from the vocabulary's main concept scheme. If found, then return this
-     * date.
-     *
-     * Finally, if no date found, return null.
-     *
-     * @param Concept $concept
+     * @param Modifiable $modifiable
      * @return DateTime|null
      */
-    protected function getModifiedDate(Concept $concept = null)
+    protected function getModifiedDate(Modifiable $modifiable = null)
     {
-        $modifiedDate = null;
-
-        $conceptModifiedDate = $this->getConceptModifiedDate($concept);
+        $modified = null;
+        $modifiedDate = $modifiable->getModifiedDate();
         $gitModifiedDate = $this->getGitModifiedDate();
         $configModifiedDate = $this->getConfigModifiedDate();
 
         // max with an empty list raises an error and returns bool
-        if ($conceptModifiedDate || $gitModifiedDate || $configModifiedDate) {
-            $modifiedDate = max($conceptModifiedDate, $gitModifiedDate, $configModifiedDate);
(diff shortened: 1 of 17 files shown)


kinow:
Oh, the commits should more or less match what is in the text, I mean the section titles :+1:

osma:
Thanks a lot! Will review once I have a suitable time slot, probably not before next week unfortunately...

kinow:
One downside of not using the `Concept` in all methods, is that the request to a Concept page may use the Concept `dc:modified` date, while the Concept REST calls (the Concept page calls `/<vocid>` and `/hierarchy`) will be using only the git and configuration modification dates. Which not sure if it is a problem or not. Here's what happens in the second request.

image

Then after changing `skosmos:useModifiedDate` to `false`:

image

So with 304 enabled, it took `996 ms` to load the page, and without 304 it took `1.5s`.

And using `master` branch, `1.22s`, with only one request being 304'd.

image

kinow:
>Thanks a lot! Will review once I have a suitable time slot, probably not before next week unfortunately...

Perfect. I wrote a lot too (sorry). Probably best to review one day with some time to read it with some good :coffee: ! Thanks @osma !

osma:
> One downside of not using the Concept in all methods, is that the request to a Concept page may use the Concept dc:modified date, while the Concept REST calls (the Concept page calls /<vocid> and /hierarchy) will be using only the git and configuration modification dates. Which not sure if it is a problem or not. 

Hmm, this is not ideal, although I'm unsure whether it's a real problem. Both the HTML concept page and the REST data for a concept are based on the same underlying SPARQL query (the CONSTRUCT query implemented in queryConceptInfo) so both in principle could use the same timestamp.

kinow:
>Hmm, this is not ideal, although I'm unsure whether it's a real problem. Both the HTML concept page and the REST data for a concept are based on the same underlying SPARQL query (the CONSTRUCT query implemented in queryConceptInfo) so both in principle could use the same timestamp.

Yeah... I also finished it and then realized there was something wrong... the Concept page has two REST calls... one for the vocabulary information, which does not contain any timestamp. And one for the hierarchy of the Concept. The JSON-LD top level element is the concept. In theory I think that query could be modified to return the `dc:modified` value. 

But will wait for your complete feedback later to see which direction to go from here. Just wanted to prepare this draft pull request so we have something to discuss and take the next steps from here :+1:

osma:
Oh, now I realize we are talking about different things. For the REST calls performed on the concept page, it's OK to rely on the vocabulary timestamp (as well as git and config), since their results are not concept-specific.

I was thinking of the `data` REST method, which returns the same data which is used to generate the concept page, but in RDF. For that method, the timestamp should be based on the concept-level timestamp, if there is one.

osma:
...and the `data` method is the one that is shown as download links (RDF/XML, Turtle) on the concept page.

kinow:
The `data` method is exactly one of those that I did not include in this change... `data` and `search`.

When I clicked on a link and received a download pop up in the browser, I thought it would not work sending a 304 instead. But I can try. Is it the intention? To return a 304 for a call to `/data/...` that would trigger the download in the browser (of a cached version I assume?)?

osma:
I'm not sure about whether Last-Modified/304 helps for the browser in the download case, but the browser is just one possible client for the `data` method. It's the main way of serving Linked Data from Skosmos, so for example RDF-aware web crawlers could benefit from it.

kinow:
That makes sense @osma. I notice a few methods that behave different when a `uri` was present or not. In this case I think it wouldn't hurt to return 304 for the other possible formats. Will update the PR after you've had time to review the rest of the code changes and my lengthy text :grimacing:

osma:
> First modification: move methods related to modified date to Controller

Okay!

> Second modification: per vocabulary setting to enable/disable modified date [...]
> in the end I thought the easiest approach would be to simply make skosmos:modifiedDate a global configuration. What do you think?

This isn't an ideal choice. For example in Finto, which serves dozens of different vocabularies, some have good concept-level timestamps, some have only a single timestamp on the ConceptScheme, and some have no timestamps at all. In some cases the timestamps in a vocabulary can be unreliable. So I think it really should be a vocabulary-level setting. Maybe the easiest solution would be to avoid introducing timestamps and 304s to global REST methods, thus circumventing the problem?

> Third modification: method signature for getModifiedDate

Good!

> Fourth modification: realized I was always doing:

Great simplification!

> Methods that were not modified

>    data: it looks strange to return 304 for a url that may start the download of a file... what do you think?

As discussed above, this would be desirable (e.g. Linked Data crawlers could use it), but we can leave it for a later PR if you prefer.

>   search: I have never seen a search result returning 304 I think... what do you think?

Hmm...this would be difficult to implement when there are concept level timestamps. Okay to not support timestamps/304 for search operations.

> Q: I realized that I was not using the Concept in the REST methods related to concepts... but the reason is that the methods used by those REST methods are querying single string values (e.g. label), never returning a complete Concept. I could alter each query to return the dc:modified too, change the method to return... an array? Then use the dc:modified value. Is that the right direction?

Sounds like it could easily become an invasive change. I would leave this for later.

kinow:
>> Second modification: per vocabulary setting to enable/disable modified date [...] in the end I thought the easiest approach would be to simply make skosmos:modifiedDate a global configuration. What do you think?
>
>This isn't an ideal choice. For example in Finto, which serves dozens of different vocabularies, some have good concept-level timestamps, some have only a single timestamp on the ConceptScheme, and some have no timestamps at all. In some cases the timestamps in a vocabulary can be unreliable. So I think it really should be a vocabulary-level setting. Maybe the easiest solution would be to avoid introducing timestamps and 304s to global REST methods, thus circumventing the problem?

Right, so the per vocabulary setting will be added back. This would force us to have both, global and per-vocabulary, and some sort of preferred order to choose which one to use when both were available.

But I think your suggestion of avoiding to introduce timestamps and 304 to global REST methods would be simpler for now. Let me see how it looks like after this change.

controller/Controller.php
```diff
@@ -119,6 +125,126 @@ protected function returnError($code, $status, $message)
...
+        $modifiedDate = $concept->getModifiedDate();
+        if (!$modifiedDate) {
+            $vocab = $concept->getVocab();
+            $conceptSchemeURI = $vocab->getDefaultConceptScheme();
+            if ($conceptSchemeURI) {
+                $conceptSchemeGraph = $vocab->getConceptScheme($conceptSchemeURI);
+                if (!$conceptSchemeGraph->isEmpty()) {
+                    $literal = $conceptSchemeGraph->getLiteral($conceptSchemeURI, "dc:modified");
+                    if ($literal) {
+                        $modifiedDate = $literal->getValue();
```
kinow: Here we use only the Concept date, or the date of the Vocabulary's main concept scheme. We never use the Vocabulary's dc:modified value.

kinow:
Haven't had time to massage the commits, but will reply what changed first, so we can discuss more before I go around squashing (and deleting some) old changes.

>>Second modification: per vocabulary setting to enable/disable modified in the end I thought the easiest approach would be to simply make skosmos:modifiedDate a global configuration. What do you think?
>>
>This isn't an ideal choice. For example in Finto, which serves dozens of different vocabularies, some have good concept-level timestamps, some have only a single timestamp on the ConceptScheme, and some have no timestamps at all. In some cases the timestamps in a vocabulary can be unreliable. So I think it really should be a vocabulary-level setting. Maybe the easiest solution would be to avoid introducing timestamps and 304s to global REST methods, thus circumventing the problem?

Now we have again the vocabulary-level setting, just like before. And the 304s were added to vocabulary and concept methods.

>>Methods that were not modified
>>data: it looks strange to return 304 for a url that may start the download of a file... what do you think?
>>
>As discussed above, this would be desirable (e.g. Linked Data crawlers could use it), but we can leave it for a later PR if you prefer.

Easy to include this one. Just wasn't sure if required. Done, using the vocabulary `dc:modified`: 

https://github.com/NatLibFi/Skosmos/pull/869/files#diff-1e05dc421aa9a88eab0ae775652e552cR615

---

One important change was the addition of `Modifiable` interface. I initially used a `trait`, but it got a bit complicated, as `Concept` and `Vocabulary` have different ways to retrieve the modified data. And the interface has two methods, one to get the modified date, and the other to tell whether 304s are enabled.

For a `Concept`, `getModifiedDate` does the same as before, return the `dc:modified` if any, otherwise the `dc:modified` of its vocabulary main concept scheme. And 304s are enabled when the `Vocabulary` of that `Concept` has `skosmos:useModifiedDate = true`.

And for a `Vocabulary`, `getModifiedDate` returns its `dc:modified` or null. And it is enabled when this `Vocabulary` has `skosmos:useModifiedDate = true`.

I realized there is a slight different for the `Vocabulary` when using a `Concept`, or when using a `Vocabulary`. If you are using a `Concept`, the `dc:modified` of a `Vocabulary` is not used, but instead its main concept scheme's `dc:modified` may be used. But if you are using a `Vocabulary` directly, then the concept scheme is not used. I've added a comment in this PR where this happens to see what you think, and change if required :+1: 

Tests updated, and manual tests with Docker appear to have worked OK.

model/Modifiable.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+/**
+ * Interface Modifiable.
+ */
+interface Modifiable {
+
```
kinow: New interface. Adding to a parent class did not work, as well as a trait, because the two modifiables that we have (Vocabulary and Concept) each have a different way to get the modified date, and whether 304s are enabled or not.

model/Vocabulary.php
```diff
@@ -632,4 +632,20 @@ public function getId() {
       return $this->config->getId();
     }
 
+    public function getModifiedDate()
+    {
+        $modified = null;
+        // finding the modified properties
+        /** @var \EasyRdf\Resource $modifiedResource */
+        $modifiedResource = $this->resource->get('dc:modified');
+        if ($modifiedResource) {
```
kinow: For a Vocabulary, I am simply getting its `dc:modified`. Should we use the concept main scheme for a Vocabulary too @osma ? Or, should we fallback to the Vocabulary's `dc:modified` in the `Concept#getModifiedDate`, if no `dc:modified` in the Concept, and no `dc:modified` in the Concept main scheme?

osma:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- controller/Controller.php  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/NatLibFi/Skosmos/pullRequest?prid=4951455&bid=16365803)

osma:
This PR has been waiting for review way too long, sorry about that!

I merged recent work from the `master` branch and fixed a couple of merge conflicts (and reinstated a unit test that was dropped for seemingly no good reason, just had to fix it slightly). I pushed those directly to your PR branch @kinow since you've allowed it - I hope that's OK to you!

Now I'm going to take a new look to see if we could actually merge this in the current dev cycle.

osma:
Oh no, the GitHub diffs got really messed up by my latest fix-merge-conflicts-and-push operation. Now it's impossible to review. I'll try to rebase this branch instead...

osma:
Rebased on current master (fixing several conflicts along the way), then force-pushed. I hope that worked out correctly...

kinow:
> Rebased on current master (fixing several conflicts along the way), then force-pushed. I hope that worked out correctly...

Had a look at the changes, and it looks like the old changes are OK. I even remembered some of the code around that git code form StackOverflow. Nice work @osma !

osma:
I keep pushing my work to the wrong branch (NatLibFi/kinow-add-304-to-rest-controller - now deleted again - instead of the correct kinow/add-304-to-rest-controller) and now Travis seems to track the wrong branch :( Let's see if closing and reopening the PR has any effect.

osma:
Closing and reopening didn't help, but restarting the Travis build seems to have worked this time...

osma:
Phew. Now I think it's getting close to merge-ready... Still a few TODOs I can think of:

- [ ] Address Scrutinizer complaint: [`$modifiedResource->getValue()`](https://github.com/NatLibFi/Skosmos/pull/869/files#diff-b6ab98c57e17f591c74172393ba896ddR685) could break if the `dc:modified` value is a resource, not a literal
- [ ] Update Swagger documentation (add 304 return codes for the relevant methods)

What do you think @kinow?

kinow:
>Address Scrutinizer complaint: $modifiedResource->getValue() could break if the dc:modified value is a resource, not a literal

For this one see my latest commit. Had to dig into the code of EasyRDF to look at the returned types. Just added `\EasyRDF\Literal`, and used `instanceof \EasyRDF\Literal\DateTime` (child of `Date`, which is child of `Literal`)

kinow:
Fixed another Scrutinizer issue in `Vocabulary.php` also related to types in phpdocs.

kinow:
@osma 

More for the todo list:

- [ ] Fix Travis CI tests
- [ ] Manual functional test (at least for me, as I haven't tested it in a while)

Plus the swagger doc in your list, and we should be good to go. WDYT?

model/Vocabulary.php
```diff
@@ -244,7 +244,7 @@ public function getDefaultConceptScheme()
     /**
      * Returns the main Concept Scheme of that Vocabulary, or null if not set.
      * @param string $defaultConceptSchemeURI default concept scheme URI
-     * @return \EasyRdf\Graph|mixed
+     * @return \EasyRdf\Graph|\EasyRdf\Sparql\Result
```
osma, line 247: I don't think this is correct. The return type (Graph vs Result) depends on the SPARQL query type: Graph for CONSTRUCT, Result for SELECT (and ASK). The query type is hardcoded so the type is already known within Skosmos code - but of course, code inspection tools cannot know the intricacies of SPARQL queries, so some kind of explicit PHP typecheck may be necessary to keep them happy. But I don't think this kind of "it could be either" annotation makes sense. Same applies for the similar changes you did to GenericSparql phpdocs.

swagger.json
```diff
@@ -2017,6 +2072,133 @@
...
+        },
+        "notation": {
+          "type": "string",
+          "description": "Notation of the concept"
+        },
+        "type": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "description": "concepts type, e.g. \"skos:exactMatch\"."
```
osma, line 2091: I think this should be "mapping type" not "concepts type"

osma:
Great work @kinow! A couple of minor points above, and I need to take a closer look at the mappings Swagger documentation - in fact we were going to add the missing documentation in this sprint (see #862) but you did it first!

model/Vocabulary.php
kinow: Should we use `Result` here then? I will look at the other change and use the value that we expect too if that's better?

swagger.json
kinow: Changed to mapping type

model/Vocabulary.php
kinow: Dropped commit with interactive rebase, push-force'd.

kinow:
>Great work @kinow! A couple of minor points above, and I need to take a closer look at the mappings Swagger documentation - in fact we were going to add the missing documentation in this sprint (see #862) but you did it first!

Sorry, hadn't seen the issue. Hopefully either closes the issue or at least can be used as starting point. :crossed_fingers:

osma:
> Sorry, hadn't seen the issue. Hopefully either closes the issue or at least can be used as starting point. crossed_fingers

Sure, this is an excellent starting point and I think we can now say to GitHub that this PR fixes #862
(that should be enough to trigger closing the issue when this PR is merged)

model/Vocabulary.php
osma: Okay great! I'm being careful here by not saying anything about whether to fix this or not, just wanted to point out that the change was not necessarily an improvement :)

osma:
I think we can merge this now. I can't think of anything else to improve. If problems crop up, we can make new PRs to address them

osma:
Thanks again @kinow for your persistence and great work!

kinow:
Thank you for the patience and help @osma

Issue #868, opened 2019-04-23
Links to YSO Places not shown correctly on YSO concept pages


osma:
Found while testing PR #851 

## At which URL did you encounter the problem?
http://localhost/Skosmos/yso/fi/page/p25064

## What steps will reproduce the problem?
1. Use a version of YSO where the `dct:spatial` links have been changed to point to YSO Places instead of the currently used YSA geographical concepts
2. Go to the concept page for [Saint John's Church (Helsinki)](http://www.yso.fi/onto/yso/p25064)

## What is the expected output? What do you see instead?

Expected to see the "spatial coverage" shown as "Helsinki (YSO Places)"
Instead it is shown as "ysopaikat:p94137 (YSO - General Finnish ontology)":

![image](https://user-images.githubusercontent.com/1132830/56584410-5edcfa80-65e4-11e9-83fc-4f2dd769d600.png)

The problem is probably that both YSO and YSO Places share the same URI space. The concept page tries to look up the URI from within YSO (and fails) instead of YSO Places where it is really defined.

osma:
I tried to debug this a bit. There are at least two problems:

1. Concept.getProperties [calls guessVocabularyFromUri()](https://github.com/NatLibFi/Skosmos/blob/master/model/Concept.php#L600) using the current vocabulary ID (here `yso`) as the preferred vocabulary ID, leading to the conclusion that the YSO Places URI is within YSO, which it is not, and a ConceptPropertyValue is constructed instead of a ConceptMappingPropertyValue. This problem was introduced in PR #851.
2. Even if the above change from PR #851 is reverted, guessVocabularyFromURI is called in many places from within ConceptMappingPropertyValue again setting the current vocabulary ID (`yso`) as the preferred vocabulary ID. Again the invalid conclusion is that the URI exists within YSO, not YSO Places.

There should be some form of double-checking to see whether the vocabulary disambiguation was actually succesful...but I'm unsure whether it should be done from within guessVocabularyFromURI (where it could introduce additional overhead) or in the calling code (which means the fix should be done in many places).

osma:
Tried to add an additional check to disambiguateVocabulary (called from guessVocabularyFromUri) but it does introduce significant additional overhead even with some additional caching for the queryLabel method: YSO search result page response time for the query "aut" goes from 1.1 seconds to 1.9 seconds and the number of SPARQL queries from 5 to 158.

kinow:
>and the number of SPARQL queries from 5 to 158

[...]

PR #870, opened 2019-04-24
Improve guessVocabularyFromURI 

osma:
This is a fix for #868. I had to do a bit of work to fix the problem without introducing additional overhead, so the solution is divided into several parts:

1. Make disambiguateVocabulary (called from within guessVocabularyFromURI) double-check that a URI exists in a vocabulary, even though it is the "preferred" one. This is necessary in some edge cases, e.g. to distinguish between YSO and YSO Places concepts, because the vocabularies have the same URI namespace, but it does introduce an additional SPARQL query to check for a label.
2. To avoid triggering the overhead, in Concept.getProperties(), we first check whether a label for a URI already exists within the graph (the CONSTRUCT query response), and if a label is found, avoid calling guessVocabularyFromURI.
3. Similarly in Model.searchConcepts, when examining the search results, check first if their URIs are within the URI namespace of the current vocabulary (this requires a new method Vocabulary.containsURI) and if it is, skip calling guessVocabularyFromURI.

All in all this PR seems to fix the problem on the concept page while avoiding additional overhead.

While crafting this I also found a way to speed up the search result page a little bit (~10%) by using a whitelist parameter for Concept.getProperties (similar to the one in Concept.getMappingProperties introduced in PR #849), but I'll leave that for another PR.

### DIFF
--- a/model/Concept.php
+++ b/model/Concept.php
@@ -599,8 +599,8 @@ class Concept extends VocabularyDataObject
                         // create a ConceptPropertyValue first, assuming the resource exists in current vocabulary
                         $value = new ConceptPropertyValue($this->model, $this->vocab, $val, $prop, $this->clang);
                         // check whether we know the label of the resource
-                        $label = $value->getLabel();
-                        if ($label == $value->resource->shorten() || $label == $value->getUri()) {
+                        $label = $value->getLabel('', 'null');
+                        if ($label === null) {
                             // we don't know a label for the resource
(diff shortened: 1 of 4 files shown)


osma:
> While crafting this I also found a way to speed up the search result page a little bit (~10%) by using a whitelist parameter for Concept.getProperties (similar to the one in Concept.getMappingProperties introduced in PR #849), but I'll leave that for another PR.

This turned out to be problematic, so I won't do a PR now.

PR #872, opened 2019-05-13
Fix #865: By upgrading PHPUnit version to 7.5.10

### LINKED ISSUE #865: Composer warning: Package phpunit/phpunit-mock-objects is abandoned

No URL, running `composer install`

1. Run `composer install` after checking out the project
2. Search the output for

```
$ composer install
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 1 update, 0 removals
  - Downgrading twitter/typeahead.js (v0.11.1 => v0.10.5): Downloading (100%)         
Package phpunit/phpunit-mock-objects is abandoned, you should avoid using it. No replacement was suggested.
Writing lock file
Generating autoload files

```

Normal output, without warnings.

henriyli:
This fixes issue #865 by upgrading the PHPUnit version. The configuration file had to be slightly altered, but the upgrade only broke one test, which I fixed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -77,7 +77,7 @@
     "grimmlink/qtip2": "3.0.3"
   },
   "require-dev": {
-    "phpunit/phpunit": "4.8 - 6.5.5",
+    "phpunit/phpunit": "4.8 - 7.5.10",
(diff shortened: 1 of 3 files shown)


joelit:
Thanks!

Issue #734, opened 2018-03-09
Enable downloading a vocabulary in MARC


kouralex:
As @osma pointed out in https://github.com/NatLibFi/Skosmos/issues/733#issuecomment-371744149, the REST API and related Skosmos code should be aware of `.mrc` and `.mrcx` file types configured via `void:dataDump`s in `vocabulary.ttl` to be able to serve vocabularies in those downloading formats.

osma:
So the idea would be that in `config.ttl` there can be a setting like this:

```
        void:dataDump <http://dev.finto.fi/download/yso/yso.mrcx> ;
```

Changes needed in the Skosmos codebase to support this:

1. [VocabularyConfig.getDataURLs()](https://github.com/NatLibFi/Skosmos/blob/master/model/VocabularyConfig.php#L111) should be changed so that it recognizes the `.mrcx` extension and returns the correct MIME type (`application/marcxml+xml`).
2. HTML template code for the download link should be added to the `download-links` section in the template [vocab-shared.twig](https://github.com/NatLibFi/Skosmos/blob/master/view/vocab-shared.twig#L47).

jarmosaarikko:
How  would you produce the two languages in separate files, i.e. yso-fi.mrcx AND yso-sv.mrcx? Would it be as simple as adding two separate void dataDump <filename> rows?

osma:
@jarmosaarikko Good point. I think Skosmos needs to know the language of each of those `mrcx` files - and parsing it from the file name is not good. It already understands a `dct:format` property on the dump URL; we could add support for `dct:language` as well. So the configuration file would look like this:

```
:yso
[...]
    void:dataDump <http://dev.finto.fi/download/yso/yso-fi.mrcx>, <http://dev.finto.fi/download/yso/yso-sv.mrcx> .

<http://dev.finto.fi/download/yso/yso-fi.mrcx>
    dct:format "application/marcxml+xml";
    dct:language "fi" .

<http://dev.finto.fi/download/yso/yso-sv.mrcx>
    dct:format "application/marcxml+xml";
    dct:language "sv" .
```

Then the `getDataURLs` method would need to return also the language of each of those URLs (or perhaps better, take a `lang` parameter that filters the returned URLs) so that we can show the link to the right language version. Or do we want to show the links to all language versions in any case, regardless of the current UI/content language?

kouralex:
I am closing this since this was fixed in https://github.com/NatLibFi/Skosmos/pull/877.

PR #877, opened 2019-05-31
Vocabularies to be downloadable as application/marcxml+xml

joelit:
Solves #734


model/VocabularyConfig.php
```diff
@@ -108,7 +108,7 @@ public function showChangeList()
      * get the URLs from which the vocabulary data can be downloaded
      * @return array Array with MIME type as key, URL as value
      */
-    public function getDataURLs()
+    public function getDataURLs($lang = null)
```
osma: This adds a new parameter `$lang`, but I don't see where this parameter is passed to the function?

model/VocabularyConfig.php
```diff
@@ -119,17 +119,28 @@ public function getDataURLs()
                 $mimetypelit = $url->getLiteral('dc11:format');
             }
 
-            // if still not found, guess MIME type using file extension
             if ($mimetypelit !== null) {
-                $mimetype = $mimetypelit->getValue();
+                if ($lang !== null) {
+                    if ($lang == $mimetypelit->getLang()) {
```
osma: Wait, is this taking the language from the MIME type literal? That's not how `dc:format` values normally look like...

view/vocab-shared.twig
```diff
@@ -52,6 +52,9 @@
         {% if 'text/turtle' in vocab.config.dataURLs|keys %}
           <a href="rest/v1/{{ request.vocabid }}/data?format=text/turtle">TURTLE</a>
         {% endif %}
+        {% if 'application/marcxml+xml' in vocab.config.dataURLs|keys %}
+          <a href="rest/v1/{{ request.vocabid }}/data?format=application/marcxml%2Bxml">MARC21/XML</a>
```
osma: MARCXML is the standard name for the format. In principle, it could contain another flavor of MARC than MARC21 (e.g. UniMARC, FinMARC or USMARC). So just call it MARCXML, not MARC21/XML

model/VocabularyConfig.php
joelit: Thanks! Fixed the handling language parameter. (It was never used since I had the secondary preg_match handling the marc data dump according to its filename extension.) I removed the regex as unnecessary.

osma:
I think we still need a unit test (or two)?

model/VocabularyConfig.php
joelit: Ok, I'll remove the option to specify the language for marcxml downloads. Instead, I'll let all available language versions to be available.

joelit:
@osma - now this seems to be in order, although testing whether the header redirect produces the wanted download file is a bit difficult, and I left them out from this PR.

kouralex:
I think this can now safely be merged. Issues in the code have been resolved in a meaningful way.

PR #884, opened 2019-08-16
make the vocabulary-supported preflabel languages clickable

### LINKED ISSUE #871: Change content language when the user clicks on the preferred term in another language
Currently the content language of a multilingual vocabulary can be changed from a separate button next to the search box. It could be useful if a user could also change language by simply clicking the preferred term of another language.

kouralex:
fixes #871

### DIFF
--- a/view/concept-shared.twig
+++ b/view/concept-shared.twig
@@ -151,7 +151,7 @@
           {% for language,labels in concept.foreignLabels %}
             {% for value in labels %}
               <div class="row other-languages{% if prevlang != language %} first-of-language{% endif %}">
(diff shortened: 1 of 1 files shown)


danmichaelo:
Did you test this? Seems to be producing URLs on a wrong form on my test system, leading to 404s. I think

    {{ request.vocabid }}/{{ request.lang }}/{{ request.page }}/{{ request.uri }}?clang={{ value.lang }}

should perhaps be changed to

    {{ request.vocabid }}/{{ value.lang }}/{{ request.page }}/?uri={{ request.uri }}

? Will test and submit a PR

osma:
@danmichaelo Thanks for testing. It works for us in dev.finto.fi on several vocabularies. Are your URIs special in some way?

Regarding your proposed fix, the `clang` parameter is crucial but you seem to have dropped it. The idea is to change the content language and that is set by the `clang` parameter.

PR #885, opened 2019-08-16
fix 788

### LINKED ISSUE #788: Handling of long prefLabels in alphabetical listing deteriorates visual UE
As can be seen from the above image, long prefLabels in alphabetical listing cause confusion amongst users due to lack of visual differentiation between concepts.
image

Note that the same does not apply to hierarchical view as there is a visual aide in the form of the tree:
image

kouralex:
Fixes #788 by introducing a text indent to such wrapped links.

The only downside of this method is that it does not "cut" the <a> block since it is set to be a block (inline-block, to be precise), thus, creating a clickable "empty" space in the ends of wrapped lines.

### DIFF
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -959,6 +959,12 @@ div.input-group > .input-group-addon {
   list-style-type: none;
 }
 
+.alphabetical-search-results li a {
+  display: inline-block;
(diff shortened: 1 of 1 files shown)


kouralex:
Current implementation has no inadvertent side effects.

PR #890, opened 2019-09-12
Correct sorting of notated concepts in concept info page

### LINKED ISSUE #889: Values of notated concepts in concept info are incorrectly sorted
http://finto.fi/ykl/en/page/91.6
1. Observe related concepts
2. For example, 32 POLITICAL SCIENCE. POLITICS is shown below 32.5 International politics

I would prefer to see them the other way around i.e., correctly sorted.

As I promised in https://github.com/NatLibFi/Skosmos/pull/887, here is the issue (waiting for the PR).

kouralex:
Fix #889 via introducing a space between the notation and label.

Remember to squash before merging.


kouralex:
Need to address the inadvertent side effects of this before mergeing (see Travis tests).

jarmosaarikko:
The notation is a string, not a number. at leat in UDC the "decimal" period is just a visua helping mark put after every three digits and should not affect sorting at all.  Other classifications may have different meanings.  So when sorting according to notation, should take the string up to the first space and  then sort according to that string, alphabethically. Thus "49.26" should list before "49.262" and "502/504" should become before "51" and "40.02" should be before "40.1".  The last example is interesting: on the left hierarchical menu  40.02 is listed AFTER 40.1 but in the class view sublass list the 40.02 is correctly above 40.1   https://finto.fi/ykl/fi/page/40
So obviously there are two different sorting rules  being used for the same notations in the same view?

joelit:
Turned out, two independent testing environments both had insufficient configuration for this PR.  The improvement requires ```skosmos:sortByNotation true ;``` to be set. Terribly sorry this improvement was left out hanging because of it!

PR #892, opened 2019-10-03
fix #854

### LINKED ISSUE #854: Use subproperty of prefLabel for concept page headline
http://finto.fi/cn/en/page/218139A
http://finto.fi/cn/fi/page/218139A

1. Look at any CN concept page

The main headline for the page says 

    PREFERRED TERM: Sanoma Media Finland

Lower down we have

    HAS PREFERRED NAME FOR THE CORPORATE BODY: Sanoma Media Finland

This is unnecessary duplication. Also in this case the entity is a corporate body so the heading "preferred term" is a bit misleading.

What we could do here is to notice that the RDA property for "name of corporate body" is a sub-property of skos:prefLabel and 

- use the label of the subproperty instead of "preferred term" next to the headline
- hide the subproperty elsewhere to avoid repetition

kouralex:
Here you have my proposition for the issue. Feel free to comment - all comments are welcome!
I had in mind to generalize the longly-named function but after inspecting it I got the feeling that it is already done for other properties.

May require tests?

#854


model/Concept.php
```diff
@@ -662,9 +662,27 @@ public function removeDuplicatePropertyValues($ret, $duplicates)
 
             }
         }
+        // the logic is inverted for skos:prefLabel; remove the more specific property (see #854)
```
osma: The comment seems misleading; the logic is the same for skos:prefLabel as for other properties (remove the generic property, keep the specific one), but this special case code for skos:prefLabel is required because it's listed in DELETED_PROPERTIES and thus not handled the same way as other properties.

view/concept-shared.twig
```diff
@@ -39,7 +39,7 @@
       {% spaceless %}
       <div class="row{% if concept.type == 'skosext:DeprecatedConcept' %} deprecated{% endif %}">
         <div class="property-label property-label-pref">
-          <span class="versal">{{ 'skos:prefLabel'|trans|upper }}</span>
+          <span class="versal">{% if concept.preferredSubpropertyLabelTranslation(request.lang) %}{{ concept.preferredSubpropertyLabelTranslation(request.lang)|upper }}{% else %}{{ 'skos:prefLabel'|trans|upper }}{% endif %}</span>
```
osma: The template calls `preferredSubpropertyLabelTranslation` twice; this might be bad for performance. It could be better to assign a variable so that only a single call is necessary.

PR #894, opened 2019-10-10
Enhancement for passing notification messages for a general-purpose plugin

joelit:
This feature enables vocabulary specialists to create notifications for concept plages of a specific vocabulary without the need to write special plugins for each message or mess with the update schedule for the vocabularies in the main triple store. The messages are stored in vocabulary config file as RDF resources.

As it is, the feature for displaying the messages is still implemented as a separate message plugin - but should the functionality be more closely integrated as a part of Skosmos?

### DIFF
--- a/model/VocabularyConfig.php
+++ b/model/VocabularyConfig.php
@@ -6,6 +6,7 @@
 class VocabularyConfig extends BaseConfig
 {
     private $plugins;
+    private $pluginMessages;
     private $languageOrderCache = array();
 
     public function __construct($resource, $globalPlugins=array())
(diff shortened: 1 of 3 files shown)


joelit:
The separate message plugin can be found at https://github.com/NatLibFi/Skosmos-widget-message

osma:
Looks good in general, but I wonder if this could be made a bit more generic - passing any kind of configuration parameters to plugins, instead of just messages. For example, a plugin showing information from an external source (like the Finna and Wikipedia widgets do) could require configuration such as API base URL, API credentials, language, number of items to show etc.

So instead of messages, talk about parameters in general. The message plugin needs just one parameter, the message to show. Other plugins may require multiple parameters. The parameters could be given as an RDF resource with properties. These could then be encoded into a JSON(-LD?) object in the template.

osma:
There was a request for functionality like this on skosmos-users recently:
https://groups.google.com/d/msg/skosmos-users/PGKtQch79WU/ZNWmB9evCQAJ

model/VocabularyConfig.php
```diff
@@ -6,18 +6,32 @@
...
             foreach ($plugins as $pluginlit) {
                 $pluginArray[] = $pluginlit->getValue();
             }
         }
+        //Get message plugins defined as resources and their respective message literals
+        $pluginResources = $this->resource->allResources('skosmos:useMessagePlugin');
+        if ($pluginResources) {
+            foreach($pluginResources as $pluginResource) {
+                $pluginName = $pluginResource->getLiteral('skosmos:usePlugin')->getValue();
+                $pluginMessageLiterals = $pluginResource->allLiterals('skosmos:pluginMessage');
```
osma: Codacy Issue found: [Avoid excessively long variable names like $pluginMessageLiterals. Keep variable name length under 20.](https://app.codacy.com/manual/NatLibFi/Skosmos/pullRequest?prid=4951449)

osma:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/NatLibFi/Skosmos/pullRequest?prid=4951449&bid=16365797)

kouralex:
Hey, nice work @joelit ! The implementation seems reasonable - even though I believe if-ing an array is unnecessary as it is looped only if there are values in array (but! better be safe than sorry!). Also, now the values of `pluginMessages` are "preferred" - say displayed/returned first from any variables to be used in JavaScript - however, this causes no trouble and is just a very minor detail.

One side note: the naming of `skosmos:useMessagePlugin` makes it impossible to reuse this as aforementioned (thanks @osma!) general parameter mediator. Consider naming it in a more general fashion, if possible.

That said, however, this PR has my consent.

Ps. after mergeing, remember to document the new skosmos predicate.

PR #895, opened 2019-10-25
Update Dockerfile

danmichaelo:
Bumping the PHP version to 7.3

Combining some logically related RUN instructions to reduce the number of layers (https://docs.docker.com/develop/develop-images/dockerfile_best-practices/)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7.0-apache
+FROM php:7.3-apache
 
 RUN apt-get update
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,11 @@
 FROM php:7.3-apache
 
-RUN apt-get update
-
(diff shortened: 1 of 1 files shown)


kinow:
>Combining some logically related RUN instructions to reduce the number of layers (https://docs.docker.com/develop/develop-images/dockerfile_best-practices/)

Great idea! I read that best practices page some time ago, and only then learned about how/why to avoid having so many layers. +1

danmichaelo:
Thanks!

I was a bit unsure about whether to introduce the for loop by the way. It means less repetition, but also more complexity. Let me know if you see a more elegant way to do this! Unfortunate that `locale-gen` cannot just accept the locale names as arguments on all Linux variants.

osma:
Thanks @danmichaelo ! Looks good to me. I will merge this now and if any issues arise, they can be handled in subsequent PRs.

PR #897, opened 2019-11-12
Fix preflabel language links (#871)

danmichaelo:
Per my comment on https://github.com/NatLibFi/Skosmos/pull/884#issuecomment-552882869 

This follows the same URL structure as the language links in the upper right corner.

```diff
- {{ request.vocabid }}/{{ request.lang }}/{{ request.page }}/{{ request.uri }}?clang={{ value.lang }}
+ {{ request.vocabid }}/{{ value.lang }}/{{ request.page }}/?uri={{ request.uri }}
```


osma:
That's not correct - it would change the UI language, not the content language (`clang` parameter)

The problem is real but this is not the right fix. I'm just looking at this myself but I think that the link should be constructed using the `link_url` custom Twig filter function defined in WebController.php

danmichaelo:
Ah, you're right! I leave it for you then since you're already looking at it

osma:
I think this one's for @kouralex ;)

PR #899, opened 2019-11-12
Fix "Any language" label not showing in edge case

danmichaelo:
Came across an edge case where the search language selector button showed as "undefined"

Skjermbilde 2019-11-12 kl  17 14 03

It only seems to happen when 

1. the UI language (English for instance) is not a valid search language (because we don't have any English vocabularies) 
2. and when no vocabulary has been selected (that is, on the front page), since otherwise the default language for the vocabulary would kick in
3. and the `SKOSMOS_SEARCH_LANG` cookie has not been set

This PR fixes the issue for me, but please check if it makes sense. My *guess* is that `[hreflang="anything"]` was changed to `[hreflang=""]` at some point and eventually removed since it's not a valid language code.

### DIFF
--- a/resource/js/docready.js
+++ b/resource/js/docready.js
@@ -543,7 +543,7 @@ $(function() { // DOCUMENT READY
   } else if (!search_lang) {
       langPretty = $('a[hreflang=' + lang + ']').html();
       search_lang = lang;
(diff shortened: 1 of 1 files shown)


osma:
I looked around in ancient history of the codebase but couldn't find out exactly what happened. In any case, the fix looks sane!

PR #904, opened 2019-11-27
Issue903 missing foreign language tag

kouralex:
Seems like I messed up with @joelit PR that is included in this PR. However, if it gets merged, I think there should be no issues with regarding that.


kouralex:
I am closing this in order to fix unnecessary commits.

Issue #905, opened 2019-11-29
Using proxy headers to determine base url


danmichaelo:
I know you can set `skosmos:baseUrl` manually in config.ttl, but I want to avoid that since I'm trying to use the same config file in testing and production.

It would be nice if Skosmos could determine base url based on proxy headers (`X-Forwarded-Proto` and `Host`) in the same way as [Symfony does](https://symfony.com/doc/current/deployment/proxies.html).

---

Rather than duplicating that functionality, I tried puling in [symfony/http-foundation](https://github.com/symfony/http-foundation) and using it: https://github.com/danmichaelo/skosmos/commit/5bfa6cf6c2e2c11541a751fdcb62a8105377ee99

But if we add this new dependency, it immediately seems like it would make sense to also integrate the `Request` class of Symfony with the `Request` class of Skosmos. Either by having the  Skosmos class extend the Symfony one, or by wrapping it. Then it would be possible to gradually utilize more and more of the functionality provided by Symfony.

But then we need a way to get the Request object into the `guessBaseHref()` method in `Controller`. Which turned out to be harder than i first imagined! Some options:

- Add it as an argument: `guessBaseHref(Request $request)`. **Positive**: Makes it explicit that the method depends on the request. Does not add state. **Negative**: `getBaseHref()` is called from many places, including a few methods in EntityController that do not have `$request` in their scope now, so we would have to add that too. And `getBaseHref()` is even called from the `WebController` constructor... which leads to the next option:
- Add it to the `Controller` constructor: `__construct(Model $model, Request $request)`. **Positive**: Easy to implement. **Negative**: More state. And need to rewrite all the controller methods to use `$this->request` rather than take `$request` as an argument. **If** that's even possible. It seems like the object is mutated in `index.php` before being passed to the controller methods, so it might not work to just inject it into the constructor. Hm. Since it's passed by reference, it might still work, but then it becomes increasingly hard to trace what happens :/
- A third option could be to follow the first option, but avoid having to pass the request into the `WebController` constructor by extracting the Twig initialization stuff into a new method `initTwig(Response $response)` or something like that, which is called from alle the class methods. **Positive**: Perhaps the cleaner approach? **Negative**: Lots of refactoring?

---

[...]

PR #911, opened 2019-12-09
Automatic base url protocol detection behind reverse proxy (#905)

danmichaelo:
<s>Use proxy headers (X-Forwarded-Proto and X-Forwarded-Host) to determine
the base url if configured to trust these headers (new configuration
option).

If this is merged, we also need to update https://github.com/NatLibFi/Skosmos/wiki/Configuration</s>

I scaled this back to just use X-Forwarded-Proto to avoid any security implications, as discussed in #905. It still solves my use case of correctly guessing the base url behind a TLS termination proxy.

### DIFF
--- a/controller/Controller.php
+++ b/controller/Controller.php
@@ -82,6 +82,15 @@ class Controller
         return $format;
     }
 
+    private function isSecure()
+    {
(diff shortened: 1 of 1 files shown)


osma:
Thanks, and sorry for the delay! Simple and sweet!

PR #919, opened 2020-02-03
Upgrade to PHP 7.1-7.4 and Fuseki 3.14.0

### LINKED ISSUE #917: Bump PHP version
Currently Skosmos requires 7.0 and supports versions 7.1 and 7.2. But PHP 7.3 and 7.4 have been available for some time. I suggest the following:

* drop support for PHP 7.0 (this would enable using more advanced parameter type hints introduced in 7.1 - but actually doing that is out of scope for this issue)
* try to run Skosmos under PHP 7.3 and 7.4
* enable running Travis CI tests under 7.3 and 7.4 (with allowed_failures set to true if they cannot be run succesfully already)
* open separate issues for any problems found with 7.3 and 7.4

osma:
This PR changes the Travis CI environment where tests are run.

Support for PHP 7.0 is dropped (we no longer run the tests using it)
Support for PHP 7.1 and 7.2 remains
Support for PHP 7.3 is introduced (and tests confirm it)

Tests are also run with PHP 7.4, but currently the [builds fail](https://travis-ci.org/NatLibFi/Skosmos/jobs/645471054) - they are marked as "allowed failures" in Travis configuration. I will open a separate issue on fixing the PHP 7.4 related problems.

The Fuseki version used for running integration tests is upgraded from 3.9.0 to 3.14.0. There was a problem in this version that scripts were not executable in the tarball distribution, so this PR adds a chmod command to work around it (issue [reported and fixed](https://issues.apache.org/jira/browse/JENA-1834) in subsequent Jena Fuseki versions)

Fixes #917

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ env:
   global:
     - CC_TEST_REPORTER_ID=fb98170a5c7ea9cc2bbab19ff26268335e6a11a4f8267ca935e5e8ff4624886c
   matrix:
-    - FUSEKI_VERSION=3.9.0
+    - FUSEKI_VERSION=3.14.0
     - FUSEKI_VERSION=SNAPSHOT
 matrix:
   exclude:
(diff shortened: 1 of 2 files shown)


kouralex:
Looking good, mergeing. Thanks @osma !

PR #922, opened 2020-02-07
Copy to clipboard button refers to notation element in the title bar if present

### LINKED ISSUE #832: Copy-to-Clipboard for skos:notation when it is displayed in Classification class view
https://finto.fi/udcs/en/page/015172
1. View any class view
2. PREFERRED TERM shows the skos:notation and skos:prefLabel
3. The latter has a copy to Clipboard function which copies the prefLabel, not the notation.

For classification use the notation is generally stored in content description systems as the representation of the class. the prefLabel is only human readable "Caption" for the class.
Expected "111"
seen: "General metaphysics. Ontology"
Chrome.

&lt;span class="notation"&gt;111&lt;/span&gt;

joelit:
Fixes #832


joelit:
Good idea - I'll add the improvements for the copy to clipboard button as a new issue. It seems my repository configuration broke somehow, which caused commits being added to this feature branch. (Somehow, it shoudn't be possible!) I made a new feature branch for this issue, just in case.

PR #923, opened 2020-02-07
Use phpdbg instead of Xdebug for Travis coverage

osma:
Travis CI environment by default uses Xdebug to create coverage reports. But this slows down unit tests significantly. This PR disabled Xdebug and instead uses phpdbg, which reduces the time spent on Travis CI jobs by around half, from 12 minutes to 6 minutes (per job, but those are usually run in parallel).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_script:
   - sh ./init_fuseki.sh
   - cd ..
 script:
-  - phpdbg -qrr phpunit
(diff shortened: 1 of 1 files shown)


osma:
The reported coverage seems to be lower for phpdbg than it was with Xdebug: for example lines containing `} else {` clauses are not counted. 

Still, I think the improvement in speed is worth it; we could look at switching to PCOV later it the phpdbg reports turn out to be too inaccurate.

PR #930, opened 2020-02-10
Fix null-handling of skosmos:sparqlGraph

danmichaelo:
The `skosmos:sparqlGraph` vocabulary property is listed as optional, but when not set, Skosmos will produce a notice:

    Notice:  Trying to access array offset on value of type null in /var/www/html/model/sparql/GenericSparql.php on line 155

### The first commit:

- Adds a null check to the `GenericSparql::isDefaultEndpoint()` method.
- Moves the null check from `GenericSparql::generateFromClause()` to `GenericSparql::getVocabGraphs()` and updates the former method to use the latter to reduce code duplication.
- Updates docstrings to use `string|null`

### The second commit:

- Updates `GenericSparql::formatFilterGraph` to not add a FILTER clause if none of the vocabularies have a `skosmos:sparqlGraph`. This adreses the issues where you would always get zero results if you didn't specify `skosmos:sparqlGraph` for any vocabulary, since that would result in this: `FILTER (?graph IN ())`.

Note: Apart from the case `skosmos:sparqlGraph` is not defined for any vocabulary, there is also the case where `skosmos:sparqlGraph` is specified for *some* of the vocabularies, but not for others. In that case, the default search will only return results for the vocabularies that do specify `skosmos:sparqlGraph`, which is potentially confusing I guess. Perhaps it should be noted in the docs that it's a good idea to either include `skosmos:sparqlGraph` for all vocabularies or for none?

### DIFF
--- a/model/Model.php
+++ b/model/Model.php
@@ -568,7 +568,7 @@ class Model
      * Returns a SPARQL endpoint object.
      * @param string $dialect eg. 'JenaText'.
      * @param string $endpoint url address of endpoint
-     * @param string $graph uri for the target graph.
+     * @param string|null $graph uri for the target graph.
      */
     public function getSparqlImplementation($dialect, $endpoint, $graph)
     {
--- a/model/Vocabulary.php
+++ b/model/Vocabulary.php
@@ -37,7 +37,7 @@ class Vocabulary extends DataObject
     /**
      * Get the SPARQL graph URI for this vocabulary
      *
-     * @return string graph URI
+     * @return string|null graph URI
      */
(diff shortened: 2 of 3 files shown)


osma:
Thanks, I will review it soon!

osma:
Hmm the Codecov report looks surprising. I'm pretty sure this PR won't increase coverage by over 7 % :)

osma:
Thanks, looks good and I merged it.

I fixed a couple of minor issues (based on Scrutinizer complaints) in subsequent commits.

PR #944, opened 2020-03-03
Fix problems displaying external links 

### LINKED ISSUE #935: Invalid links to external URIs in non-mapping properties
https://finto.fi/yse/en/page/Y382431

1. Go to above URL
2. Click on the GitHub URL for the home page field

Expected to get to the GitHub URL shown as the link text. Instead getting a page for an empty concept with the URL https://finto.fi/yse/en/page/?clang=fi&uri=https%3A%2F%2Fgithub.com%2FFinto-ehdotus%2FYSE%2Fissues%2F3173

This probably broke in the refactoring that was part of PR #921, in particular the use of ConceptPropertyValue instead of ConceptPropertyMappingValue for non-mapping properties (such as `foaf:homepage` in the above example)

osma:
by more judicious use of isExternal and getExVocab. Fixes #935

ConceptMappingPropertyValue.isExternal() is removed because it is no longer being used.

### DIFF
--- a/model/Concept.php
+++ b/model/Concept.php
@@ -606,8 +606,11 @@ class Concept extends VocabularyDataObject
                             // checking if the property value is not in the current vocabulary
                             $exvoc = $this->model->guessVocabularyFromURI($val->getUri(), $this->vocab->getId());
                             if ($exvoc && $exvoc->getId() !== $this->vocab->getId()) {
-                                $value = new ConceptPropertyValue($this->model, $exvoc, $val, $prop, $this->clang, true);
+                                $voc = $exvoc;
+                            } else {
+                                $voc = $this->vocab;
                             }
+                            $value = new ConceptPropertyValue($this->model, $voc, $val, $prop, $this->clang, true);
                         }
                         $ret[$prop]->addValue($value, $this->clang);
                     }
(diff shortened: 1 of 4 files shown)


model/Concept.php
```diff
@@ -606,8 +606,11 @@ public function getProperties()
                             // checking if the property value is not in the current vocabulary
                             $exvoc = $this->model->guessVocabularyFromURI($val->getUri(), $this->vocab->getId());
                             if ($exvoc && $exvoc->getId() !== $this->vocab->getId()) {
-                                $value = new ConceptPropertyValue($this->model, $exvoc, $val, $prop, $this->clang, true);
+                                $voc = $exvoc;
+                            } else {
+                                $voc = $this->vocab;
                             }
+                            $value = new ConceptPropertyValue($this->model, $voc, $val, $prop, $this->clang, true);
```
kouralex: Now this object is created twice in case of `else` clause. It might be nicer and more efficient to just set the `external` value of ConceptPropertyValue.

osma: The `external` value is not the only thing that changes. Also the `vocab` parameter (second parameter) is different in the second constructor call.

But I agree this is not very clean, I will try to implement another approach.

osma:
> in twig template that still uses the old code that must be taken care of

This is as it should be. ConceptPropertyValue still has an isExternal method. It was only removed from ConceptMappingPropertyValue where it was not used anymore.

osma:
Scrutinizer reports bogus code duplication issues - not really a problem with the code.

SonarCloud reports code smells because of repeated literal strings in tests - not a problem and we have many of these already in the test suite.

model/Concept.php
osma: I've now moved this logic inside the ConceptPropertyValue constructor.

PR #945, opened 2020-03-04
Issue896 notated alphabetical listing

### LINKED ISSUE #896: Showing notation in the alphabetical listing for classifications
http://dev.finto.fi/ykl/fi/index
or any concept view.

1. The alphabetical listing shows 10 equal captions. This is a problem with numerical classifications. the same caption may be used several times in different branches. This produces a problem for browsing the alphabetical listing which is not easy to solvel.
2.Aasia
Aasia
Aasia
Aasia
etc.
3.

One possible solution: on the **alphabetical view** always show the **skos:notation**  in parenthesis AFTER the caption (skos:prefLabel).  
---  in principal, almost the same way as the skos:notation (class="tree-notation") is shown BEFORE the caption in the **hierachical view**.
Aasia (26.1)
Aasia (48.1)
Aasia (70.981)
Aasia (72.981)
Aasia (73.981)
etc.

Chrome

Related issue: https://github.com/NatLibFi/Skosmos/issues/264

kouralex:
Fixes #896 via introducing a `qualifier` attribute to the search result, if one exists. In case of multiple, they are returned in separate rows.

For literals, only plain literal value is shown. For IRIs, local name is shown.

### DIFF
--- a/tests/GenericSparqlTest.php
+++ b/tests/GenericSparqlTest.php
@@ -181,6 +181,133 @@ class GenericSparqlTest extends PHPUnit\Framework\TestCase
   /**
    * @covers GenericSparql::queryConceptsAlphabetical
    * @covers GenericSparql::generateAlphabeticalListQuery
+   * @covers GenericSparql::transformAlphabeticalListResults
+   */
+  public function testQualifiedNotationAlphabeticalList() {
+    $voc = $this->model->getVocabulary('test-qualified-notation');
+    $res = new EasyRdf\Resource("http://www.w3.org/2004/02/skos/core#notation");
+    $sparql = new GenericSparql('http://localhost:13030/skosmos-test/sparql', $voc->getGraph(), $this->model);
+
+    $actual = $sparql->queryConceptsAlphabetical("a", "en", $limit = null, $offset = null, $classes=null, $showDeprecated=false, $qualifier=$res);
+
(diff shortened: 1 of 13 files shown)


tests/GenericSparqlTest.php
```diff
@@ -178,6 +178,133 @@ public function testQueryConceptsAlphabeticalLimitAndOffset() {
...
+   * @covers GenericSparql::queryConceptsAlphabetical
+   * @covers GenericSparql::generateAlphabeticalListQuery
+   * @covers GenericSparql::transformAlphabeticalListResults
+   */
+  public function testQualifiedNotationAlphabeticalList() {
+    $voc = $this->model->getVocabulary('test-qualified-notation');
+    $res = new EasyRdf\Resource("http://www.w3.org/2004/02/skos/core#notation");
+    $sparql = new GenericSparql('http://localhost:13030/skosmos-test/sparql', $voc->getGraph(), $this->model);
+
+    $actual = $sparql->queryConceptsAlphabetical("a", "en", $limit = null, $offset = null, $classes=null, $showDeprecated=false, $qualifier=$res);
```
osma: Here (and in a few similar places) you use "named parameters" (like in Python) but PHP doesn't support it, so in effect you are declaring a bunch of unused local variables. SonarCloud complains about them.

PR #938, opened 2020-02-24
fixes #934; also adds some tests for upcoming #854

kouralex:

Issue #934, opened 2020-02-11
Running PHPUnit tests causes extra output


kouralex:
At the moment running the tests defined in `tests` directory causes some extra output in the form of risky tests as well as some printing. I'd prefer to have none, however, this may require updating the PHPUnit in use, in case there are new PHPUnit features introduced to Skosmos code base.

PR #947, opened 2020-03-04
additional fixes for #934

kouralex:
Previously, there were two risky tests as mentioned by #946. These were fixed in #938, however, they got intermingled with issue #854. This PR fixes the remaining risky tests by introducing a new, better sorting for ConceptProperties, and calls appropriate methods for the aforementioned tests.

For testing, please remember to set the `skosmos:sortByNotation` predicate to `true`.

### DIFF
--- a/model/ConceptProperty.php
+++ b/model/ConceptProperty.php
@@ -86,10 +86,35 @@ class ConceptProperty
 
     private function sortValues()
     {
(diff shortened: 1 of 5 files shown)


joelit:
I ran the tests and resolved a conflict in the test config file. Looking good.

PR #950, opened 2020-03-06
Issue617 search by notation

### LINKED ISSUE #617: Support truncating notation in search
The search help states that that you can truncate with "*" and that it right-truncates by default. This currently doesn't work for notations though, which are not part of the search index (#456). At least for decimal classification systems it would be quite handy – e.g. if I search for a Dewey number "539.7" to see what classes there are having a number starting with that number.

joelit:
Fixes #617

### DIFF
--- a/model/VocabularyConfig.php
+++ b/model/VocabularyConfig.php
@@ -278,6 +278,15 @@ class VocabularyConfig extends BaseConfig
     {
         return $this->getBoolean('skosmos:explicitLanguageTags');
     }
+    
+    /**
+     * Returns a boolean value set in the config.ttl config.
+     * @return boolean
+     */
+    public function searchByNotation()
+    {
+        return $this->getBoolean('skosmos:searchByNotation');
+    }
 
     /**
      * Returns skosmos:marcSourcecode value set in config.ttl.
--- a/model/VocabularyConfig.php
+++ b/model/VocabularyConfig.php
@@ -278,7 +278,7 @@ class VocabularyConfig extends BaseConfig
     {
         return $this->getBoolean('skosmos:explicitLanguageTags');
     }
-    
+
     /**
      * Returns a boolean value set in the config.ttl config.
      * @return boolean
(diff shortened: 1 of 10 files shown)


joelit:
To test this feature, you need to add `skos:searchByNotation "true";` for a vocab with notation codes. You also need set up lucene index for skos:notations (in /etc/fuseki/configuration/ or similar) in the same manner as in tests/fuseki-assembler.ttl, restart fuseki and load the data in the text index before trying it out: 

`
<#entMap> a text:EntityMap ;
    text:entityField      "uri" ;
    text:graphField       "graph" ; ## enable graph-specific indexing
    text:defaultField     "pref" ; ## Must be defined in the text:map
    text:uidField         "uid" ;
    text:langField        "lang" ;
    text:map (
         # skos:prefLabel
         [ text:field "pref" ;
           text:predicate skos:prefLabel ;
           text:analyzer [ a text:LowerCaseKeywordAnalyzer ] ]
         # skos:altLabel
         [ text:field "alt" ;
           text:predicate skos:altLabel ;
           text:analyzer [ a text:LowerCaseKeywordAnalyzer ] ]
         # skos:hiddenLabel
￼         [ text:field "hidden" ;
￼           text:predicate skos:hiddenLabel ;
￼           text:analyzer [ a text:LowerCaseKeywordAnalyzer ] ]
￼           # skos:notation
￼         [ text:field "notation" ;
￼           text:predicate skos:notation ;
￼           text:analyzer [ a text:LowerCaseKeywordAnalyzer ] ]
￼         ) .`

joelit:
> Some outstanding issues that we just covered face to face:
> 
> * in the SPARQL query, the IF clauses for priorities are slightly off, the numbers are overlapping
> * the `?blabel` variable is bound but never used, it could be dropped
> * last but not least, it's probably not a good idea to use an UNION of two kinds of jena-text queries; instead the language parameter should be passed another way

I made the changes and tested the result. However, there should be some further work with how search results are sorted, as I didn't really go through the situation of displaying notation matches before hidden labels matches.

PR #956, opened 2020-03-12
Adjust style of the "counts by type" table. 

### LINKED ISSUE #955: "counts" table in vocabulary page is missing class="table"

http://erato.sparna.fr/skosmos-ar/thesaurus/en/

  - Go to a vocabulary page.
  - In the HTML code, look for the `<table id="counts"` 
  - it is missing the class="table"

This has a side-effect when displaying in Arabic, RTL.

The table should have class="table", but also have a max-width=450px just like the table id="statistics" below it.

osma:
Adds proper CSS styles of the "counts" table on the vocabulary page.
Fixes #955

### DIFF
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -1288,13 +1288,15 @@ tr.replaced-by > td > ul > li > a {
   padding-left: 0;
 }
 
+#counts { max-width: 300px; }
+
 #statistics { max-width: 450px; }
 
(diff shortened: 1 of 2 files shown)


osma:
@tfrancart Does this solve your Arabic RTL issue (#955)?

tfrancart:
Yes it works, thanks !

osma:
Thanks for the quick confirmation @tfrancart !

PR #958, opened 2020-03-19
Fix jena-text searches and alphabetical index restrictions by language

### LINKED ISSUE #957: Labels are displayed in every language
1. Go to the front page of some vocabulary, e.g. YSO
2. The alphabetical list on the right shows the same concepts several times, these would be the prefLabel in fi, sv and en:
image
3. The drop-down search displays the label for every matching language:
image

Each concept should be displayed once with its suitable label.

It seems that the language parameter is passed incorrectly for the alphabetical listing and the drop-down search results, which means the results are matched in every language.

osma:
Fixes #957

Comes with unit tests that verify the desired result.

### DIFF
--- a/model/sparql/JenaTextSparql.php
+++ b/model/sparql/JenaTextSparql.php
@@ -30,11 +30,11 @@ class JenaTextSparql extends GenericSparql
      *
      * @param string $term search term
      * @param string $property property to search (e.g. 'skos:prefLabel'), or '' for default
-     * @param string @$searchLang for language code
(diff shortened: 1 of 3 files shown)


osma:
SonarCloud complains about duplicated code in tests (GenericSparqlTest and JenaTextSparqlTest have many duplicate tests, which only differ in their PHPUnit annotations). This is by design, although not ideal.

joelit:
Looks fine overall. I had some problems running the tests locally, but it doesn't seem to be related to this as everything seems to work in Travis.

PR #962, opened 2020-03-25
Add Docker files to run self-contained image for trying out Skosmos

### LINKED ISSUE #910: Self-contained Docker image for trying out Skosmos
The current [Docker support](https://github.com/NatLibFi/Skosmos/wiki/Install-Skosmos-with-Fuseki-in-Docker) is based on the idea that the Docker image provides an environment where Skosmos can run, but Skosmos itself isn't included within the image. Dependencies must be installed outside the container and the result bind-mounted to the container.

This isn't ideal for a new user that just wants to try out Skosmos. I think it should be possible to make a more self-contained Docker image that includes the Skosmos codebase, all dependencies etc. It wouldn't have to be fully production-ready, because local requirements are likely to be slightly different each time, but at least the installation should be enough to get started. Fuseki would still be outside the image, I think.

What do others think? @kinow?

kinow:
Closes #910 

Initial Docker files to try out Skosmos 2.4. Uses the Ubuntu base image, and installs Apache2, PHP, composer, git. It is similar to the InstallTutorial, but there are still some steps missing. Some of the steps require trying to customize/parameterize the Jena image.

I am creating this PR now so that I can check what else needs to be done.

The `Dockerfile.ubuntu` is for the Skosmos 2.4 image, that at the moment is using YSO and YSA vocabularies from Finto dev server. I haven't done the extra step of the tutorial, to use different vocabularies as that would require Fuseki. I thought some users would prefer to be able to run just one container to try Skosmos first?

There is also a `docker-compose.yml` that uses a different config.ttl, which points to the Jena Fuseki container. It is not loading any data at the moment, but I was starting to work on having 1 vocabulary only, then leave the commands to load the data to the user. Not sure what's the best way here, maybe the `docker-compose.yml` should try to do more, and also load the data (not sure if I can do that), and perhaps complete the rest of the tutorial?

Cheers
Bruno


osma:
Thanks @kinow, testing this now!

kinow:
Thanks @osma! There's further work that can be done on this. Just let me know which direction you think we should go, and anything that is missing or that needs fixing in the image :-)

dockerfiles/Dockerfile.ubuntu
```diff
@@ -0,0 +1,75 @@
...
+    tar -zxvf skosmos.tar.gz -C /var/www/html/ --strip-components=1 && \
+	rm -rf /Skosmos* /skosmos.tar.gz
+
+WORKDIR /var/www/html
+
+# Install Skosmos dependencies
+RUN /usr/local/bin/composer install --no-dev --no-interaction
+
+# Configure Skosmos
+COPY config/config.ttl /var/www/html/config.ttl
```
osma, line 73: When running `docker build` as instructed in README.md, this line fails:

```
Step 18/19 : COPY config/config.ttl /var/www/html/config.ttl
COPY failed: stat /var/lib/docker/tmp/docker-builder526231501/config/config.ttl: no such file or directory
```

Did you forget to include `config/config.ttl` in your commit?

kinow: That's funny. I think it should be in this PR. Did you `cd dockerfiles` before running the command?

kinow: Ignore that! I think it was filtered out by gitignore :sweat_smile: sorry!

kinow: Fixed, sorry @osma !

osma:
Thanks for the quick fix @kinow !

osma:
Got both the single image and the docker-compose variants running. I will do more detailed testing tomorrow and try to come up with answers to your questions!

dockerfiles/Dockerfile.ubuntu
```diff
@@ -0,0 +1,75 @@
+FROM ubuntu:18.04
+
+ARG DEBIAN_FRONTEND=noninteractive
+ARG SKOSMOS_TARGZ_RELEASE_URL=https://github.com/NatLibFi/Skosmos/archive/v2.4.tar.gz
+
+# git is necessary for some composer packages e.g. davidstutz/bootstrap-multiselect
+RUN apt-get update && apt-get install -y \
```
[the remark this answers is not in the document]
kinow: Oh, good catch!

dockerfiles/Dockerfile.ubuntu
```diff
@@ -0,0 +1,75 @@
+FROM ubuntu:18.04
+
+ARG DEBIAN_FRONTEND=noninteractive
+ARG SKOSMOS_TARGZ_RELEASE_URL=https://github.com/NatLibFi/Skosmos/archive/v2.4.tar.gz
```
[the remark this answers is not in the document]
kinow: This makes perfect sense! Always easier to say that afterwards, but not sure why I didn't think of that. First version actually had 2.3, then I updated it after 2.4 was released. I should have thought in simply using the current commit.

@osma only downside of this approach is that if the current branch/commit is broken, then so would the container be. Should we get the latest tag, or latest commit? (I'm happy with either, as later we should be able to upload 2.3, 2.4, etc, using this image).

kinow:
Hi @osma ! Thanks for the great feedback. And lots of good ideas, so +1 from me. Next updates here should have:

- Skosmos
  * [ ] Try to combine the existing `Dockerfile` and `docker-compose` with these, to reduce/simplify the Docker files in this repository
    - perhaps we could have a shell entrypoint that by default uses the current commit of this repository (not sure if doable, but worth investigating), or the user can provide a tag/commit. maybe users should also be able to ask it to enable xdebug or extra modules... :thinking: 
  * [ ] Set timezone to UTC
  * [ ] Add the APC module to PHP (`php-apcu`)
  * [ ] Instead of using a fixed version like 2.4, use the current code to build the project
  * [ ] Putting the example vocabularies (STW and UNESCO) in place (i.e. in config.ttl)
- Jena/Fuseki
  * [ ] Use a port different than 3030 for the Fuseki image
  * [ ] Set up jena-text
  * [ ] Putting the example vocabularies (STW and UNESCO) in place (i.e. load data into fuseki/jena-text)
- Bonus
  * [ ] reverse HTTP cache - Varnish or nginx, probably - to speed up the Fuseki queries, as explained in the tutorial

@osma these are the requirements I could think of from the reviews so far. What do you think?

dockerfiles/Dockerfile.ubuntu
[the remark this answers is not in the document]
osma: > only downside of this approach is that if the current branch/commit is broken, then so would the container be

I don't see that as a problem, really. Once the Dockerfile is in place, we will probably try to set up an autobuild pipeline that creates Docker images based on git tags and the current `master` branch - just like we already do for Annif (there is a Drone pipeline that uploads images to quay.io - see here: https://quay.io/repository/natlibfi/annif?tab=tags ). So casual users could find images corresponding to the latest releases there and they could simply run `docker pull` to get whichever version they like, without even having to clone or download the git repo. Developers can build images for themselves using the Dockerfile in this repo, but they are then responsible for the result :)

osma:
Sounds like a great plan @kinow!

> Try to combine the existing Dockerfile and docker-compose with the files in this PR, to reduce/simplify the Docker files in this repository 

Just to clarify: I think it would be great to have:

1. A `Dockerfile`, which by default creates a stand-alone image that relies on the dev.finto.fi SPARQL endpoint. The Skosmos code is copied from the surrounding repo.
2. A `docker-compose.yml` file which uses the image created by the above Dockerfile, but also includes a Fuseki container (with jena-text index and the example vocabularies STW & UNESCO) and changes the configuration of the Skosmos container to use this local Fuseki instead of dev.finto.fi.
3. A variant of 1. (perhaps just explained in the documentation) where the Skosmos code and `config.ttl` is mounted from the surrounding repo, so that it remains editable. This would be appropriate for development purposes, so that one can quickly test code changes without having to rebuild the whole container. It would replace the existing top-level Dockerfile.

PS. I just had to test the new "convert to draft PR" button on this PR. Draft PRs are a great feature, but originally you had to mark PRs as drafts when you created them, which I often forgot, and it wasn't possible to change that later.

kinow:
Agreed on the Docker files. Should we move everything under `dockerfiles`?

>PS. I just had to test the new "convert to draft PR" button on this PR. Draft PRs are a great feature, but originally you had to mark PRs as drafts when you created them, which I often forgot, and it wasn't possible to change that later.

Today I learned. Didn't see any announcement of that feature. Thanks!!

osma:
> Should we move everything under dockerfiles?

Yes, I think it's a good idea to reduce the clutter at the root level.

> Today I learned. Didn't see any announcement of that feature. Thanks!!

Me neither, I just noticed the link in the right hand column.

osma:
Any news @kinow? We're just starting a new sprint next week (though focused on API issues).

kinow:
Not much progress, I have some local changes in the other repository I had for docker+skosmos from when I was testing building an image with parameters. Refreshing my memory and will update it soon. I just finished working on two other images (SLURM and another interesting one to run multiple versions of `bash`). Gotta review the current image too to see if there are any improvements I could do :-) great to hear about the new sprint!

kinow:
@osma 

>Putting the example vocabularies (STW and UNESCO) in place (i.e. load data into fuseki/jena-text)

Did a bit more of progress today, but had some issues with this item. I first tried using the `load.sh` or `tdbloader` as instructions from `jena-fuseki` container. However, these require only one access per time to the fuseki DB. So it's harder to automate with `docker-compose`.

Tried adding an extra container responsible only for loading the data. But the issue now was that I had to either include the two .ttl (~7MB total), or gzipped (<1MB). But then had to automate decompressing the files, loading into the volume, and stopping the container to free the lock to Fuseki in another container.

It started to look a bit too convoluted this set up.

I've downloaded both ttl from their vocabulary pages. Then loaded with

```bash
curl -I -X POST -H Content-Type:text/turtle -T unescothes.ttl -G http://localhost:9030/skosmos/data --data-urlencode graph=http://skos.um.es/unescothes/
curl -I -X POST -H Content-Type:text/turtle -T stw.ttl -G http://localhost:9030/skosmos/data --data-urlencode graph=http://zbw.eu/stw/
```

Would it be all right if users had to start `docker-compose up` followed by some manual steps to load the data into Fuseki/Jena? This way we wouldn't need to include the .ttl source data, and the setup would be a bit simpler.

osma:
> Would it be all right if users had to start docker-compose up followed by some manual steps to load the data into Fuseki/Jena? This way we wouldn't need to include the .ttl source data, and the setup would be a bit simpler.

I think that makes sense, if the alternative is so complicated.

kinow:
Got it working with most of the new requirements. Will look at varnish this weekend, give it some more testing, check what docs need to be updated, then it should be ready for review again :+1:

osma:
Great! Thanks for the update @kinow, I was just wondering about this PR

kinow:
All done @osma !

Did a bit of testing today, and it seems to be working fine.

The docker compose uses Jena Fuseki image. Besides the limitation to load the data automatically, we have one new problem: https://github.com/stain/jena-docker/issues/35

If you restart the setup you may have this issue, and then you have to try to kick the process or re-create the container (the latter is the easiest, simply `docker-compose rm -rf` then `docker-compose up` again).

I will take a look to see if it's easy to fix in the jena-fuseki image.

Here's how to test (will add to the Running Fuseki with Docker wiki page later):

- `docker-compose up` (if you already have the images, you may want `docker-compose rm -rf` first)
- Download STW and Unesco vocabulary data
- Visit `http://localhost:9030/` and upload the turtle files using the right graphs
- Visit `http://localhost:9090`. The request goes to `skosmos-web` PHP+Apache container
- The `skosmos-web` container will mount a volume for `/var/www/html` using the location of the `Skosmos` directory, i.e. the same version you are running. **It means you need to run composer install` before**. That's because otherwise we would need that in the container, running with user `root`, that would change the permission of the composer data (one alternative is investigate running with the same user and see if that works, though a bit harder)
- The requests from `skosmos-web` go to `fuseki-cache:80` (default port of the Varnish Docker image), that redirects to `fuseki:3030` (3030 in the docker network, for the host it's mapped as 9030)

I think now we are ready for another review and decide whether it can be merged, or if it needs further requirements/changes :+1: 

Thanks
Bruno

kinow:
Note: Jena is including a Fuseki Docker image: https://github.com/apache/jena/pull/786

osma:
Hi @kinow , I'm very sorry this PR has been without a response for so long. I'm taking a closer look at it this week to see how it's working and if it can be merged.

Regarding the Fuseki container for Jena - do you think it would make sense to switch to that instead of stian's version?

dockerfiles/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+Now Skosmos should be available at `http://localhost:9090/`.
+
+To stop:
+
+```bash
+docker-compose down
+```
+
+## License
```
osma, line 63: Why is this necessary here? I mean, Skosmos is under MIT License yes, but what does it mean to state it here in this README?

dockerfiles/config/config-docker-compose.ttl
```diff
@@ -0,0 +1,94 @@
...
+    skosmos:httpTimeout 5 ;
+    # customize the service name
+    skosmos:serviceName "Skosmos" ;
+    # customize the base element. Set this if the automatic base url detection doesn't work. For example setups behind a proxy.
+    # skosmos:baseHref "http://localhost/Skosmos/" ;
+    # interface languages available, and the corresponding system locales
+    skosmos:languages (
+        [ rdfs:label "fi" ; rdf:value "fi_FI.utf8" ]
+        [ rdfs:label "sv" ; rdf:value "sv_SE.utf8" ]
+        [ rdfs:label "en" ; rdf:value "en_GB.utf8" ]
```
osma, line 40: Perhaps there could be a few more languages here, e.g. `fr`, `de`, `es`...or even all the ones for which locales were built?

dockerfiles/config/config-docker.ttl
```diff
@@ -0,0 +1,130 @@
...
+    skosmos:sparqlCollationEnabled false ;
+    # HTTP client configuration
+    skosmos:sparqlTimeout 20 ;
+    skosmos:httpTimeout 5 ;
+    # customize the service name
+    skosmos:serviceName "Skosmos" ;
+    # customize the base element. Set this if the automatic base url detection doesn't work. For example setups behind a proxy.
+    # skosmos:baseHref "http://localhost/Skosmos/" ;
+    # interface languages available, and the corresponding system locales
+    skosmos:languages (
```
osma, line 37: Same comment here as above, could add some more languages?

dockerfiles/docker-compose.yml
```diff
@@ -0,0 +1,52 @@
...
+      - fuseki
+      - fuseki-cache
+    volumes:
+      - type: bind
+        source: ../dockerfiles/config/config-docker-compose.ttl
+        target: /var/www/html/config.ttl
+      - ..:/var/www/html
+
+# once the environment has been created:
+# 1. create dataset
```
osma, line 49: Is this step necessary? I managed to get it up and running just by uploading stw.ttl

dockerfiles/Dockerfile.ubuntu
```diff
@@ -0,0 +1,78 @@
...
+
+RUN a2enmod rewrite
+RUN a2enmod expires
+
+RUN echo "ServerName localhost" >> /etc/apache2/apache2.conf
+
+# Install Composer
+RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
+
+# Install Skosmos
```
osma, line 62: I don't think we want to install Skosmos from a tarball here, even commented out... Instead it would perhaps be better to show how the Skosmos code can be copied from the repository itself? Though the docker-compose file solves this in another way, by mounting a volume (if I understood it correctly). In my opinion, it would perhaps be best to copy the Skosmos files inside the container, so that the container becomes independent of the host system after it's been created.

osma (changes requested):
I finally managed to test this. In addition to the specific comments, here are some general ones.

I tried following both the README file included in the PR, and the instructions in the PR comments. There were some issues with the README, as detailed below.

In general I got this working smoothly and I think this is very close to merging. If necessary I can do any remaining changes on my own - I don't expect that @kinow you are available to work on this anymore but of course I'm more than happy if you still can at least chime in :)


# dockerfiles/README.md

First, I'm trying to follow the instructions at the top of the README.md file to get the simple skosmos-web container built and running, without yet using docker-compose. Some issues I stumbled on:

## Path mismatch

Running the commands

    cd dockerfiles
    docker build -t skosmos:test . -f Dockerfile.ubuntu

fails at

    Step 9/17 : COPY dockerfiles/config/000-default.conf /etc/apache2/sites-available/000-default.conf
    COPY failed: stat /var/lib/docker/tmp/docker-builder950169892/dockerfiles/config/000-default.conf: no such file or directory

From Skosmos root directory, this works:

    docker build -t skosmos:test . -f dockerfiles/Dockerfile.ubuntu

So either the command in the README.md file or the paths in Dockerfile.ubuntu should be adjusted.

## Varnish not mentioned

The README file describes the skosmos and fuseki containers, but not the varnish container (fuseki-cache). This should be documented as well.

## Default page

The README says that when I've started the container:

> Now Skosmos should be available at `http://localhost:9090/`.

But when I open this URL, it simply shows the Apache2 Ubuntu default page. No Skosmos in sight! I think this is simply because Skosmos was never installed there in the first place, the relevant section is commented out in Dockerfile.ubuntu (for good reasons). Probably the README needs to be changed to explain how to get the Skosmos code inside the container.

# Ubuntu base image

The current image is based on Ubuntu 18.04 LTS (bionic). This is OK for now but at some point we probably want to upgrade to Ubuntu 20.04.

dockerfiles/README.md
kinow: Not necessary, I think I copied it from another image and kept the `README.md` structure. Let me remove this :+1:

kinow:
(rebased)

dockerfiles/docker-compose.yml
kinow: Oh, I think I had been doing that one out of costume, without confirming whether that was necessary or not :smile: removed.

dockerfiles/Dockerfile.ubuntu
kinow: Installing inside the container from git now. I used the latest tag @osma . WDYT? (BTW the latest tag doesn't follow the pattern of the previous ones... `v2.8` vs `v.2.9`)

kinow:
Hi @osma !

>I finally managed to test this. In addition to the specific comments, here are some general ones.

:tada: 

>In general I got this working smoothly and I think this is very close to merging. If necessary I can do any remaining changes on my own - I don't expect that @kinow you are available to work on this anymore but of course I'm more than happy if you still can at least chime in :)

I actually would like to finish it if possible. I don't mind if you or others commit directly here @osma , but otherwise I'm happy to update the PR based on testing feedback :)

>dockerfiles/README.md
>(…)
>Path mismatch

Ha! What a mess I did with the paths. That smells to me like I was in the mode write-rebuild-tryagain when I wrote that, sorry. I've pushed a commit — hopefully — fixing it. https://github.com/kinow/Skosmos/commit/9770ecedd311da992baf65c1354dc243daf93a0b

>Default page
>But when I open this URL, it simply shows the Apache2 Ubuntu default page. No Skosmos in sight! I think this is simply because Skosmos was never installed there in the first place, the relevant section is commented out in Dockerfile.ubuntu (for good reasons). Probably the README needs to be changed to explain how to get the Skosmos code inside the container.

I had a similar issue today when updating the image, and realized I never deleted the default `index.html`. I'm now removing it, before installing Skosmos. This way we don't run the risk of having `index.html` and `index.php` (I suppose `index.html` comes first in Apache's default `DirectoryIndex` directive, unless we change it).

>Ubuntu base image

Agree. I can take a look at it later. I quickly replaced `18.04` by `20.04` and the build failed due to different PHP package names. I think fixing it (we can probably look at the php container to see what it's using) should make the build pass without other changes :crossed_fingers:

dockerfiles/Dockerfile.ubuntu
osma: Good catch re: the tag name! Must have been a mistake during the release process. I just pushed a tag `v2.9` to try to fix this after the fact, but I guess it's best to leave the invalid tag there as removing it might break things.

Not sure I like the idea of installing the code from git. I think it would be better to copy the code from the current repo itself, in whatever state it is. This way if you want to try out a change, you can adjust the code and test it in a container, without even having to do a commit, never mind pushing it to the main Skosmos repo (and getting it merged etc).

dockerfiles/README.md
```diff
@@ -0,0 +1,61 @@
...
+The second container created is `skosmos-web`, using the same image mentioned
+in the previous section. The only difference is that we bind a new Skosmos
+configuration `config/config-docker-compose.ttl` on `/var/www/html/config.ttl`.
+
+This `config-docker-compose.ttl` file uses `http://fuseki:3030/ds/sparql` for
+`skosmos:sparqlEndpoint`. You can use this example to start Skosmos pointing
+to any existing Apache Jena, preferably with the Jena Text extension.
+
+Note that `fuseki:3030` is the internal Docker network socket. To the host
+machine, Docker Compose is exposing it as `fuseki:9030` to avoid conflicts
```
osma, line 48: I would rather say it's exposed as `http://localhost:9030` because that's where you can access it using a browser, right?

dockerfiles/config/config-docker.ttl
```diff
@@ -0,0 +1,146 @@
...
+    # whether to enable the spam honey pot or not, enabled by default
+    skosmos:uiHoneypotEnabled true ;
+    # default time a user must wait before submitting a form
+    skosmos:uiHoneypotTime 5 ;
+    # plugins to activate for the whole installation (including all vocabularies)
+    skosmos:globalPlugins () .
+
+# Skosmos vocabularies
+
+:ysa a skosmos:Vocabulary, void:Dataset ;
```
osma, line 89: Could we use the configuration for STW and UNESCO here as well? I'm a bit nervous about referring to the `api.dev.finto.fi` endpoint here. Of course you then have to give them some SPARQL endpoint URL that you assume exists...maybe a Fuseki running on the host system in port 3030?

In the past we've had situations where someone uses an old Skosmos version which has YSA/YSO in the default configuration, and that bombards our endpoint with heavy queries that used to work fine on some older Fuseki version but not with a newer version. (And believe me, this kind of thing where a particular query that used to be fast suddenly becomes very heavy after a Fuseki update has happened quite a few times!)

We should probably change the default `config.ttl` that is included with Skosmos as well - that's a separate task from this, but at least we shouldn't have more copies of a configuration file with YSA and YSO from the `api.dev.finto.fi` endpoint.

dockerfiles/config/config-docker.ttl
```diff
@@ -0,0 +1,146 @@
...
+@prefix : <#> .
+
+# Skosmos main configuration
+
+:config a skosmos:Configuration ;
+    # SPARQL endpoint
+    # a local Fuseki server is usually on localhost:3030
+    #skosmos:sparqlEndpoint <http://localhost:3030/ds/sparql> ;
+    # use the dev.finto.fi endpoint where the example vocabularies reside
+    skosmos:sparqlEndpoint <http://api.dev.finto.fi/sparql> ;
```
osma, line 23: Same here as below...could this be changed to `localhost:3030` or something?

dockerfiles/config/config-docker-compose.ttl
```diff
@@ -0,0 +1,110 @@
...
+    # customize the css by adding your own stylesheet
+    skosmos:customCss "resource/css/stylesheet.css" ;
+    # default email address where to send the feedback
+    skosmos:feedbackAddress "" ;
+    # email address to set as the sender for feedback messages
+    skosmos:feedbackSender "" ;
+    # email address to set as the envelope sender for feedback messages
+    skosmos:feedbackEnvelopeSender "" ;
+    # whether to display the ui language selection as a dropdown (useful for cases where there are more than 3 languages) 
+    skosmos:uiLanguageDropdown false ;
```
osma, line 79: Now that there are many UI languages defined, this should be set to `true`! Otherwise the UI looks very messy with all the language links...

dockerfiles/config/config-docker.ttl
```diff
@@ -0,0 +1,146 @@
...
+    # customize the css by adding your own stylesheet
+    skosmos:customCss "resource/css/stylesheet.css" ;
+    # default email address where to send the feedback
+    skosmos:feedbackAddress "" ;
+    # email address to set as the sender for feedback messages
+    skosmos:feedbackSender "" ;
+    # email address to set as the envelope sender for feedback messages
+    skosmos:feedbackEnvelopeSender "" ;
+    # whether to display the ui language selection as a dropdown (useful for cases where there are more than 3 languages) 
+    skosmos:uiLanguageDropdown false ;
```
osma, line 79: Same as above, should be changed to `true` since there are so many UI languages now

dockerfiles/docker-compose.yml
```diff
@@ -0,0 +1,48 @@
...
+    depends_on:
+      - fuseki
+      - fuseki-cache
+    volumes:
+      - type: bind
+        source: ./config/config-docker-compose.ttl
+        target: /var/www/html/config.ttl
+
+# once the environment has been created load a vocabulary, e.g.:
+# >curl -I -X POST -H Content-Type:text/turtle -T yso-skos.ttl -G http://localhost:9030/skosmos/data --data-urlencode graph=http://www.yso.fi/onto/yso/
```
osma, line 48: It would be better if this matched STW and UNESCO...and perhaps even better to put this in the README file instead?

osma (changes requested):
> Could you take a look at this latest version, please?

I tested this again. Excellent work!

I already commented about how the Skosmos code gets into the container. I think it should be copied from the surrounding repository, not from GitHub.

> Agree. I can take a look at it later. I quickly replaced 18.04 by 20.04 and the build failed due to different PHP package names. I think fixing it (we can probably look at the php container to see what it's using) should make the build pass without other changes crossed_fingers

Ubuntu 20.04 uses PHP 7.4. We still have some outstanding issues with PHP 7.4 - mainly EasyRdf (see PR #1128) and PHPUnit (see PR #1127). So it probably won't work out of the box yet. I will try to sort those out very soon, probably tomorrow. Then it should be easier to upgrade to 20.04.

dockerfiles/Dockerfile.ubuntu
kinow: Ah, that's even easier! Will change it today 👍

dockerfiles/config/config-docker-compose.ttl
kinow: Done!

dockerfiles/config/config-docker.ttl
kinow: Done!

dockerfiles/config/config-docker.ttl
kinow: The issue here is that this `config-docker.ttl` is used when you run only the Skosmos docker container, without a Fuseki endpoint. That means that by default anyone running this image would get an empty/blank Skosmos page, and have to understand that they need to re-create the configuration with the right endpoint… or mount a volume overwriting the `config.ttl`…

If we set to `localhost:3030`, then by default every container will show errors (in the console if enabled I think?) unless the user has a) a fuseki endpoint running and b) it is mapped to the Docker container so it can access it as localhost:3030 (I think it would have to be <fuseki-container-name>:3030 or some Docker-networking-fu trick; though not sure if using localhost inside the skosmos-web container would work here).

WDYT?

dockerfiles/config/config-docker.ttl
kinow: Sorry, I was reviewing the comments from the bottom up, and read this after I commented in the one below.

Let me see if I can find a way to map an external `<some-host-or-localhost>:3030` to a container created with this `Dockerfile.ubuntu` and `config-docker.ttl`.

If we manage to do that, we might be able to have a single `config-docker.ttl` used by both Docker command, and `docker-compose` command.

dockerfiles/README.md
kinow: Oh, nice catch. Fixed.

dockerfiles/docker-compose.yml
kinow: Moved to README, now using STW and UNESCO.

dockerfiles/config/config-docker.ttl
kinow: I've used `localhost:3030`, but had to change the command to use `--net=host`. That effectively means that the container started with `docker run ...` has access to the host network, and can access `localhost:3030`, which can be another container, a `jvm` process, etc.

I've also added notes telling users they can customize it, other ways to achieve the same without Docker Compose, like creating a network. I was going to write the complete example creating a network, but that would become quite extensive.

dockerfiles/docker-compose.yml
kinow: Done and done!

kinow:
Hi @osma 

I've update the branch after your feedback, thanks for testing it again so quickly. Notes:

- I've updated the `dockerfiles/README.md`. It now contains information for starting a Docker container that doesn't use Finto API, but requires the user to make `localhost:3030` available with a Fuseki instance with JenaText. The readme should contain enough in the same section so that users can start it and connect to another instance of a Fuseki container.
- Added a couple paragraphs about Varnish cache.
- In order to build the container using the source code, I had to update the instructions to call the Dockerfile from the project source folder (i.e. docker build ... -f dockerfiles/Dockerfile.ubuntu). I've added a note about it. But basically, when you do a `docker build .`, that sets the context as `.`. So the container cannot “escape” the context and copy files from the parent folder (or any folder outside its context).

Ready for another round of testing :nerd_face: thanks @osma !

Bruno

osma:
Thanks a lot @kinow - very fast and efficient! This seems almost like we are doing time travel or going at warp speed since we are working from different time zones! I will take a new look today and also work on the PHP 7.4 issues.

kinow:
Thanks @osma! It's been really efficient these reviews! I tried today php 7.4 too, but had issues with a twig extension and had to give up. If you have a fix I'd be happy to give it a try since I'm on Ubuntu 20.04. Cheers!

osma:
Okay good to know. What Twig extension did you have problems with? I assume I will encounter the same problem soon as well, but I've approached this from the side of PHPUnit tests so I've not had Twig related problems - yet!

kinow:
> Okay good to know. What Twig extension did you have problems with? I assume I will encounter the same problem soon as well, but I've approached this from the side of PHPUnit tests so I've not had Twig related problems - yet!

I think there were two issues with twig. One was with twig-... ext-something I think? I think it was the only twig-something in composer.json.

Then I bumped twig up to 3.x I think. And added a... statement in composer.json that's like a maven exclusion or a yarn resolution... something like `excludes: { twig: * }`.

composer install succeeded, but visiting the landing page, I think vocab-shared.twig failed to parse due to the play statements? Hmm. A kind of twig macro that removes whitespaces.

I removed that statement (and the end--- closing statement) but then in another page concept-shared had the same problem.

Found a phpmyadmin issue that was similar and had been fixed, but I couldn't figure out if I had to alter code, php deps, or a ubunto pkg.

On mobile going to bed (sorry for brevity, typos and lack of formatting) but thought this brain dump could be useful in case you have some similar issue.

Cheers
Bruno

osma:
Thanks @kinow, and good night! :)

osma:
All the changes look good to me based on reading your comment above, and the current diff. But unfortunately I didn't have time today to try this for real, as I got stuck on the PHP 7.4 issues. I was able to merge an update of EasyRdf (PR #1128), discovered some bugs in the honeypot code used by the feedback form (PR #1131) and made good progress on getting the unit tests working on PHP 7.4 (#1127) but didn't yet finish that part. The code in current `master` should now work much better on PHP 7.4 than it used to but I can't guarantee that everything works. If you have time to test upgrading to the Ubuntu 20.04 base image, that would be helpful!

I have some other work that needs attending in the meantime, but we will continue the sprint next Tuesday and finish the coding work by Thursday (followed by testing and hopefully a release). This PR is already looking so good that I'm pretty sure it can be merged next week. We can keep using 18.04 as the base image if 20.04 turns out to be too tricky.

kinow:
Not a problem @osma. Thanks for the work on PHP 7.4, and for fixing the bugs in honeypot (introduced by me :sweat: ). I've made two notes in post-its while testing the Docker for `noscript` and an 404 error I was getting from a stylesheet. Will prepare the PR's in the meantime :+1:

kouralex:
> 
> 
> > Okay good to know. What Twig extension did you have problems with? I assume I will encounter the same problem soon as well, but I've approached this from the side of PHPUnit tests so I've not had Twig related problems - yet!
> 
> I think there were two issues with twig. One was with twig-... ext-something I think? I think it was the only twig-something in composer.json.
> 
> Then I bumped twig up to 3.x I think. And added a... statement in composer.json that's like a maven exclusion or a yarn resolution... something like `excludes: { twig: * }`.
> 
> composer install succeeded, but visiting the landing page, I think vocab-shared.twig failed to parse due to the play statements? Hmm. A kind of twig macro that removes whitespaces.
> 
> I removed that statement (and the end--- closing statement) but then in another page concept-shared had the same problem.
> 
> Found a phpmyadmin issue that was similar and had been fixed, but I couldn't figure out if I had to alter code, php deps, or a ubunto pkg.
> 
> On mobile going to bed (sorry for brevity, typos and lack of formatting) but thought this brain dump could be useful in case you have some similar issue.
> 
> Cheers
> Bruno

Twig cannot be upgraded to 3.x without some extra work as there are still some deprecated features used in the code base. This is why I, not long ago, only updated it to the latest 2.x version. [Here](https://twig.symfony.com/doc/2.x/deprecated.html) you have the list of things that must be addressed first. That said, I do not believe the main Twig library was to blame for the failure in the first place.

Ps. [as 3.x is only supported by PHP 7.2.5+](https://twig.symfony.com/doc/3.x/intro.html#prerequisites) and we still officially support PHP 7.1 (at least until @osma gets his work done with PHP 7.4 support, which is close now!), we may not proceed with upgrading anyways.

kinow:
Good to know @kouralex , that explains why I had so many issues after upgrading to 3.x.

osma:
> we still officially support PHP 7.1 (at least until @osma gets his work done with PHP 7.4 support, which is close now!), we may not proceed with upgrading anyways.

PHP 7.1 support is going away very soon anyway, with the PHPUnit upgrade in PR #1127. So there's really no reason to insist on it, and I for one would warmly welcome a PR for upgrading to Twig 3.x, even if it meant dropping PHP 7.1 support.

dockerfiles/README.md
```diff
@@ -0,0 +1,116 @@
+Dockerfiles for Skosmos.
+
+## Running with Docker
+
+The following commands will build and tag the image it with `skosmos:test`,
+and run the container. The container name is `skosmos-web`, but you can customize
+the name, port, and other flags as necessary.
+
+```bash
```
osma, line 9: This is a pretty minor detail, but IMHO this block syntax is a bit hard to read when looking at the raw Markdown file in e.g. a text editor. I suggest using the [indented code block syntax](https://www.markdownguide.org/basic-syntax/#code-blocks) instead (although that may lose the bash specific syntax highlighting).

dockerfiles/README.md
```diff
@@ -0,0 +1,116 @@
...
+with the vocabulary data, and to be available at `http://localhost:3030`.
+For this last requisite you must create a
+[Docker network](https://docs.docker.com/network/network-tutorial-standalone/),
+use [`--net=host`](https://docs.docker.com/network/host/) or other mechanisms for
+that. See the section [Running with docker-compose](#running-with-docker-compose)
+if you would like to use Docker Compose.
+
+To stop the container:
+
+```bash
```
osma, line 31: Same here as above, just indent the command

dockerfiles/README.md
```diff
@@ -0,0 +1,116 @@
...
+The Apache virtual host configuration is located at `config/000-default.conf`. And
+the configuration file used for Skosmos is at `config/config.ttl`. Customize these
+two files as necessary.
+
+**NOTE**: If you would like to start a Fuseki container to test with Docker only,
+without Docker Compose, you can try the following command before loading your
+vocabulary data. It starts a container in the same way our other example with
+the `docker-compose` command.
+
+```bash
```
osma, line 49: Ditto

dockerfiles/README.md
```diff
@@ -0,0 +1,116 @@
...
+pointing to any other existing Apache Jena server, preferably with the Jena Text
+extension.
+
+**NOTE**: `fuseki:3030` and `fuseki-cache:80` are from the internal Docker network.
+To the host machine Docker Compose is exposing these values as `localhost:3030`
+and `localhost:9031` respectively.
+
+To create the containers in this example setup, you can use this command:
+
+```bash
```
osma, line 87: Ditto

dockerfiles/README.md
```diff
@@ -0,0 +1,116 @@
...
+docker-compose up -d
+```
+
+Now Skosmos should be available at `http://localhost:9090/` from your
+host. See the [section below](#loading-vocabulary-data) to load vocabulary data.
+
+To stop:
+
+```bash
```
osma, line 96: Ditto

dockerfiles/README.md
```diff
@@ -0,0 +1,116 @@
...
+docker-compose down
+```
+
+## Loading vocabulary data
+
+After you have your container running, with either Docker or `docker-compose`,
+you will need to load your vocabulary data.
+
+```bash
```
osma, line 105: Ditto

dockerfiles/README.md
```diff
@@ -0,0 +1,116 @@
...
+# NOTE: the container copies the project sources during build, so the
+# context must be the parent directory, i.e. you MUST build the image
+# from the Skosmos source directory, not from $sources/dockerfiles/
+docker build -t skosmos:test . -f dockerfiles/Dockerfile.ubuntu
+docker run -d --rm --name skosmos-web --net=host skosmos:test
+```
+
+Now Skosmos should be available at `http://localhost/`. See the
```
osma, line 17: Using port 80 on the host is a bit awkward if you already have, say, Apache listening there (which is normally the case for me). Wasn't there a `-p` option for the `docker run` in an earlier version of this README that mapped the container port 80 to something else? I think it would be better to expose this as e.g. port 9090, as is done in the docker-compose setup.

dockerfiles/README.md
```diff
@@ -0,0 +1,116 @@
...
+as `skosmos:sparqlEndpoint`, forcing `skosmos-web` to go through the `fuseki-cache`
+for a better performance. You can customize this example setup to start Skosmos
+pointing to any other existing Apache Jena server, preferably with the Jena Text
+extension.
+
+**NOTE**: `fuseki:3030` and `fuseki-cache:80` are from the internal Docker network.
+To the host machine Docker Compose is exposing these values as `localhost:3030`
+and `localhost:9031` respectively.
+
+To create the containers in this example setup, you can use this command:
```
osma, line 85: Maybe specify here that you need to be in the `dockerfiles` directory? (unlike in the `docker build` case above, where you had to be at the root level)

dockerfiles/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+To stop:
+
+```bash
+docker-compose down
+```
+
+## Loading vocabulary data
+
+After you have your container running, with either Docker or `docker-compose`,
```
osma, line 102: In the docker-compose case, the ports need to be changed to 9030 in the commands below, otherwise they won't work (or will access the wrong Fuseki)

osma (changes requested):
I tested this again - mostly an exercise in debugging the documentation, the implementation itself looks very solid now! I gave a few suggestions on how to improve the README file.

Now that PR #1127 has been merged to master, we should have support for PHP 7.4, so it should be possible to test changing the Ubuntu base image to 20.04. Though I haven't tested actually running Skosmos on PHP 7.4 yet, it's still possible that things will break in places that are not exercised by the test, for example related to Twig.

dockerfiles/README.md
kinow: Yes, but not if you want the container to access host URLs like localhost:3030 (localhost being the host not container with skosmos)

There are alternatives though

- use a value like fuseki:3030 and ask user to link containers or create a network
- include fuseki in the image (would need 2 Dockerfiles, so docker compose  or AWS etc can deploy separately)
- there is another docker image that does this kind of host-container port mapping (from host to container). I haven't used it, but it'd be 1 more image for the standalone container

dockerfiles/README.md
kinow: Done :+1:

kinow:
@osma updated based on your feedback, also updated to Ubuntu 20.04. Left one reply to a question that needs your reply, but I think everything else was addressed. Thanks!

dockerfiles/README.md
osma: Ah, now I see - you have to use `--net=host` to make the container share the host network (so that it can access Fuseki on localhost) instead of sitting on a separate bridged network. So the Apache port inside the container (80) has to match what is seen on the outside, there is no room for a mapping layer.

I think this is OK, as long as it's documented - the README has to say that you can't have anything else listening on port 80 (such as Apache).

It should also be possible to switch the Apache in the container to another port by changing the Listen directive, but maybe that's not a great idea since 80 is the expected, standard port number.

I did find a [discussion about accessing services on the host](https://stackoverflow.com/questions/24319662/from-inside-of-a-docker-container-how-do-i-connect-to-the-localhost-of-the-mach) which mentioned the possibility of using `host.docker.internal` but apparently this only works on macOS and Windows, and the most recent versions of Docker on Linux (version [20.10.0 released 2020-12-08](https://docs.docker.com/engine/release-notes/#20101) included https://github.com/moby/moby/pull/40007). So that could be an option as well, but requires a fairly recent Docker installation - the versions that come pre-packaged with distros are usually older.

kinow: >Ah, now I see - you have to use --net=host to make the container share the host network (so that it can access Fuseki on localhost) instead of sitting on a separate bridged network. So the Apache port inside the container (80) has to match what is seen on the outside, there is no room for a mapping layer.

Exactly.

>I think this is OK, as long as it's documented - the README has to say that you can't have anything else listening on port 80 (such as Apache).

Added the following.

```diff
diff --git a/dockerfiles/README.md b/dockerfiles/README.md
index b09a1705..8d0e0ed5 100644
--- a/dockerfiles/README.md
+++ b/dockerfiles/README.md
@@ -4,7 +4,11 @@ Dockerfiles for Skosmos.
 
 The following commands will build and tag the image it with `skosmos:test`,
 and run the container. The container name is `skosmos-web`, but you can customize
-the name, port, and other flags as necessary.
+the name, and other flags as necessary. The container will listen to port
+`80` at both the host and container since we are using `--net=host` to allow the
+Skosmos application to access Fuseki at `http://localhost:3030`. You are free to
+modify the command line and the `Dockerfile.ubuntu` and configuration files if you
+would like to deploy it differently.
 
     # NOTE: the container copies the project sources during build, so the
     # context must be the parent directory, i.e. you MUST build the image

```

>It should also be possible to switch the Apache in the container to another port by changing the Listen directive, but maybe that's not a great idea since 80 is the expected, standard port number.

I think anyone that wants to deploy Skosmos to AWS/Google/Azure/etc should be able to understand what we did here, and modify the `Dockerfile` and configuration files to match their network/volume/processing requirements.

>I did find a discussion about accessing services on the host which mentioned the possibility of using host.docker.internal but apparently this only works on macOS and Windows, and the most recent versions of Docker on Linux (version 20.10.0 released 2020-12-08 included moby/moby#40007). So that could be an option as well, but requires a fairly recent Docker installation - the versions that come pre-packaged with distros are usually older.

I think we went down the same rabbit hole here :grinning: I tried the `host.docker.internal` but that didn't work, then found an [issue on github](https://github.com/docker/for-linux/issues/264) to add it to Linux Docker. We can review it later after new versions of Docker are released, or if a user or dev has feedback on how to improve it I think.

Thanks @osma!
Bruno

kinow:
(rebased, and included a new commit to document the `--net=host` and port `80` use)

osma:
Tested this once more, removed lines referring to Finto SPARQL endpoint from the config files (they were already commented out), and merged. :tada: Great work @kinow!

Now the [wiki page for Docker use](https://github.com/NatLibFi/Skosmos/wiki/Install-Skosmos-with-Fuseki-in-Docker) still needs to be updated. The README.md file included in this PR is already pretty extensive, so maybe it can mostly just be replaced with a link to that? @kinow do you want to do this or should I?

kinow:
> Tested this once more, removed lines referring to Finto SPARQL endpoint from the config files (they were already commented out), and merged. tada Great work @kinow!

:tada: thanks for the patience reviewing and testing it, and for the help along the way @osma, it was fun :-)

> Now the [wiki page for Docker use](https://github.com/NatLibFi/Skosmos/wiki/Install-Skosmos-with-Fuseki-in-Docker) still needs to be updated. The README.md file included in this PR is already pretty extensive, so maybe it can mostly just be replaced with a link to that? @kinow do you want to do this or should I?

I did a quick change, but feel free to amend/edit it as necessary @osma : https://github.com/NatLibFi/Skosmos/wiki/Install-Skosmos-with-Fuseki-in-Docker

osma:
> I did a quick change, but feel free to amend/edit it as necessary @osma : https://github.com/NatLibFi/Skosmos/wiki/Install-Skosmos-with-Fuseki-in-Docker

Thanks! I edited it slightly. All good now!

PR #964, opened 2020-03-31
Use a row for the bread crumb (as in other visual elements nearby) and include padding

### LINKED ISSUE #963: Bad layout of preferred term

http://finto.fi/yso/fi/page/p23265

atk

kinow:
Use a `<div class="row">` around the breadcrumb as in other elements to force it to occupy the whole row-area.

Would be easier if the elements were [flex](https://css-tricks.com/snippets/css/a-guide-to-flexbox/). Haven't used bootstrap in a while, but looks like it supports it too in the [version 4.x](https://getbootstrap.com/docs/4.0/utilities/flex/).

Closes #963 ~but needs a bit more of testing before it's ready for review.~ (tested with `docker-compose.yml`, and found nothing broken in the UI).

Applying changes locally in firefox development mode it looks as follows:

image

### DIFF
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -1159,6 +1159,8 @@ li.sub-group {
 .crumb-path {
   float:left;
   width: 100%;
+  padding-left: 15px;
+  padding-right: 15px;
 }
 
 .foundby {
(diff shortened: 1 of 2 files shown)


osma:
Wow thanks @kinow ! We were just discussing how to approach this but you got there first with your fix :)

osma:
Tested it locally, seems to work fine.

kinow:
> Wow thanks @kinow ! We were just discussing how to approach this but you got there first with your fix :)

Thanks for testing so quickly. Hopefully will fix the issue for the user that reported it (dev.finto appears to be working, but not sure if it was deployed already)

osma:
Yes the fix is now deployed in dev.finto.fi, as it follows the master branch

PR #968, opened 2020-04-03
Fix static analysis issues

kinow:
Left multiple commits for now in-case we need to drop some. But happy to squash (or OK if somebody else squashes them) commits. Had some time today while another thing was running, so went through the code, the static analysis issues reported in SonarCloud and PhpStorm and fixed a few ones.

Cheers
Bruno


kinow:
Hm, not sure why that unit test failed. Can't tell if related to this change or not.

osma:
Thanks @kinow! Looks good! I will test locally and if everything works fine after the changes, I will merge this. No need to squash - it's good to have small, descriptive commits.

One of the Travis jobs failed, probably just some random transient issue. I restarted it.

osma:
Couldn't find any issues, and the Travis CI job passed on the second attempt. Thanks again!

PR #969, opened 2020-04-08
Shorten long lists of property values, with link to expand

### LINKED ISSUE #833: Tediously long property lists for concepts in certain classification systems
http://dev.finto.fi/ykl/fi/page/38.55

When viewing a model of a large classification system in Skosmos, one might get hundreds of labels for the same property for one concept, in each language of the model. This crowds the concept view quite a lot.

Preferably the model would have been structured in a way that avoids situations where a concept has 400+ labels in various languages.

But when it happens, one solution would be adding a check in the relevant Skosmos twig template if there's a list of e.g. over 30 statements with the same property, the corresponding `<li>` element would have `max-height: 500px; overflow-y: scroll;` added to it. Of course, the maximum allowed number of statements without a scrollable list should depend on the font size.

osma:
This PR changes the way property values (e.g. skos:narrower or skos:related) are displayed when there are many of them. If there are more than 10 values (current hardcoded value), the list will be cut to the first 10 and a link `[show all X values]` will be shown after those. Example:

image

The message is translated - the PR includes Finnish, Swedish and English translations.

Fixes #833


osma:
Thanks @joelit ! Potential further work items before this is merged:

- [ ] increase the cutoff limit (currently set to 10 values)
- [ ] optimize the repeated jQuery expressions

PR #981, opened 2020-05-06
Issue640 method for altlabels

### LINKED ISSUE #640: Return altLabels too for REST API  "/vocid/label"
Hi,

I see that in the context of 'Concept-specific methods'  the "/vocid/label" allows you to retrieve the prefLabel of a term starting from the uri.

Is it possible to enrich the response in order to retrieve also the altLabel?

Maybe a query that allows to return prefLabel and altLable from the uri already exists, but I did not find it.

Could you please help?

Thank you

joelit:
Fixes #640

### DIFF
--- a/controller/RestController.php
+++ b/controller/RestController.php
@@ -709,29 +709,28 @@ class RestController extends Controller
             return null;
         }
 
-        $prefResults = $request->getVocab()->getConceptLabel($request->getUri(), $request->getLang());
-        if ($prefResults === null) {
+        $labelResults = $request->getVocab()->getOtherConceptLabels($request->getUri(), $request->getLang());
+        if ($labelResults === null) {
             return $this->returnError('404', 'Not Found', "Could not find concept <{$request->getUri()}>");
         }
-        $otherResults = $request->getVocab()->getOtherConceptLabels($request->getUri(), $request->getLang());
 
         $ret = array_merge_recursive($this->context, array(
             '@context' => array('prefLabel' => 'skos:prefLabel', 'altLabel' => 'skos:altLabel', 'hiddenLabel' => 'skos:hiddenLabel', '@language' => $request->getLang()),
             'uri' => $request->getUri())
         );
 
-        if (isset($prefResults[$request->getLang()])) {
-            $ret['prefLabel'] = $prefResults[$request->getLang()]->getValue();
+        if (!empty($labelResults['prefLabel'])) {
+            $ret['prefLabel'] = $labelResults['prefLabel'][0]->getValue();
         }
-        if (!empty($otherResults['altLabel'])) {
+        if (!empty($labelResults['altLabel'])) {
             $ret['altLabel'] = array();
-            foreach ($otherResults['altLabel'] as $altResult) {
+            foreach ($labelResults['altLabel'] as $altResult) {
                 $ret['altLabel'][] = $altResult->getValue();
             }
         }
-        if (!empty($otherResults['hiddenLabel'])) {
+        if (!empty($labelResults['hiddenLabel'])) {
             $ret['hiddenLabel'] = array();
-            foreach ($otherResults['hiddenLabel'] as $hidResult) {
+            foreach ($labelResults['hiddenLabel'] as $hidResult) {
                 $ret['hiddenLabel'][] = $hidResult->getValue();
             }
         }
--- a/controller/RestController.php
+++ b/controller/RestController.php
@@ -710,31 +710,21 @@ class RestController extends Controller
         }
 
         $labelResults = $request->getVocab()->getAllConceptLabels($request->getUri(), $request->getLang());
-
         if ($labelResults === null) {
             return $this->returnError('404', 'Not Found', "Could not find concept <{$request->getUri()}>");
         }
 
-        $ret = array_merge_recursive($this->context, array(
-            '@context' => array('prefLabel' => 'skos:prefLabel', 'altLabel' => 'skos:altLabel', 'hiddenLabel' => 'skos:hiddenLabel', '@language' => $request->getLang()),
-            'uri' => $request->getUri())
-        );
+        if (!isset($labelResults['prefLabel']) || empty($labelResults['prefLabel'])) {
+            return $this->returnError('404', 'Not Found', "Could not find preferred label for concept <{$request->getUri()}>");
+        }
 
-        if (!empty($labelResults['prefLabel'])) {
-            $ret['prefLabel'] = $labelResults['prefLabel'][0]->getValue();
-        }
-        if (!empty($labelResults['altLabel'])) {
-            $ret['altLabel'] = array();
-            foreach ($labelResults['altLabel'] as $altResult) {
-                $ret['altLabel'][] = $altResult->getValue();
-            }
-        }
-        if (!empty($labelResults['hiddenLabel'])) {
-            $ret['hiddenLabel'] = array();
-            foreach ($labelResults['hiddenLabel'] as $hidResult) {
-                $ret['hiddenLabel'][] = $hidResult->getValue();
-            }
-        }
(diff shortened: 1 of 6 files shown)


controller/RestController.php
```diff
@@ -683,18 +683,31 @@ public function label($request)
             return $this->returnError(400, "Bad Request", "uri parameter missing");
         }
 
-        $results = $request->getVocab()->getConceptLabel($request->getUri(), $request->getLang());
-        if ($results === null) {
+        $prefResults = $request->getVocab()->getConceptLabel($request->getUri(), $request->getLang());
+        if ($prefResults === null) {
             return $this->returnError('404', 'Not Found', "Could not find concept <{$request->getUri()}>");
         }
+        $otherResults = $request->getVocab()->getOtherConceptLabels($request->getUri(), $request->getLang());
```
osma, line 690: This code is doing triggering two SPARQL queries (one for the prefLabel, another for altLabel+hiddenLabel). Wouldn't it be better to use just a single SPARQL query that asks for all three label properties at once? Simply adding skos:prefLabel to the VALUES block should be enough. That would be more efficient.

model/sparql/GenericSparql.php
```diff
@@ -1402,6 +1402,27 @@ private function generateLabelQuery($uri, $lang) {
...
+    /**
+     * @param string $uri
+     * @param string $lang
+     * @return string sparql query string
+     */
+    private function generateOtherLabelQuery($uri, $lang) {
+        $fcl = $this->generateFromClause();
+        $labelcondLabel = ($lang) ? "FILTER( langMatches(lang(?val), '$lang') )" : "";
+        $query = <<<EOQ
+CONSTRUCT { <$uri> ?prop ?val . } $fcl
```
osma, line 1415: It's not wrong to use a CONSTRUCT here, but the implied RDF serialization (Fuseki) and parsing (EasyRdf) seems like unnecessary overhead. This could easily be changed into `SELECT ?prop ?val` instead.

model/sparql/GenericSparql.php
```diff
@@ -1429,6 +1450,34 @@ public function queryLabel($uri, $lang) {
...
+            foreach($altLabels as $label) {
+                $ret['altLabel'][] = $label;
+            }
+        }
+        if (!empty($hiddenLabels)) {
+            $ret['hiddenLabel'] = array();
+            foreach($hiddenLabels as $label) {
+                $ret['hiddenLabel'][] = $label;
+            }
+        }return $ret;
```
osma, line 1478: Missing a line break here?

controller/RestController.php
joelit: Good point! No need to use the old label query in this case at all.

model/sparql/GenericSparql.php
joelit: Can do. This was for my own convenience since listing the literals from the response graph seemed the simplest. I'll change it to an iterable result.

model/sparql/GenericSparql.php
```diff
@@ -1429,6 +1450,55 @@ public function queryLabel($uri, $lang) {
...
+    public function queryOtherLabels($uri, $lang) {
+        $query = $this->generateOtherLabelQuery($uri, $lang);
+        $result = $this->query($query);
+        $ret = array();
+
+        $prefLabels = array();
+        $altLabels = array();
+        $hiddenLabels = array();
+
+        foreach ($result as $row) {
```
osma, line 1468: How about replacing the contents of the `foreach` loop with something like:

```php
$lname = $row->prop->localName();
if (!in_array($lname, $ret) {
    $ret[$lname] = array();
}
$ret[$lname][] = $row->val;
```

Then you don't need the arrays `$prefLabels`, `$altLabels` and `$hiddenLabels` at all, and can also drop the three `if (!empty(...))` clauses below.

joelit: That is certainly much more elegant, thanks for the suggestion.

joelit: I did the changes you recommended. It seems much cleaner now.

controller/RestController.php
```diff
@@ -683,18 +683,30 @@ public function label($request)
...
         $ret = array_merge_recursive($this->context, array(
-            '@context' => array('prefLabel' => 'skos:prefLabel', '@language' => $request->getLang()),
+            '@context' => array('prefLabel' => 'skos:prefLabel', 'altLabel' => 'skos:altLabel', 'hiddenLabel' => 'skos:hiddenLabel', '@language' => $request->getLang()),
             'uri' => $request->getUri())
         );
 
-        if (isset($results[$request->getLang()])) {
-            $ret['prefLabel'] = $results[$request->getLang()]->getValue();
+        if (!empty($labelResults['prefLabel'])) {
+            $ret['prefLabel'] = $labelResults['prefLabel'][0]->getValue();
```
osma, line 697: I think the `->getValue()` call is in the wrong layer here. It should be already inside GenericSparql, so that there is no need to pass EasyRdf Literal objects from the model all the way to the controller.

model/sparql/GenericSparql.php
```diff
@@ -1429,6 +1449,35 @@ public function queryLabel($uri, $lang) {
...
+        $query = $this->generateAllLabelsQuery($uri, $lang);
+        $result = $this->query($query);
+        $ret = array();
+
+        foreach ($result as $row) {
+            $labelName = $row->prop->localName();
+            if (!in_array($labelName, $ret)) {
+                $ret[$labelName] = array();
+            }
+            $ret[$labelName][] = $row->val;
```
osma, line 1468: The `->getValue()` call should be already here

controller/RestController.php
osma: Also, if there are no `getValue()` calls here, then there should also be no need for the `if (!empty(...))` checks and copying of arrays. `$labelResults` can just be returned directly I think

model/sparql/GenericSparql.php
```diff
@@ -1429,6 +1449,35 @@ public function queryLabel($uri, $lang) {
...
+     * @return array array of prefLabels, altLabels and hiddenLabels - or null if resource doesn't exist
+     */
+    public function queryAllConceptLabels($uri, $lang) {
+        $query = $this->generateAllLabelsQuery($uri, $lang);
+        $result = $this->query($query);
+        $ret = array();
+
+        foreach ($result as $row) {
+            $labelName = $row->prop->localName();
+            if (!in_array($labelName, $ret)) {
```
osma, line 1465: This `if` block may actually be completely redundant, since PHP is quite relaxed about creating new arrays on the fly...so just assigning to `$ret[$labelName][]` (as is done after the `if` block) should be enough, there's no need to create the array explicitly.

controller/RestController.php
joelit: Good point. I changed it to pass only stings for the controller. I suppose the empty checks are sort of unnecessary - my design intent was to only give out data that is there:

```
"uri":"http://www.yso.fi/onto/yso/p1191",
"prefLabel":"aallot",
"hiddenLabel":["aalto"]
```

instead of:
```
"uri":"http://www.yso.fi/onto/yso/p1191",
"prefLabel":"aallot",
"altLabel":[] (or "altLabel": null)
"hiddenLabel":["aalto"]
```

But I suppose is is sort of explicit in the latter example that the method tried to find alt labels, but didn't find them. I'll make the changes.

osma: I understand the intent. But this is already handled by the `foreach` loop. If there are no hiddenLabels in the result set, there will be no `hiddenLabel` key in the array since it will never get added. No need for an explicit `if (!count(...))` check.

joelit: True. I realised the simplest thing to do was reassigning the value of prefLabel key, and otherwise just merging the $labelResult into the $ret.

PR #982, opened 2020-05-06
Initial draft for listing new and modified concepts for a vocabulary in the REST API

joelit:
Output can be seen e.g.
`curl -X GET --header 'Accept: application/json' http://localhost/Skosmos/rest/v1/yso/modifiedConcepts?lang=en'`

### DIFF
--- a/controller/RestController.php
+++ b/controller/RestController.php
@@ -1095,11 +1095,14 @@ class RestController extends Controller
         //$this->setLanguageProperties($request->getLang());
         $vocab = $request->getVocab();
         $offset = ($request->getQueryParam('offset') && is_numeric($request->getQueryParam('offset')) && $request->getQueryParam('offset') >= 0) ? $request->getQueryParam('offset') : 0;
-        $contentLang = ($request->getContentLang() != null) ? $request->getContentLang() : $request->getLang();
-        $changeList = $vocab->getChangeList($prop, $contentLang, $request->getLang(), $offset);
+        $changeList = $vocab->getChangeList($prop, $request->getLang(), $offset);
 
         return $this->returnJson(array_merge_recursive($this->context,
-                                                        array('@context' => array('@language' => $request->getLang())),
+                                                        array('@context' => array( '@language' => $request->getLang(),
+                                                                                     'prefLabel' => 'skos:prefLabel',
+                                                                                     'xsd' => 'http://www.w3.org/2001/XMLSchema#',
+                                                                                     'date' => array( '@id' => 'http://purl.org/dc/terms/date', '@type' => 'http://www.w3.org/2001/XMLSchema#date') )
+                                                        ),
                                                         array('changeList' => $changeList)));
 
     }
--- a/controller/RestController.php
+++ b/controller/RestController.php
@@ -1071,7 +1071,7 @@ class RestController extends Controller
      */
     public function newConcepts($request)
     {
-        return $this->changed($request);
+        return $this->changedConcepts($request, 'dc:created');
     }
 
     /**
@@ -1081,7 +1081,7 @@ class RestController extends Controller
      */
     public function modifiedConcepts($request)
     {
-        return $this->changed($request, 'dc:modified');
+        return $this->changedConcepts($request, 'dc:modified');
     }
 
     /**
@@ -1089,11 +1089,12 @@ class RestController extends Controller
      * @param Request $request
      * @return object json-ld wrapped list of changed concepts
      */
-    private function changed($request, $prop='dc:created')
+    private function changedConcepts($request, $prop)
     {
         $vocab = $request->getVocab();
         $offset = ($request->getQueryParam('offset') && is_numeric($request->getQueryParam('offset')) && $request->getQueryParam('offset') >= 0) ? $request->getQueryParam('offset') : 0;
-        $changeList = $vocab->getChangeList($prop, $request->getLang(), $offset);
+        $limit = ($request->getQueryParam('limit') && is_numeric($request->getQueryParam('limit')) && $request->getQueryParam('limit') >= 0) ? $request->getQueryParam('limit') : 200;
+        $changeList = $vocab->getChangeList($prop, $request->getLang(), $offset, $limit);
 
         $simpleChangeList = array();
         foreach($changeList as $conceptInfo) {
@@ -1107,7 +1108,7 @@ class RestController extends Controller
                                                         array('@context' => array( '@language' => $request->getLang(),
                                                                                      'prefLabel' => 'skos:prefLabel',
                                                                                      'xsd' => 'http://www.w3.org/2001/XMLSchema#',
-                                                                                     'date' => array( '@id' => 'http://purl.org/dc/terms/date', '@type' => 'http://www.w3.org/2001/XMLSchema#date') )
+                                                                                     'date' => array( '@id' => 'http://purl.org/dc/terms/date', '@type' => 'http://www.w3.org/2001/XMLSchema#dateTime') )
                                                         ),
                                                         array('changeList' => $simpleChangeList)));
 
--- a/controller/RestController.php
+++ b/controller/RestController.php
@@ -1063,4 +1063,44 @@ class RestController extends Controller
         $ret = $this->transformPropertyResults($request->getUri(), $request->getLang(), $related, "related", "skos:related");
         return $this->returnJson($ret);
     }
+
+    /**
+     * Used for querying new concepts in the vocabulary
+     * @param Request $request
+     * @return object json-ld wrapped list of changed concepts
+     */
+    public function newConcepts($request)
+    {
+        return $this->changed($request);
+    }
+
+    /**
+     * Used for querying modified concepts in the vocabulary
+     * @param Request $request
+     * @return object json-ld wrapped list of changed concepts
+     */
+    public function modifiedConcepts($request)
+    {
+        return $this->changed($request, 'dc:modified');
+    }
+
+    /**
+     * Used for querying changed concepts in the vocabulary
+     * @param Request $request
+     * @return object json-ld wrapped list of changed concepts
+     */
+    private function changed($request, $prop='dc:created')
+    {
+        // set language parameters for gettext
+        //$this->setLanguageProperties($request->getLang());
+        $vocab = $request->getVocab();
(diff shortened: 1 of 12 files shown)


controller/RestController.php
```diff
@@ -1002,4 +1002,44 @@ public function related($request)
...
+     * @param Request $request
+     * @return object json-ld wrapped list of changed concepts
+     */
+    private function changed($request, $prop='dc:created')
+    {
+        // set language parameters for gettext
+        //$this->setLanguageProperties($request->getLang());
+        $vocab = $request->getVocab();
+        $offset = ($request->getQueryParam('offset') && is_numeric($request->getQueryParam('offset')) && $request->getQueryParam('offset') >= 0) ? $request->getQueryParam('offset') : 0;
+        $contentLang = ($request->getContentLang() != null) ? $request->getContentLang() : $request->getLang();
```
osma, line 1037: I don't think this method needs a separate contentLang setting, because as a REST method it doesn't really return any UI strings, just concept labels. So I'd just use `$request->getLang()` here for all language settings

rest.php
```diff
@@ -95,6 +95,10 @@
             $controller->groups($request);
         } elseif ($parts[2] == 'groupMembers') {
             $controller->groupMembers($request);
+        } elseif ($parts[2] == 'newConcepts') {
```
osma, line 98: This is a matter of taste but I think `new` and `modified` would be better REST method names. Also, `new` matches with the URL of the corresponding web page (e.g. http://finto.fi/yso/en/new)

controller/RestController.php
joelit: Yep, the existing method parameters looked really odd for this purpose. I will write a more general-purpose method for the REST API that takes only one language parameter and returns the modified concepts by timestamp, in contrast to the existing version.

controller/RestController.php
```diff
@@ -1063,4 +1063,53 @@ public function related($request)
...
+    {
+        return $this->changed($request, 'dc:modified');
+    }
+
+    /**
+     * Used for querying changed concepts in the vocabulary
+     * @param Request $request
+     * @return object json-ld wrapped list of changed concepts
+     */
+    private function changed($request, $prop='dc:created')
```
osma, line 1092: I think it would better to omit the `$prop` default value here and instead pass the property it explicitly from both `newConcepts` and `modifiedConcepts`. That would be more symmetrical.

The parameter `$prop` is missing a docstring definition.

Minor style comment: renaming the function to `changedConcepts` would make it better in line with `newConcepts` and `modifiedConcepts`.

controller/RestController.php
```diff
@@ -1063,4 +1063,53 @@ public function related($request)
...
+                $simpleChangeList[] =  array( 'uri' => $conceptInfo['uri'],
+                                               'prefLabel' => $conceptInfo['prefLabel'],
+                                               'date' => $conceptInfo['date']->format("Y-m-d\TH:i:sO") );
+            }
+        }
+        return $this->returnJson(array_merge_recursive($this->context,
+                                                        array('@context' => array( '@language' => $request->getLang(),
+                                                                                     'prefLabel' => 'skos:prefLabel',
+                                                                                     'xsd' => 'http://www.w3.org/2001/XMLSchema#',
+                                                                                     'date' => array( '@id' => 'http://purl.org/dc/terms/date', '@type' => 'http://www.w3.org/2001/XMLSchema#date') )
```
osma, line 1110: The datatype is set to `xsd:date`, but it seems to me that the actual values are formatted as `xsd:dateTime` (see format string above). So maybe the `@type` should be changed to `xsd:dateTime` ?

controller/WebController.php
```diff
@@ -559,22 +559,34 @@ public function invokeGenericErrorPage($request, $message = null)
...
-                'vocab' => $vocab,
+                'vocab' => $request->getVocab(),
                 'languages' => $this->languages,
                 'request' => $request,
-                'changeList' => $changeList)
+                'changeList' => $bydate)
             );
     }
 
+    public function formatChangeList($request, $prop='dc:created') {
```
osma, line 577: This should be a private function don't you think?

Also here I would not use a default value for `$prop` since it could be either `dc:created` or `dc:modified`, with roughly equal probability

osma: Also, the name of the method seems to imply that the function only performs formatting, but it actually also fetches the change list. Either change it to `getFormattedChangeList`, or - perhaps even better - move the call to `$vocab->getChangeList` outside the method, so that it only performs formatting by date.

osma:
Other comments on the PR:

- Swagger spec still needs to be done
- I wonder if the methods should support setting `limit` and `offset`, like some other REST methods (e.g. `search`, `index/X`)

controller/WebController.php
joelit: Good point, I will split the method in two. I left the function as public so that it could be included in the unit tests. Would there be a better way?

controller/RestController.php
joelit: Fixed. Originally my intention was to support both dates and dateTimes (one being subclass of the other), but as now the timestamps are formatted all as dateTimes, I changed this accordingly to dateTime.

controller/RestController.php
joelit: Good point, I made the changes you suggest.

osma:
Scrutinizer complains about the getChangeList method docstring:
> * @param string $lang UI language for the dates
> There is no parameter named $lang. Was it maybe removed?

Method signature:
> public function getChangeList($prop, $clang, $offset, $limit)

osma:
Postponing to the next release, will release 2.6 without this PR since there are still things to improve

controller/WebController.php
```diff
@@ -559,22 +559,54 @@ public function invokeGenericErrorPage($request, $message = null)
      */
     public function invokeChangeList($request, $prop='dc:created')
     {
-        // set language parameters for gettext
-        $this->setLanguageProperties($request->getLang());
-        $vocab = $request->getVocab();
-        $offset = ($request->getQueryParam('offset') && is_numeric($request->getQueryParam('offset')) && $request->getQueryParam('offset') >= 0) ? $request->getQueryParam('offset') : 0;
-        $changeList = $vocab->getChangeList($prop, $request->getContentLang(), $request->getLang(), $offset);
+        $bydate = $this->formatChangeList($request, $prop);
```
osma, line 562: I think it would be cleaner to call getChangeList here in invokeChangeList, instead of within formatChangeList.
So something like

```php
$changeList = $this->getChangeList($request, $prop); // extract $offset and $limit from $request within getChangeList
$bydate = $this->formatChangeList($changeList, $request->getLang());
```

This would also mean less data being passed around as parameters.

model/Vocabulary.php
```diff
@@ -615,19 +615,15 @@ public function verifyVocabularyLanguage($lang)
...
+    public function getChangeList($prop, $clang, $offset, $limit)
     {
-      $changelist = $this->getSparql()->queryChangeList($clang, $offset, $prop);
-      $bydate = array();
-      foreach($changelist as $concept) {
-        $concept['datestring'] = Punic\Calendar::formatDate($concept['date'], 'medium', $lang);
-        $bydate[Punic\Calendar::getMonthName($concept['date'], 'wide', $lang, true) . Punic\Calendar::format($concept['date'], ' y', $lang) ][strtolower($concept['prefLabel'])] = $concept;
-      }
-      return $bydate;
+      return $this->getSparql()->queryChangeList($clang, $offset, $prop, $limit);
```
osma, line 626: I can't help noticing that the parameter order for queryChangeList is completely different from getChangeList for no obvious reason. I suggest changing the parameter order to `$prop, $clang, $offset, $limit` which seems logical (e.g. offset and limit are next to each other)

swagger.json
```diff
@@ -1121,6 +1121,116 @@
...
+          "application/ld+json"
+        ],
+        "responses": {
+          "200": {
+            "description": "list of most recently created concepts of the vocabulary",
+            "schema": {
+              "$ref": "#/definitions/changedConceptsResult"
+            }
+          },
+          "304": {
```
osma, line 1167: I see no indication that 304 responses would be supported by the `new` method? If they are not supported, then this response code should be omitted from the Swagger spec.

swagger.json
```diff
@@ -1121,6 +1121,116 @@
...
```
osma, line 1222: Same as above

model/sparql/GenericSparql.php
```diff
@@ -2243,7 +2243,11 @@ private function transformChangeListResults($result) {
             }
 
             if (isset($row->date)) {
-                $concept['date'] = $row->date->getValue();
+                try {
+                    $concept['date'] = $row->date->getValue();
+                } catch (Exception $e) {
+                    //don't do anything to malformed dates e.g. 1986-21-00
```
osma, line 2249: There should be a unit test that checks that omitting the `$concept['date']` value in the case of malformed dates - as is done here - doesn't cause further problems upstream, for example in the `formatChangeList` method.

tests/RestControllerTest.php
```diff
@@ -340,4 +340,33 @@ public function testLabelNoConcept() {
       $this->assertEquals($expected, $out);
   }
 
+  /**
+   * @covers RestController::newConcepts
+   */
+  public function testNewConcepts() {
```
osma, line 346: The test name is confusing. It's called `testNewConcepts` but it actually calls `modifiedConcepts`. Also the `@covers` annotation above is wrong. In any case, there should be tests like this both for `newConcepts` and `modifiedConcepts`.

model/Vocabulary.php
joelit: The inconsistency of the parameter order was running pretty deep in the method call stack. Fixed it, thanks for noticing it!

model/sparql/GenericSparql.php
```diff
@@ -2207,11 +2207,13 @@ public function listConceptGroupContents($groupClass, $group, $lang,$showDepreca
 
     /**
      * Generates the sparql query for queryChangeList.
-     * @param string $lang language of labels to return.
+     * @param string $property the property uri pointing to timestamps, eg. 'dc:modified'
```
osma, line 2210: This should be `$prop` not `$property`, to match the variable declaration below

model/sparql/GenericSparql.php
```diff
@@ -2253,12 +2261,15 @@ private function transformChangeListResults($result) {
 
     /**
      * return a list of recently changed or entirely new concepts
+     * @param string $property the property uri pointing to timestamps, eg. 'dc:modified'
```
osma, line 2264: same as above

PR #983, opened 2020-05-08
Global REST label method

### LINKED ISSUE #952: Global REST label method
We have a global `data` method which can return RDF data for any URI regardless of its vocabulary, but no way to make a similar query for a label for a URI.

This came up recently because Annif returns YSO URIs, but they can be either from YSO or YSO Places. The URIs look the same. If you need a label for the URI, you could use the `label` method, but you need to know which vocabulary to target.

We could add a global `label` method, that works similarly to the global `data` method (guess the vocabulary based on URI space) but returning label data like the vocabulary-specific `label` method.

osma:
Fixes #952

Still missing Swagger spec. Creating a draft PR to get feedback from QA tools.

### DIFF
--- a/controller/RestController.php
+++ b/controller/RestController.php
@@ -709,7 +709,15 @@ class RestController extends Controller
             return null;
         }
 
-        $labelResults = $request->getVocab()->getAllConceptLabels($request->getUri(), $request->getLang());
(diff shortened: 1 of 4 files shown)


joelit:
I tested this locally, and it looked good!

PR #984, opened 2020-05-11
HTTP date format fixes

osma:
While testing HTTP 304 / If-Modified-Since functionality, recently implemented in PR #869 by @kinow, I noticed that the date formats expected by Skosmos in If-Modified-Since headers and produced in Last-Modified headers are not actually compliant with the relevant RFCs (RFC 7231 / 2616 / 1123). I also noticed other problems related to that functionality. This PR fixes:

- Read the If-Modified-Since header using the `filter_input` function instead of `$_SERVER` (possibly more secure, and in line with other places in the codebase where HTTP headers are accessed)
- Change to RFC 1123 date format for HTTP 304 functionality
- Improve the Http304Test unit tests so that they verify that the mocked methods are actually called exactly once with the expected arguments (otherwise they couldn't detect invalid date formats)

Any comments @kinow before I merge this?

### DIFF
--- a/controller/Controller.php
+++ b/controller/Controller.php
@@ -293,7 +293,7 @@ class Controller
     {
         if ($modifiedDate) {
             $ifModifiedSince = $this->getIfModifiedSince();
-            $this->sendHeader("Last-Modified: " . $modifiedDate->format('Y-m-d H:i:s'));
+            $this->sendHeader("Last-Modified: " . $modifiedDate->format('D, d M Y H:i:s \G\M\T'));
             if ($ifModifiedSince !== null && $ifModifiedSince >= $modifiedDate) {
                 $this->sendHeader("HTTP/1.0 304 Not Modified");
                 return true;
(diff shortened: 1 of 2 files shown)


osma:
Verified with REDbot.org.

Current situation (before the merge):
Last-Modified: 2020-05-12 05:36:10

kinow:
>Verified with REDbot.org.

Bookmarked! Great tool, thanks!

osma:
After the merge:

> Last-Modified: Tue, 12 May 2020 06:01:45 GMT
>
> If-Modified-Since conditional requests are supported.

(and the error is gone)

> Bookmarked! Great tool, thanks!

Yes, it's good - but the CAPTCHAs are tiring. They didn't use them before but probably there was some abuse they needed to counter

kinow:
>Yes, it's good - but the CAPTCHAs are tiring. They didn't use them before but probably there was some abuse they needed to counter

I'd like to enable it for the build of some tools. They have the source code available https://github.com/mnot/redbot, they even provide a Docker image. So in theory it should be possible to spin up a Docker instance with the latest SKOSMOS deployed, and another container with REDbot.

And then just use curl or something to query the REDbot container giving the URL to the SKOSMOS container :thinking:

osma:
> I'd like to enable it for the build of some tools I am working on.

Hmm - in my understanding it's an interactive debugging tool, not something to automate. But maybe I've missed some aspect?

kinow:
>Hmm - in my understanding it's an interactive debugging tool, not something to automate. But maybe I've missed some aspect?

No, I think you are correct. 

But assuming it's possible to create a `curl` call to send an HTTP request to a locally deployed REDbot instance, then I think it means it would be possible to also automate it in the build process :nerd_face:

PR #989, opened 2020-06-08
foreign language/explicit language detection improved;fixes #903

### LINKED ISSUE #903: Missing explicit/foreign language langtag for grouping concepts
http://skosmos.dev.finto.fi/gacs/fi/page/?clang=en&uri=C21092
1. Go to page
2. Observe no (fi) after plants in "Kuuluu ryhmään"

I'd like to see (fi) after plants.

kouralex:
This PR uses the same logic for checking foreign language/explicit language setting for groupingConcepts as it is already done for other properties in the Twig template.

### DIFF
--- a/view/concept-shared.twig
+++ b/view/concept-shared.twig
@@ -119,7 +119,7 @@
           {% for grouppath in concept.groupProperties %}
               <li>
               {% for group in grouppath %}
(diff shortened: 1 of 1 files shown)


joelit:
Looks good, sorry I didn't notice this earlier as it wasn't in our kanban!

PR #991, opened 2020-06-08
set explicit language setting via twig to mappings; fixes #990

### LINKED ISSUE #990: AJAX-loaded mapped properties do not respect skosmos:explicitLanguage setting
Currently, setting the `skosmos:explicitLanguage` property to `true` has no effect for the AJAX-loaded mapped properties on a concept page. There seems to be an issue in how it is resolved in the Twig template. Just look at the code below:
https://github.com/NatLibFi/Skosmos/blob/d110308446fb441cd56796cb10a9c2c7e556f2d1/view/concept-shared.twig#L201-L224

.. and in the corresponding function

https://github.com/NatLibFi/Skosmos/blob/d110308446fb441cd56796cb10a9c2c7e556f2d1/resource/js/scripts.js#L373-L383

It seems like there are two options for a possible fix:

1) pass `explicit_languages` variable via twig to the function
or
2) declare a new global JS variable (as it is currently done in https://github.com/NatLibFi/Skosmos/blob/master/view/scripts.twig) responsible for the value of the `skosmos:explicitLanguage` property and use it directly from within the `scripts.js` file.

Second phase is to fix the logic in the `scripts.js` file (expect the value of `explicitLangCodes` parameter to be a boolean).

kouralex:
This resolves #990 via firstly mentioned method.

### DIFF
--- a/resource/js/scripts.js
+++ b/resource/js/scripts.js
@@ -372,10 +372,7 @@ function renderPropertyMappings(concept, contentLang, properties) {
   });
   Handlebars.registerHelper('ifDifferentLabelLang', function(labelLang, explicitLangCodes, opts) {
     if (labelLang !== undefined && labelLang !== '' && labelLang !== null) {
-      if (explicitLangCodes !== undefined && typeof explicitLangCodes === "boolean") {
-        return opts.fn(explicitLangCodes);
-      }
-      if (labelLang !== contentLang) {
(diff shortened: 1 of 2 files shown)


osma:
I propose closing this PR and opening another that uses the second method.

joelit:
This was addressed in PR #998 so this PR is not necessary.

PR #992, opened 2020-06-09
Make property order configurable on concept page

### LINKED ISSUE #554: Configurable order of properties in Concept pages
Some thesaurus managers would like to have the ordering of properties in the Concept page according to what ISO-25964 recommends, which I am copy-pasting below :

> 12.2.2 Single record display
> When the record for a concept or a preferred term is viewed individually the preferred term should normally be displayed first and the other fields are customarily presented after it in the following sequence:
> a) CC concept code or notation, locating the term/concept in a hierarchical sequence;
> b) SN scope note;
> c) UF references to non-preferred equivalent term(s);
> d) TT references to top terms;
> e) BT references to broader terms;
> f) NT references to narrower terms;
> g) RT references to related terms;
> h) DEF definition of the preferred term (see 6.2.3);
> i) HN history note (see 6.2.4);
> j) SC subject category or other concept group to which the concept has been assigned.
> In this arrangement, the SN and UF are given near the beginning of the list because they clarify the scope of the concept. They are followed by relationships with other terms. DEF and HN are given last because they are viewed as administrative fields, used more by the thesaurus editor than the indexer or searcher. As an option, these last fields may be shown after the scope note, bringing all text fields together. However, it is important to ensure that SN and DEF are not confused when both are present.

It is not complicated to reorder the properties of the concept by editing line 62 in Concept.php :

> $this->order = array("rdf:type", "dc:isReplacedBy", "skos:definition", "skos:broader", "skos:narrower", "skos:related", "skos:altLabel", "skosmos:memberOf", "skos:note", "skos:scopeNote", "skos:historyNote", "rdfs:comment", "dc11:source", "dc:source", "skos:prefLabel");

However this doesn't allow to control the order of the synonyms and the group information. Having the possibility to easily configure this - without tempering with the HTML template - would be nice. A configuration option, maybe, like "UseIso25964PropertyOrder", or something similar ?

Cheers

osma:
This draft PR implements the first steps of making the property order configurable on concept pages, defined in #554.

Three sorts of property order can be selected using the new, vocabulary-specific configuration setting `skosmos:propertyOrder`:

1. Default/traditional property order, used in previous Skosmos releases
2. ISO 25964 property order
3. Custom property order

There are still several unresolved problems with this PR:

- [x] one unit test is failing, apparently because of a concept that doesn't know its vocabulary
- [x] the property order only affects the "core" properties in the middle of the concept page; it is not possible to adjust the order of "belongs to group", "in other languages" and "URI" properties near the bottom since their position is currently hardwired in the Twig templates
- [x] ISO 25964 only defines the order for some concept properties, not all of the common ones (e.g. `dc:source`); it's unclear what to do with these (try to find a sensible slot? leave them all at the bottom?)
- [x] the current code probably doesn't work too well when a custom order is given with URIs from namespaces not known to EasyRdf (e.g. RDA properties)

Opening a draft PR to get early feedback from QA tools.

Fixes #554

### DIFF
--- a/model/VocabularyConfig.php
+++ b/model/VocabularyConfig.php
@@ -9,6 +9,14 @@ class VocabularyConfig extends BaseConfig
     private $pluginParameters;
     private $languageOrderCache = array();
 
+    private $DEFAULT_PROPERTY_ORDER = array("rdf:type", "dc:isReplacedBy",
+    "skos:definition", "skos:broader", "isothes:broaderGeneric",
+    "isothes:broaderPartitive", "isothes:broaderInstantial",
+    "skos:narrower", "isothes:narrowerGeneric", "isothes:narrowerPartitive",
+    "isothes:narrowerInstantial", "skos:related", "skos:altLabel",
+    "skosmos:memberOf", "skos:note", "skos:scopeNote", "skos:historyNote",
+    "rdfs:comment", "dc11:source", "dc:source", "skos:prefLabel");
+
     public function __construct($resource, $globalPlugins=array())
     {
         $this->resource = $resource;
@@ -533,5 +541,14 @@ class VocabularyConfig extends BaseConfig
     public function isUseModifiedDate()
     {
         return $this->getBoolean('skosmos:useModifiedDate', false);
+    }
+
+    /**
+     * @return array
+     */
+    public function getPropertyOrder()
+    {
+        //return $this->getResources('skosmos:propertyOrder');
+        return $this->DEFAULT_PROPERTY_ORDER;
     }
 }
--- a/model/VocabularyConfig.php
+++ b/model/VocabularyConfig.php
@@ -17,6 +17,13 @@ class VocabularyConfig extends BaseConfig
     "skosmos:memberOf", "skos:note", "skos:scopeNote", "skos:historyNote",
     "rdfs:comment", "dc11:source", "dc:source", "skos:prefLabel");
 
+    private $ISO25964_PROPERTY_ORDER = array("rdf:type", "skos:scopeNote",
+    "skos:altLabel", "skos:broader", "isothes:broaderGeneric",
+    "isothes:broaderPartitive", "isothes:broaderInstantial",
+    "skos:narrower", "isothes:narrowerGeneric", "isothes:narrowerPartitive",
+    "isothes:narrowerInstantial", "skos:related", "skos:definition",
+    "skos:historyNote", "skosmos:memberOf");
+
     public function __construct($resource, $globalPlugins=array())
     {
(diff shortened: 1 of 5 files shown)


osma:
The order setting now affects also the group and array properties after the changes in commit 
ca6ed67 . The location of these properties within the page can be specified using the property URIs `skosmos:memberOf` and `skosmos:memberOfArray`, respectively. By default, and in the ISO order, they are still at the bottom.

model/VocabularyConfig.php
```diff
@@ -9,6 +9,25 @@ class VocabularyConfig extends BaseConfig
...
 
+    const DEFAULT_PROPERTY_ORDER = array("rdf:type", "dc:isReplacedBy",
+    "skos:definition", "skos:broader", "isothes:broaderGeneric",
+    "isothes:broaderPartitive", "isothes:broaderInstantial",
+    "skos:narrower", "isothes:narrowerGeneric", "isothes:narrowerPartitive",
+    "isothes:narrowerInstantial", "skos:related", "skos:altLabel",
+    "skos:note", "skos:scopeNote", "skos:historyNote", "rdfs:comment",
+    "dc11:source", "dc:source", "skosmos:memberOf", "skosmos:memberOfArray");
+
+    const ISO25964_PROPERTY_ORDER = array("rdf:type", "dc:isReplacedBy",
```
joelit, line 20: It would be a good idea to mention this reasoning (like grouping skos:definition together with the other text fields) in the Skosmos wiki, too.

osma:
> It would be a good idea to explain the general idea of the choice of the ordering (currently only the ISO 25964-1 and the default Skosmos order, but there might come more) in the Skosmos wiki to help Skosmos users choose a suitable ordering for themselves.

I agree, will document this when explaining the property

osma:
Hmm, meanwhile some conflicts have emerged, need to fix those before merging this

view/concept-shared.twig
```diff
@@ -111,42 +111,7 @@
...
-              {% endif %}
-            {% endfor %}
-          {% endfor %}
-          </ul>
-        </div></div>
-      </div>
-      {% endif %}
-      {% set foreignLabels = concept.foreignLabels %}
-      {% if foreignLabels %}
+      {% if concept.foreignLabels %}
```
kouralex, line 114: Here one has made a bit too eager conflict resolve as twig variable `foreignLabels` is used later on in the template. I will fix this in a separate commit, pushed directly to master.

osma: Thanks @kouralex, good catch! I did the merge in GitHub UI as I thought it was trivial, but I didn't pay enough attention to this change.

PR #993, opened 2020-06-11
Issue835 altlabel sort

### LINKED ISSUE #835: AltLabels in other languages are not sorted alphabetically
http://dev.finto.fi/ykl/fi/page/71.3

It's easy to see in a vocabulary with lots of altLabels per concept. Here the Swedish altLabels happen to be the exact same ones as in Finnish - the labels for the selected language are sorted, whereas the labels for the other languages are not.

kouralex:
Fixes #835 via introducing a language-aware collation of foreign prefLabels/altLabels.

### DIFF
--- a/model/Concept.php
+++ b/model/Concept.php
@@ -859,11 +859,11 @@ class Concept extends VocabularyDataObject implements Modifiable
      */
     public function getForeignLabels()
     {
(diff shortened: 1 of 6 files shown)


kouralex:
Good catch @joelit! Yes, I will update the test vocab accordingly.

Issue #974, opened 2020-04-23
The layout of the concept page could be cleaner


MikkoAleksanteri:
## At which URL did you encounter the problem?
http://finto.fi/yso/fi/page/p5808 
using Firefox

At least in finto.fi, the information in the concept page has been piling up over the years and the current layout of the concept page as a whole is quite messy (in YSO).

It could be a good idea to go through the elements currently in the concept page, check the borders, empty spaces, font sizes etc. and try to get it more balanced.

One place to start would be to line up the left edges of 
- the name of the language of alt labels
- the information about the creation and modification data of the concept
- the name of a vocaburay of a linked concept

See picture below.
![Finto issue concept page](https://user-images.githubusercontent.com/8804680/80101149-b3a9ad00-8579-11ea-9345-227de31642ce.jpg)

joelit:
The rightmost column of the mapping properties - section at least seems to be in a different font. Some kind of a spring cleaning seems to be in order.

osma:
The concept page layout has been redesigned for Skosmos 3.

Closing this as stale. Please reopen if still relevant.

PR #994, opened 2020-06-11
Fix the font definitions for timestamps and linked vocabulary names

osma:
This minor CSS problem (not using the Fira font everywher) was mentioned in #974.
It's a trivial CSS fix. An extra semicolon is also removed.


osma:
Sorry @kinow, I didn't intend to tag you when picking a reviewer, I clicked in the wrong location.

PR #995, opened 2020-06-12
Show tooltips (rdfs:comment or skos:definition) for custom properties

### LINKED ISSUE #824: Support tooltips for custom properties
The tooltips is only available for the properties recognized by Skosmos (i.e. definition, broader concept,...). 
It would be nice if it is supported for custom properties such as: Is Hosted By, See Also,... in the picture below.

image

osma:
The tooltip content is read from `rdfs:comment` or `skos:definition` values of the property URI.

Example of Finnish Corporate Names showing RDA property definition as a tooltip:

kuva

Fixes #824

### DIFF
--- a/model/ConceptProperty.php
+++ b/model/ConceptProperty.php
@@ -11,6 +11,8 @@ class ConceptProperty
     private $super;
     /** stores the property label */
     private $label;
+    /** stores the property tooltip */
+    private $tooltip;
     /** stores the property values */
     private $values;
     /** flag whether the values are sorted, as we do lazy sorting */
@@ -22,11 +24,11 @@ class ConceptProperty
      * @param string $prop property type eg. 'rdf:type'.
      * @param string $label
      */
-    public function __construct($prop, $label, $help=null, $super=null, $sort_by_notation=false)
+    public function __construct($prop, $label, $tooltip=null, $super=null, $sort_by_notation=false)
     {
         $this->prop = $prop;
         $this->label = $label;
-        $this->help = $help;
+        $this->tooltip = $tooltip;
         $this->values = array();
         $this->is_sorted = true;
         $this->super = $super;
@@ -71,11 +73,11 @@ class ConceptProperty
         }
 
        // if not, see if there was a comment/definition for the property in the graph
-        if ($this->help !== null) {
-            return $this->help;
+        if ($this->tooltip !== null) {
+            return $this->tooltip;
         }
 
-        // when nothing is found, don't show the help text at all
+        // when nothing is found, don't show the tooltip at all
         return null;
     }
 
--- a/model/ConceptProperty.php
+++ b/model/ConceptProperty.php
@@ -22,7 +22,10 @@ class ConceptProperty
     /**
      * Label parameter seems to be optional in this phase.
      * @param string $prop property type eg. 'rdf:type'.
-     * @param string $label
+     * @param string $label property label
+     * @param string $tooltip property tooltip/description
+     * @param string $super URI of superproperty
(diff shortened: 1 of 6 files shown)


model/ConceptProperty.php
```diff
@@ -22,10 +24,11 @@ class ConceptProperty
      * @param string $prop property type eg. 'rdf:type'.
      * @param string $label
      */
-    public function __construct($prop, $label, $super=null, $sort_by_notation=false)
+    public function __construct($prop, $label, $tooltip=null, $super=null, $sort_by_notation=false)
```
kouralex, line 27: One could document any new parameter(s) here.
On a side note, it's also a bit risky to add a parameter in the middle of a constructor call parameter list, but in this case it does not cause trouble. It may very well be intended, though, if one thinks it is a more important one to have.

osma: You're right, the phpdoc declaration for the new parameter is missing.

I added the parameter in the middle because in PHP, only the last parameters can be omitted. Since there is a need to call the constructor both with and without the last two parameters, this was the only clean option. I did check all calls to the constructor.

model/ConceptProperty.php
```diff
@@ -47,7 +50,7 @@ public function getLabel()
         }
 
         // if not, see if there was a label for the property in the graph
-        if ($this->label) {
+        if ($this->label !== null) {
```
kouralex, line 53: Just a side note: I can see why this was done, but does not - nevertheless - change the code behavior a bit. But one can argument for code quality measures, and I agree.

model/ConceptProperty.php
```diff
@@ -56,14 +59,26 @@ public function getLabel()
...
+     * Returns text for the property tooltip.
      * @return string
      */
     public function getDescription()
     {
         $helpprop = $this->prop . "_help";
 
-        return gettext($helpprop); // can't use string constant, it'd be picked up by xgettext
+        // see if we have a translation with the help text
+        $help = gettext($helpprop);
```
kouralex, line 70: This will prefer Skosmos translations, but that will must likely be OK.

view/concept-shared.twig
```diff
@@ -62,7 +62,7 @@
         {% if property.getSubPropertyOf != 'skos:hiddenLabel' %}
         <div class="row{% if property.type == 'dc:isReplacedBy' %} replaced-by{% endif%}">
           <div class="property-label">
-            <span class="versal{% if property.type == 'rdf:type' %}-bold{% endif %}{% if not (property.type in property.description and '_help' in property.description) %} property-click" title="{{ property.description }}"{% else %}"{% endif %}>{{ property.label|upper }}</span>
+            <span class="versal{% if property.type == 'rdf:type' %}-bold{% endif %}{% if property.description %} property-click" title="{{ property.description }}"{% else %}"{% endif %}>{{ property.label|upper }}</span>
```
kouralex, line 65: This could only have an `if` part.

model/ConceptProperty.php
osma: I copied and pasted some of the code for the `label` method into `description` and noticed this potential problem, so I fixed it in both places. In PHP, a string containing a zero `"0"` evaluates as false. It's an unlikely label, I know, but still, the behavior does change a little bit here ;)

model/ConceptProperty.php
osma: Yes, that's the intent (same as for property labels)

view/concept-shared.twig
osma: Good catch!

PR #1013, opened 2020-06-23
Issue724 group tab shorten uris

### LINKED ISSUE #724: Groups tab shows concepts from the same vocab to have URIs like they are from other vocabularies
http://finto.fi/juho/fi/page/p771
1. Go to the link provided above (using the Finnish UI)
2. Select "alueellinen yhteistyö" from the Groups tab

I'd expect the hover URI and address bar URI to be in the form consistent with the vocabulary. Instead, it is now referred to as if it was from another vocab.

Example:
http://finto.fi/juho/fi/page/?uri=http%3A%2F%2Fwww.yso.fi%2Fonto%2Fjuho%2Fp6048
instead of
http://www.yso.fi/onto/juho/p6048
Firefox

kouralex:
Fixes #724 .

Additionally addresses the following issues:
1) Adds support for characters `-_~` in shortened URIs (see https://en.wikipedia.org/wiki/Percent-encoding#Types_of_URI_characters for more information)
2) Standardizes behavior between `groups` and `hierarchy` tabs (use same code).
3) Removes a previously unnoticed error that happened to set the parameters in some cases as `?uri=xxx?clang=yy` instead of the correct `?uri=xxx&clang=yy` way.

### DIFF
--- a/resource/js/docready.js
+++ b/resource/js/docready.js
@@ -284,19 +284,16 @@ $(function() { // DOCUMENT READY
       function(event) {
         event.preventDefault();
         var targetUrl = event.target.href;
-        var parameters = (clang !== lang) ? $.param({'clang' : clang}) : $.param({});
-        var historyUrl = (clang !== lang) ? targetUrl + '?' + parameters : targetUrl;
         $('#hier-trigger').attr('href', targetUrl);
         var $content = $('.content').empty().append($delayedSpinner.hide());
         var loading = delaySpinner();
         $.ajax({
             url : targetUrl,
-            data: parameters,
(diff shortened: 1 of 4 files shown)


kouralex:
Side note: support for 1) was needed in order to address issues in, for example, AGROVOC, see http://skosmos.dev.finto.fi/agrovoc/fi/page/?uri=http%3A%2F%2Faims.fao.org%2Faos%2Fagrovoc%2Fc_36791

joelit (approved):
Tested to work. Good job cleaning up the js code!

PR #1024, opened 2020-07-02
fix #1018 by explicitly declaring the UI locale in getDate()

### LINKED ISSUE #1018: Timestamp date format should be based on UI language, not content language
http://finto.fi/yso/fi/page/p23406?clang=en

1. Go to a concept page that has created and/or modified timestamps
2. Use different content and UI languages (in the above case, Finnish UI and English content)

Expected to see timestamps (both the label and the date) formatted in the UI language, e.g.
> Luotu 9.11.2009, viimeksi muokattu 25.1.2020

Instead the timestamp labels are in the UI language, but the date format is selected based on content language:
> Luotu 11/9/09, viimeksi muokattu 1/25/20

(this "American style" date format is not used in Finnish text)

kouralex:
This PR fixes #1018 by declaring the UI locale in the getDate() method. This could, however, also be fixed by changing the signature of the method and passing the UI language parameter from the Twig template and, thus, making it more versatile, but because the method is not in use in anywhere else it can be fixed straightaway in code.

### DIFF
--- a/model/Concept.php
+++ b/model/Concept.php
@@ -85,11 +85,6 @@ class Concept extends VocabularyDataObject implements Modifiable
         }
         $this->graph = $graph;
         $this->clang = $clang;
-        // setting the Punic plugins locale for localized datetime conversions
-        if ($this->clang && $this->clang !== '') {
-            Punic\Data::setDefaultLocale($clang);
-        }
-
     }
 
     /**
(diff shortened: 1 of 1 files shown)


osma (approved):
This simpler fix looks good. I've verified that it works locally. +1 for merging.

osma:
I decided not to wait for @kouralex and just went ahead and merged this. Thanks!

PR #1026, opened 2020-07-09
Improve ajax behavior; miscellanneous fixes

kouralex:
This PR fixes/implements the following:

- implement aborting of unnecessary requests (via implementing different types of queues for content and sidebar)
- fix issue with plugin-induced content appended whilst ajaxing
- fix throbber icon not showing on hierarchy whilst ajaxing

However, there remains more to be done, e.g., enabling the usage of ajax requests in cases such as clicking a concept on content page, clicking vocabulary title, moving back/forth search and content pages, clicking hierarchy causes an extra page load in some circumstances, etc. Similarly, much of the code could and should be refactored - no need for duplicate code.


joelit (commented):
The ajax queue is a bit difficult to review and test, but as far as I can see this is a improvement when changing between concepts while the ajax is still loading.

PR #1028, opened 2020-07-10
Fixes #1027 by commenting out css file path

### LINKED ISSUE #1027: Config.ttl.dist pointing to a non-existing css file path by default
This results to 404 error with fresh installations. The file path should be commented out from:
https://github.com/NatLibFi/Skosmos/blob/8e097063f34e56e59d128f93733bd96cad77d096/config.ttl.dist#L55

Vainonen:
Comments out config.ttl.dist Skosmos custom CSS file path, which does not exist by default.


kouralex:
Nice catch and good PR! I will gladly merge this.

PR #1030, opened 2020-07-10
fixes #1029 by adding a restriction on deprecated subjects of a concept

### LINKED ISSUE #1029: Downloading a concept: deprecated subjects of a concept should not be included in end result
https://finto.fi/yso/fi/page/p3064
1. Go to URL above
2. Download as Turtle

I see a line similar to this:
`yso:p29911 skos:relatedMatch yso:p3064 .`
I would not want that to be included as yso:p29911 is deprecated.

kouralex:
Fixes #1029


joelit (approved):
Works fine!

PR #1046, opened 2020-08-31
Add HTML classes for properties so they can be targeted in JS and CSS

### LINKED ISSUE #1034: Adding HTML IDs / classes to the property value list
Enhancement proposal: 
Add HTML IDs / classes to the property list elements so that they can be addressed from Skosmos plugins. Currently, 

Example: a plugin needs to address the value of the `dc:identifier` property (in order to add a link next to the identifier value). It would be easier if relevant property values had an HTML class attribute (e.g. `class="dc_identifier"`) that allows plugin developers to directly address these HTML elements.

`<span class="dc_identifier">LNC10-000194258</span>`

osma:
Fixes #1034

All the `div` blocks which represent properties will get a HTML class `property`.

Additionally, those `div`s will get a specific class identifying the property in question. Usually, the class identifier is generated from the property URI:

1. For known property URIs that can be shortened, the class will be e.g. `prop-dc_identifier` or `prop-skos_broader`
2. For unknown URI namespaces that cannot be shortened, the class will be derived from the full URI, e.g. `prop-http___rdaregistry_info_Elements_a_P50008` (for [rdaa:P50008](http://rdaregistry.info/Elements/a/P50008) - but the `rdaa` prefix is not known by Skosmos)

There are a few special cases for some `div` blocks representing special purpose properties:

* preferred label property: `prop-preflabel`
* other languages property: `prop-other-languages`
* URI "property": `prop-uri`
* mapping properties: `prop-mapping`

Any comments @CaptSolo @kinow?

### DIFF
--- a/model/ConceptProperty.php
+++ b/model/ConceptProperty.php
@@ -62,6 +62,14 @@ class ConceptProperty
     }
 
     /**
+     * Returns an alphanumeric ID for the property, suitable for use as a CSS identifier.
+     */
+    public function getID()
+    {
+        return preg_replace('/[^A-Za-z0-9-]/', '_', $this->prop);
+    }
+
+    /**
      * Returns text for the property tooltip.
      * @return string
      */
--- a/resource/js/scripts.js
+++ b/resource/js/scripts.js
@@ -433,6 +433,7 @@ function loadMappingProperties(concept, lang, contentLang, $htmlElement, concept
         if (!found) {
           conceptProperty = {
             'type': conceptMappingPropertyValue.type[0],
+            'id': conceptMappingPropertyValue.type[0].replace(/[^A-Za-z-]/g, '_'),
             'label': conceptMappingPropertyValue.typeLabel,
             'notation': conceptMappingPropertyValue.notation,
             'description': conceptMappingPropertyValue.description,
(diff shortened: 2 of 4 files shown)


kinow (approved):
Looks good to me! Great improvement @osma!

kouralex (approved):
Very needed improvement indeed! Thanks @osma for this PR.

I just want to explicitly state that 

1) The simplistic way of just adding the `prop-{{property.ID}}` to the class name will only show the most accurate property ID due to the de-duplication procedure, e.g., on https://finto.fi/yso-paikat/fi/page/?uri=http://www.yso.fi/onto/yso/p507412 you will only have `isothes_broaderPartitive` and no `skos_broader`. Some implementation could, in theory, only search for the known skos:broader property but maybe that is a minor detail and may be shrugged off (via YAGNI principle).

2) The names used for prefixes in `config.ttl` will end up into the class names, which may cause unexpected values to be used (from the perspective of a third party). This cannot be countered as there is no prefix mappings given.

But it is good as it now, I will happily approve!

osma:
Thanks @kouralex - I'm aware of those edge cases but I think it's worth merging this as-is and then improve later on if it turns out to be necessary.

CaptSolo:
Thank you - works like a charm! :-)

PR #1050, opened 2020-09-01
Issue1048 skip to content

### LINKED ISSUE #1048: Skip to content functionality
When browsing the Skosmos pages with either vocabulary list, vocabulary info, search results or concept info, the browser focus should go first to the skip to content -link in:
https://github.com/NatLibFi/Skosmos/blob/d633b60c95b39cf9c7fe261fca03097b3e057d1e/view/light.twig#L23

This does not happen in practise, as the focus is moved to the search field first. The culprit seems to be docready.js, which alters the focus in various ways, e.g.

https://github.com/NatLibFi/Skosmos/blob/d633b60c95b39cf9c7fe261fca03097b3e057d1e/resource/js/docready.js#L1109

This issue was split from #1036 #1037 #1038 & #1039

joelit:
Tested with Firefox and Chromium. I will test screen-reader-only captions for the side bar and the main content before this is considered to be done. Fixes #1048


view/light.twig
```diff
@@ -134,7 +135,7 @@
       {% include 'left.inc' ignore missing %}
     {% endif %}
     {% block error %}
-    <div id="maincontent">
```
kouralex, line 137: Dropping the id `maincontent` will affect the contents of this element. https://github.com/NatLibFi/Skosmos/blob/5fb7bf792be2ab09b6f9928f904f1517bcfb5efe/resource/css/styles.css#L606-L608 If these lines are no longer needed or they did not do anything, I suggest removing them altogether, i.e., remove all references to `#maincontent`.

joelit: True. I didn't see it affecting a thing but in theory, it would be there to make sure a link or subheader does not change the line heights. I'll actually just keep the id on the div in case someone using Skosmos with custom CSS would refer to it.

That being said, now we'd have main element #main, div element #maincontent, and div element #main-container.

osma:
Tested the "skip to content" functionality in this PR on a text-based browser (lynx).

* Front page: works as expected, takes me to the beginning of the vocabulary list
* Vocabulary page: not what I expected - takes me to the A-Z index in the sidebar, not the main content (which would be the information about the vocabulary)
* Search result page: not what I expected - takes me to the search limit form, not the search result list
* Concept page: not what I expected - takes me to the A-Z index in the sidebar, not the main content (which would be the information about the concept)

I think the `main` element should be moved deeper/later into the DOM tree on all page types except the front page - after the sidebar (which is a navigational element, not part of main content). Otherwise the user will not land in the right place when selecting this link.

joelit:
Good points! If I didn't understand it wrong, there was a request that the sidebar is added to the elements where thew focus should go to. This is why I asked about it. With the answer I was given, I changed the main element upwards.

If the flow order between the main content and the sidebar is adjusted, it should bring out the desired functionality, right?


But is it is too confusing, I am more than happy to go against the specs to get a sensible outcome, as I found the spec to be confusing and loosely defined.

osma:
> If the flow order between the main content and the sidebar is adjusted, it should bring out the desired functionality, right?

You mean moving the sidebar after the main content in the HTML DOM tree? Yes, I believe that would fix the issue, as then the link would take the user directly to the beginning of the main content, as expected.

joelit:
We discussed this in detail today, and I'll just skip the possibility to get to the sidebar after navigating to the main content. I will just revert this back to focus on the main content instead.

view/light.twig
```diff
@@ -148,7 +148,7 @@
       {% include 'right.inc' ignore missing %}
     {% endif %}
     <footer id="footer"></footer>
-  </div>
+  </main>
```
osma, line 151: This is the closing `div` tag for `div class="main-container"` and should not be changed.

Instead, the closing `div` 6 lines earlier (line 145, just before `{% endblock %}`) should be changed to `</main>`

osma (changes requested):
I re-tested the skip links using Lynx. Now they all go to the right place. Excellent!

However, I noticed that the HTML tree structure is now broken, I think you changed the wrong `</div>` to `</main>`, see comment below.

osma (approved):
All good now, OK to merge!

view/light.twig
joelit: Damnit, I cannot believe I missed that. Fixed, sorry about it.

PR #1051, opened 2020-09-02
remove dead code: the renderPropertyMappingValues function is never used

osma:
While working on #1046 I noticed that the `renderPropertyMappingValues` function is never called. This PR removes it.


kinow (approved):
Did a quick search, used a , then case insensitive, nothing. Looks good!

PR #1055, opened 2020-09-04
Issue1052 subheadings with text

joelit:
The new layout to get rid of tables when displaying subheadings. The new heading styles should use this layout as its basis.


kouralex (changes requested):
Whilst this PR gets rid of the table structure found on both ontology page and concept page, I encountered a few issues before this can be merged:

1) The content language selection is on the left side now. See attached image below where image on the left hand side has been rendered using the code from this PR branch and the right hand side using the current master.
image
2. As one can see from the above image, this PR creates more space between the separators. Maybe a little bit less would suffice?
3. The fonts on the ontology page for some  are not alright (font + size, maybe other properties too). See below image.
image
4. I find it a little bit hard to follow the CSS declarations but e.g., setting negative values for the `bottom` property is something I would suggest to avoid as these things should now be easier to adjust with flex properties.

Otherwise, this PR seems to get rid of some tables which is good progress!

joelit:
> Whilst this PR gets rid of the table structure found on both ontology page and concept page, I encountered a few issues before this can be merged:
> 
> 1. The content language selection is on the left side now. See attached image below where image on the left hand side has been rendered using the code from this PR branch and the right hand side using the current master.
>    image

There was a conflict between bootstrap and the Skosmos css. I resolved it for now, though I suspect more effort should be made ensuring the class names won't conflict in the future.


> 1. As one can see from the above image, this PR creates more space between the separators. Maybe a little bit less would suffice?

This was the closest replication of the previous layout. There is an extra pixel row underneath the row separator bar, but it looks good to me.


> 2. The fonts on the ontology page for some  are not alright (font + size, maybe other properties too). See below image.
>    image

Fixed.


> 3. I find it a little bit hard to follow the CSS declarations but e.g., setting negative values for the `bottom` property is something I would suggest to avoid as these things should now be easier to adjust with flex properties.

It's true. The CSS declarations are hard to follow. Restructuring would be in order.

PR #1056, opened 2020-09-04
Accessibility improvements for keyboard / tab key navigation

### LINKED ISSUE #1049: Accessibility of tab key / keyboard navigation
Tab order is one of the accessibility improvements requested in issues #1036, #1037, #1038, #1039.

The tab order should allow visiting every interactive element on the page in a logical order. When the order differs from the order the elements appear in HTML, the `tabindex` attribute can be used to set an explicit order.

The autofocus functionality on the search box should probably be disabled (see #1048) before the tab order can be meaningfully adjusted.

osma:
This PR contains accessibility improvements that aim to enable keyboard navigation, especially tabbing through links.

1. The **Skip to main content** link has been enhanced so that it now becomes visible when the user presses the Tab key. The layout isn't ideal as it partially obscures the logo in the corner - ideas for improvement welcome!
2. The **Help** link in the top bar can now be activated with the keyboard just like it were a normal link.
3. Previously outlines for focused links were completely disabled. They are now enabled, with a crude but functional black outline displayed for the currently active link.

Still TODO:

- when tabbing to the search form in the top bar, the current element isn't highlighted in any way

Visible "Skip to content" link:

image

Tabbing to the Help "link":

image

All links are now outlined when active:

image

(screenshots taken using Chromium; the look is similar in Firefox except outline borders are not rounded)

Fixes #1049

### DIFF
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -1424,8 +1424,8 @@ ul.nav-tabs > li {
 }
 
 #skiptocontent:focus {
-  top: auto;
-  left: 1.5em;
+  top: 0.5em;
+  left: 0.5em;
   height: auto;
   width: auto;
   background-color: white;
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -254,8 +254,9 @@ ul {
   border-top-color: #ffffff;
 }
 
-#lang-dropdown-toggle:focus, .multiselect:focus {
+#lang-dropdown-toggle:focus, .multiselect:focus, #search-all-button:focus {
   outline: 3px solid white;
+  background-color: #00748f;
 }
 
 .navbar-form .dropdown-menu {
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -342,7 +342,6 @@ ul.dropdown-menu > li:last-child > input {
 
 .header-left > h1 > a {
   color: #ffffff;
-  display: block;
   height: 50px;
 }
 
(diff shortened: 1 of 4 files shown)


osma:
>     when tabbing to the search form in the top bar, the current element isn't highlighted in any way

This has now been addressed. Active search form elements are highlighted with a thick white outline plus a lighter background color than usual:

image

joelit:
I really like how it looks. There are some minor things I could find:

* The highlight box is not not always fully visible on all sides
  * On the link to the vocabulary list (in the nav bar at the top)
  * On the first three tabs of the left-side bar in vocabulary and concept pages
  *  The fourth tab is positioned under the scroll bar on the right, if the alphabetical tab is not avtive
I would try to tackle this by giving the focused element a higher z-index, but I'm not sure if it is a good idea. I wouldn't sweat about those too much, since the second point could change this.

*  I couldn't get the highlight boxes working on Firefox at all, which probably should be addressed somehow. (Tested on Chrome, Chromium and Firefox.)

That being said, I can really see myself using the tab navigation instead of mouse.

joelit (changes requested):
I suppose leaving this as it is, but submitting a new issue about this feature on Firefox is also an option.

osma:
Works fine on Firefox for me... Are you sure you did a full reload?

osma:
> The highlight box is not not always fully visible on all sides

You're right, it is sometimes obscured from some sides.

Issue #1058, opened 2020-09-04
Non-standard content language dropdown may cause accessibility problems


osma:
## At which URL did you encounter the problem?
http://localhost/Skosmos/en/

The content language dropdown is implemented as a `button` element and an unordered list (`ul`) containing all the language values, plus some JS+CSS code to tie them together so it looks like a menu. This is a non-standard implementation of a dropdown menu; a plain `select` form element would be better, though probably more difficult to style.

When checking for accessibility issues using the WebAIM checker (#1040), it complains about orphaned `label` elements next to the language dropdown. But the underlying issue is the implementation of the language dropdown, not the `label` element next to it.

osma:
This was a problem for Skosmos 2.

This is done in a different, hopefully better way in Skosmos 3. Closing this issue.

PR #1060, opened 2020-09-08
Topbar wcag focus improvements

kouralex:
This PR fixes some issues concerning the new focusing outline that were introduced by #1056 (namely, introducing a positive z-index to all focused elements).

Extra work was done in order to make topbar elements (excluding the #1058-related element) behave as one would expect. However, there is one old quirk still left: sizing the window to some 700px will push the `Skosmos` out of the main column and on the right side. I adjusted the ontology page so that it behaves the same now both for ontology and front page (previously it just disappeared completely (display: none) for ontology page, making it completely unreachable for any keyboard/screen reader user).

Tested with Firefox.

### DIFF
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -103,6 +103,8 @@ a:hover {
 
 #service-logo {
   max-width: 450px;
+  display: inline-block;
+  margin: 4px;
 }
 
 /* basics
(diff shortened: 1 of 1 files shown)


kouralex:
I will merge this now in order to be properly able to assist the current state of affairs.

PR #1081, opened 2020-10-18
translations for w3c geo latitude and longitude, expose vocabulary short name

kouralex:


kouralex:
This PR looks a bit odd after rebasing (I thought I did it right but it seems like there was an existing (old) PR (which is this)) but the real comparison can be found here: https://github.com/NatLibFi/Skosmos/compare/geographical-translations.

Reverted a change that made exposed vocabulary short name content language-aware (not shown in commit stack). Also switched from geo prefix into wgs84 which should be 100% explicit.

joelit (approved):

PR #1084, opened 2020-10-20
Show short name of vocabulary for property values from another vocab

osma:
This one-line (plus test) PR will change the vocabulary names that are shown on the concept page next to property values that come from another vocabulary known to Skosmos. This change was requested by the Kanto/finaf project. The long vocabulary names clutter the display of property values. Using the short names instead makes the display a bit cleaner.

Before this PR:

image

After:

image

### DIFF
--- a/model/ConceptPropertyValue.php
+++ b/model/ConceptPropertyValue.php
@@ -115,7 +115,7 @@ class ConceptPropertyValue extends VocabularyDataObject
 
     public function getVocabName()
     {
-        return $this->vocab->getTitle();
+        return $this->vocab->getShortName();
     }
 
     public function addSubMember($member, $lang = '')
--- a/tests/ConceptPropertyValueTest.php
+++ b/tests/ConceptPropertyValueTest.php
@@ -94,7 +94,7 @@ class ConceptPropertyValueTest extends PHPUnit\Framework\TestCase
   public function testGetVocabName() {
     $props = $this->concept->getProperties();
     $propvals = $props['skos:narrower']->getValues();
-    $this->assertEquals('Test ontology', $propvals['Crucian carp http://www.skosmos.skos/test/ta121']->getVocabName());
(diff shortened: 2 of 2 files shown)


kouralex:
I wonder if this could be implemented as an option, i.e., as a Skosmos property?

Nice work anyway! :)

osma:
@kouralex 
> I wonder if this could be implemented as an option, i.e., as a Skosmos property?

If there's a genuine need for both long and short names, then making it configurable would make sense. I'm a bit skeptical. I'd prefer just to merge this change as-is, since I think that short vocabulary names are better in any case. They are already used in other contexts in the UI, e.g. search results.

kouralex:
> 
> 
> @kouralex
> 
> > I wonder if this could be implemented as an option, i.e., as a Skosmos property?
> 
> If there's a genuine need for both long and short names, then making it configurable would make sense. I'm a bit skeptical. I'd prefer just to merge this change as-is, since I think that short vocabulary names are better in any case. They are already used in other contexts in the UI, e.g. search results.

I see your point. The advantage of your solution is that it is simpler, however, for example `YSO` does not yell a thing to a Swedish person (as it is ALLFO in their point of view). That said, I am willing to accept your solution, just wanted to point out this case.

osma:
> I see your point. The advantage of your solution is that it is simpler, however, for example YSO does not yell a thing to a Swedish person (as it is ALLFO in their point of view). That said, I am willing to accept your solution, just wanted to point out this case.

Short names are (or can be) language specific too, just like long names of vocabularies.

kouralex:
@osma 
> Short names are (or can be) language specific too, just like long names of vocabularies.

I stand corrected (it seems like this has been like that, as your said, since v0.6).

I suppose there is no problem then and you have pointed out your preference on this matter. Go on and merge! :)

PR #1094, opened 2020-11-19
introduce right-paddings for vocab uri & concept info labels; fixes #…

kouralex:
…1080

Backport needed?


joelit (approved):
Works on Chrome and Firefox, though I couldn't reproduce the issue on Chrome to start with.

PR #1095, opened 2020-11-26
Fix typo in swagger.json offset description

ammsalme:


kouralex:
Thank you for noticing!

PR #1097, opened 2020-11-30
add treatment for JSON-LD vocabulary data dump; fixes #1096

### LINKED ISSUE #1096: Add link to JSON-LD data dump
At the moment, Skosmos does not list JSON-LD version of the vocabulary (its data dump) on the vocabulary page, even though it may be declared and downloaded via the REST API.

The culprit seems to be here (see the code below); there is no treatment for JSON-LD. https://github.com/NatLibFi/Skosmos/blob/91a30e5029a5e8a0cbb6c349fed7890b58a61334/view/vocab-shared.twig#L58-L76

As `application/ld+json` is already in supported formats, no other changes are required. EasyRdf interprets `.jsonld` files as such so no `dc:format` declaration is required (just tested).

kouralex:
This PR fixes #1096 via adding treatment for JSON-LD data dump.

After mergeing it might be a good idea to update https://github.com/NatLibFi/Skosmos/wiki/ServingLinkedData wiki page accordingly.

### DIFF
--- a/view/vocab-shared.twig
+++ b/view/vocab-shared.twig
@@ -63,6 +63,9 @@
         {% if 'text/turtle' in vocab.config.dataURLs|keys %}
           <a href="rest/v1/{{ request.vocabid }}/data?format=text/turtle">TURTLE</a>
         {% endif %}
+        {% if 'application/ld+json' in vocab.config.dataURLs|keys %}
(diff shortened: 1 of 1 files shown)


osma:
Tested locally, seems to work fine, merging away!

Issue #751, opened 2018-03-28
Alphabet Letter not displayed if only present in altLabel


Kordishal:
## At which URL did you encounter the problem?
http://bartoc-skosmos.unibas.ch/fast-chrono/en/index

## What steps will reproduce the problem?
1. Search for "Fra" and select the first suggestion (Francis I, Reign of (Austria))
2. After the concept page is loaded click on the Alphabetical list.
3. Displayed will be a list of altLabels which begin with letter A (instead of F or the numbers).
4. Selected can only be S (Since 1900 etc.) T (To 250 B.C. etc) and numbers (all other prefLabels.

## What is the expected output? What do you see instead?
The Alphabetical Listing is generated from the prefLabels. In this case the all of the prefLabels are numbers, while the altLabels are Names of the reigns/eras/etc. with this date range.

As a result the Numbers are properly listed. You can properly search for altLabels. But there is no display in the Alphabetical List since only 2 letters are included. This makes it hard to navigate this kind of vocabulary. It would be better to generate the Alphabet letters from altLabels as well.

Similar to #729

osma:
You are right that this is a limitation. Basically this involves a trade-off between completeness and performance. The SPARQL query that finds all the letters for a particular vocabulary and language is already quite heavy even though it only looks at prefLabels. I suspect that also checking altLabels would make it heavier/slower still. It would be easy to change (see here: https://github.com/NatLibFi/Skosmos/blob/master/model/sparql/GenericSparql.php#L1297) for example by changing `skos:prefLabel` to the path expression `skos:prefLabel|skos:altLabel`.

Since you host a large number of vocabularies, including some large ones, maybe you could try this out on your Skosmos installation and see whether it adversely affects performance? If you think it's worth changing, then open a PR.

Kordishal:
Thanks Osma!

I have made the change and it is working. The performance hit for small and medium sized vocabularies is very small/barley noticeable, but for large vocabularies like MAO/TAO the load time for the Alphabetical list is a lot longer.

As such this change would only make sense for a small set of vocabularies. Would it be possible to make this a setting? Where would I have to start looking into this?

osma:
Sorry for the extremely late response @Kordishal - you are right that making this configurable would make a lot of sense. It could be a simple vocabulary-specific boolean setting, similar to e.g. `skosmos:fullAlphabeticalIndex`.

Issue #729, opened 2018-03-06
No labels displayed when only altLabels and no prefLabels for a language


Kordishal:
# The link where the issue occurs:
http://bartoc-skosmos.unibas.ch/Decoration/en/index/?clang=it

# What I would like to see:
The vocabulary has some languages without prefLabels, but various altLabels. It would be nice to be able to browse these concepts from these languages as well. As soon as there is a single prefLabel defined, altLabels are shown in the alphabetical tab, with an arrow to the prefLabel.

osma:
I think preprocessing the vocabulary would be a better option. prefLabels are pretty fundamental in SKOS. If a vocabulary doesn't have them, it should be fixed.

osma:
Closing this as stale. Please reopen if still relevant.

PR #1098, opened 2020-12-01
Optimize alphabet letter query: use GRAPH instead of FROM

osma:
It turns out that the existing SPARQL query to generate the letters of the alphabet for the alphabetical index is quite slow when using Fuseki TDB2. Switching to GRAPH instead of FROM makes it a lot faster.

See [this thread](https://lists.apache.org/thread.html/ree7a3136b972bacd4b89bb0ccd6cbf969516ab3313321880a3c3440b%40%3Cusers.jena.apache.org%3E) on the Jena users list for some more background.

### DIFF
--- a/model/sparql/GenericSparql.php
+++ b/model/sparql/GenericSparql.php
@@ -1337,15 +1337,17 @@ EOQ;
      * @return string sparql query
      */
     private function generateFirstCharactersQuery($lang, $classes) {
-        $fcl = $this->generateFromClause();
+        $gcl = $this->graphClause;
         $classes = (isset($classes) && sizeof($classes) > 0) ? $classes : array('http://www.w3.org/2004/02/skos/core#Concept');
         $values = $this->formatValues('?type', $classes, 'uri');
         $query = <<<EOQ
-SELECT DISTINCT (ucase(str(substr(?label, 1, 1))) as ?l) $fcl WHERE {
-  ?c skos:prefLabel ?label .
-  ?c a ?type
-  FILTER(langMatches(lang(?label), '$lang'))
-  $values
+SELECT DISTINCT (ucase(str(substr(?label, 1, 1))) as ?l) WHERE {
+  $gcl {
+    ?c skos:prefLabel ?label .
+    ?c a ?type
+    FILTER(langMatches(lang(?label), '$lang'))
+    $values
+  }
 }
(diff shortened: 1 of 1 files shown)


osma:
Some benchmarking of the query time for the KANTO/finaf vocabulary, using the dev.finto.fi SPARQL endpoint.

Before this PR: 16 seconds
After this PR: 1.7 seconds

The new query is almost 10x faster.

kouralex (commented):
Looks good and works, which is the most important part. I had trouble with this PR at first since there was no similar treatment for single alphabetical letter query so I added one. Feel free to re-review, git diff went bad since there are similar blocks.

One note worth thinking now that this is faster, I think #729 and #751 could be fixed with a simple skos:prefLabel|skos:altLabel change in the appropriate spot which you have already identified in the comment section of the latter. I leave this for you to decide if you want to add that to this PR given that I presume the ten-fold increase in performance could handle that/those use cases with no problems, now.

Lastly, I checked the code for JenaTextSparql and there was one place in code where the original issue was already fixed in the exact same way it is in this PR, so no further modifications are necessary.

osma:
> Looks good and works, which is the most important part. I had trouble with this PR at first since there was no similar treatment for single alphabetical letter query so I added one. Feel free to re-review, git diff went bad since there are similar blocks.

I benchmarked this change using KANTO and the alphabetical index page for the letter A. Before the PR, the query took 48s. After the change it was 12s. So a 4x improvement.

But the change you made is only relevant when using the Generic dialect. In the JenaText version of the alphabetical letter query, GRAPH is already used. I suspect you are using the Generic dialect, which is maybe not a great idea with large vocabularies such as KANTO/finaf.

> One note worth thinking now that this is faster, I think #729 and #751 could be fixed with a simple skos:prefLabel|skos:altLabel change in the appropriate spot which you have already identified in the comment section of the latter. I leave this for you to decide if you want to add that to this PR given that I presume the ten-fold increase in performance could handle that/those use cases with no problems, now.

I leave that to another PR which must be separately benchmarked.

> Lastly, I checked the code for JenaTextSparql and there was one place in code where the original issue was already fixed in the exact same way it is in this PR, so no further modifications are necessary.

Right. I think there is some room for refactoring now that the Generic and JenaText variants of the query are closer aligned.

PR #1102, opened 2020-12-07
add isset tests to fix undefined index PHP notices in RestController

kouralex:


osma:
The Travis CI build failure was transient and only happened on PHP 7.2 not other versions, so I will ignore it. Looks good, merging.

PR #1103, opened 2020-12-07
fixes undefined offset by adding an isset test

kouralex:
Whilst checking php error log for a Skosmos instance I noticed that the line 467 below https://github.com/NatLibFi/Skosmos/blob/254c2c66566527e873750fd36171086dc2d79417/controller/WebController.php#L443-L473 can cause `Undefined offset: 0` PHP Notice.

By following the code it calls  it seems like the `NULL` value is then passed here:
https://github.com/NatLibFi/Skosmos/blob/254c2c66566527e873750fd36171086dc2d79417/model/sparql/GenericSparql.php#L1187-L1216
where at line 1207 it is again (luckily) changed back to an empty string. This will, however, cause a filter clause that has a STRSTARTS() with an empty argument, which is unnecessary and will always evaluate to True.

Whilst this PR fixes the PHP notice, it won't change the behavior (i.e., the query as it is, now). I wonder if it would be more appropriate to, in this case, change the letter value to "\*", for which there is a special treatment (unlike for empty string). However, now that I look into that, it seems like the `STRSTARTS` part is unnecessary for the aforementioned "\*" value.

Additionally, I just noticed that this also affects the JenaTextSparql:
https://github.com/NatLibFi/Skosmos/blob/254c2c66566527e873750fd36171086dc2d79417/model/sparql/JenaTextSparql.php#L102-L171
Before merging it should be discussed whether something else should be done to both of these SPARQL parts, namely, if there should be a more special treatment for "*" and whether or not the letter value should be changed to that very same value.

### DIFF
--- a/controller/WebController.php
+++ b/controller/WebController.php
@@ -463,7 +463,7 @@ class WebController extends Controller
         if (!$allAtOnce) {
             $letters = $vocab->getAlphabet($contentLang);
             $letter = $request->getLetter();
-            if ($letter === '') {
+            if ($letter === '' && isset($letters[0])) {
                 $letter = $letters[0];
             }
             $searchResults = $vocab->searchConceptsAlphabetical($letter, $count, $offset, $contentLang);
--- a/model/sparql/GenericSparql.php
+++ b/model/sparql/GenericSparql.php
@@ -1327,6 +1327,9 @@ EOQ;
      * @param \EasyRdf\Resource|null $qualifier alphabetical list qualifier resource or null (default: null)
      */
     public function queryConceptsAlphabetical($letter, $lang, $limit = null, $offset = null, $classes = null, $showDeprecated = false, $qualifier = null) {
+        if ($letter === '') {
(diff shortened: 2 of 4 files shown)


osma:
The situation where this can happen is when a vocabulary has been configured to have a certain language, but there are actually no prefLabels with that language tag. The code in WebController tries to find the first alphabet letter for which prefLabels exist, but when there are no such prefLabels, the list is empty. This is basically a configuration error but these things happen, even in production as you've noticed by checking the logs.

I think we need a unit test to check for this situation. The suggested fix may be OK, I need to double-check. The sane result in this not-so-sane situation would be to show an empty list of alphabet letters as well as an empty alphabetical listing of concepts.

osma:
I will merge this assuming the Travis CI tests pass. The SonarCloud complaint is irrelevant, it complains about identical blocks in unit tests (true, but harmless).

osma:
Can't wait for Travis, but the tests pass locally, so merging this.

PR #1104, opened 2020-12-07
add an isset test and default value for possibly undefined index

kouralex:


joelit (approved):
I can verify that Skosmos can now tell us if there's no concepts in the vocabulary. Previously there was a header for the number of concepts, but no information could be shown.

PR #660, opened 2017-11-17
Issue #634 : Retrieve and display domains of ConceptScheme at the top of the tree hierarchy

tfrancart:
Coming back with a PR for #634 .
I have made this cleaner.
I am resubmitting the PR, hoping it can be useful and integrated in the code.
This answers 2 client use-cases I have (UNESCO and another).

Could you please let me know how can this proposal can progress to be integrated in the codebase ?

PR #1113, opened 2021-01-05
Issue 634 v2

tfrancart:
This revives https://github.com/NatLibFi/Skosmos/pull/660 with a clean commit, simply changing the way the hierarchy is displayed on the JS side.
It addresses https://github.com/NatLibFi/Skosmos/issues/634

Here is the result on the test data from the issue :

BEFORE

image

AFTER

image


resource/js/hierarchy.js
```diff
@@ -456,8 +553,35 @@ function getTreeConfiguration() {
                 else return -1;
             }
             else if (bNotation) return 1;
-        }
-        return naturalCompare(aNode.text.toLowerCase(), bNode.text.toLowerCase());
+        } else {
```
osma, line 556: This seems a bit overly complex. I tested replacing the whole block with this:

```js
        } else {
          // no sorting on notation requested
          // make sure the tree nodes with class 'domain' are sorted before the others
          // aDomain/bDomain will be "0" if a/b has a domain class, else 1
          var aDomain = (int) (aNode.original.a_attr['class'] != 'domain');
          var bDomain = (int) (bNode.original.a_attr['class'] != 'domain');
          return naturalCompare(aDomain + " " + aNode.text.toLowerCase(),
                                bDomain + " " + bNode.text.toLowerCase());
        }
```

It gave me the same result.

osma (changes requested):
Sorry for taking so long to review this. It looks like a solid improvement, even if we don't currently need it for any vocabulary in Finto. I gave a couple of improvement suggestions to the diff. Other than that, it looks good to me!

resource/js/hierarchy.js
kinow: Didn't you get any error due to the `(int)` casting? I normally use `parseInt()` or `Number` in JS/TS, and **think** that's still not available in ES6. But tested in a codepen [pen](https://codepen.io/kinow/pen/wvoRwym?editors=1111) just to make sure.

image

osma: Ah sorry, my bad! You are of course right @kinow, I was mixing up Java syntax with JS. Just shows how little I do JS coding :) And indeed, `showNotation` was enabled because that is the default setting for vocabularies, so execution never went to that `else` block.

Here is a better fix that I also tested, double-checking that it does get executed (I also checked that if I invert the logic here, the domains will be sorted last):

```js
        } else {
          // no sorting on notation requested
          // make sure the tree nodes with class 'domain' are sorted before the others
          // aDomain/bDomain will be "0" if a/b has a domain class, else "1"
          var aDomain = (aNode.original.a_attr['class'] == 'domain') ? "0" : "1";
          var bDomain = (bNode.original.a_attr['class'] == 'domain') ? "0" : "1";
          return naturalCompare(aDomain + " " + aNode.text.toLowerCase(),
                                bDomain + " " + bNode.text.toLowerCase());
        }
```

osma:
The changes I had requested were so trivial that I simply made them myself and merged this, so that it would make it into the next release 2.10. Thanks @tfrancart !

kouralex:
@osma you did not remove the commented out block. Could you please do that? Otherwise, great work!

osma:
> @osma you did not remove the commented out block. Could you please do that? Otherwise, great work, both of you!

I did remove it, but only when performing a manual merge to the master branch: https://github.com/NatLibFi/Skosmos/commit/7c95eedbf1b4d4823b097504b1ccd2668b59276c
Perhaps I should have pushed this commit (and the [other one for sorting](https://github.com/NatLibFi/Skosmos/commit/50923017773ec4546f4f06bf75a1ca1d10a13e1a)) into @tfrancart 's original branch, so it would have been visible here as well.

kouralex:
Oh yes, true. :) I was overly keen, should have checked the parent commit stack as well. Sorry for my mistake!

PR #1123, opened 2021-02-25
Add support for multilingual properties in case no hit in content lan…

### LINKED ISSUE #1115: The  skosmos:hasMultiLingualProperty setting doesn't work as intended
Sometimes it is useful to view information about a concept even though the information exists **only in another language than the selected content language**.  Typical examples are skos:note fields, which may exist only in the default language, while terms are offered in several languages. 
In SKOSMOS, this should be obtained by using the skosmos:hasMultiLingualProperty configuration setting

Example: The setting _skosmos:hasMultiLingualProperty skos:definition,  skos:scopeNote_
should imply  that all skos:definition and skos:scopeNote instances - in our case  written in the language Norwegian bokmål (nb)- are shown on their parent concept page also whenever another language, e.g. English is the selected content language.
See also [SKOSMOS user group conversation](https://groups.google.com/g/skosmos-users/c/QAJu_boLICw/m/iseYnt1tBAAJ).

Currently, in spite of the setting above, definitions and scopeNotes are shown only when content language is nb (like the language of the notes)

miguelahonen:
…guage; fixes #1115

### DIFF
--- a/model/Concept.php
+++ b/model/Concept.php
@@ -577,8 +577,8 @@ class Concept extends VocabularyDataObject implements Modifiable
                 // Iterating through every literal and adding these to the data object.
                 foreach ($this->resource->allLiterals($sprop) as $val) {
                     $literal = new ConceptPropertyValueLiteral($this->model, $this->vocab, $this->resource, $val, $prop);
-                    // only add literals when they match the content/hit language or have no language defined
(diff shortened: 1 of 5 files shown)


Vainonen:
Hyperlinks in multilingual properties were not working, so I decided to add support for them.

Good work @miguelahonen & @kouralex ! I will merge this now.

PR #1125, opened 2021-03-01
Issue1114 support newlines on concept page

kouralex:
This PR will bring proper support for newlines (converted into break tags) on both vocabulary and concept pages.

Additionally, will purify HTML content using a whitelist.

Tests are not passing as I removed (now futile) `getContainsHtml()` function but left tests for it untouched - one needs to write similar tests for `WebController` in order to test HTML purifier output (used in Twig filter but one may use `WebController->purifier->purify($content)` directly for testing).

Also, one may want to check https://github.com/ezyang/htmlpurifier/blob/master/INSTALL for more information on how to improve autoloading and (opcode) caching capabilities of HTMLPurifier and, thus, performance of the software.


kouralex:
Now that I tested this a bit it seems like the `nl2br` filter used in Twig template does not remove the possibly existing newline character, resulting in
```text element
<br>
text element starting with newline, following its text
```
It seems like we would be better of it.

A simple fix would consist of setting the following declaration for the container element: `white-space: pre-wrap`. This would even preserve any leading whitespace.

PR #1127, opened 2021-03-03
Upgrade to PHPUnit 8; bump required PHP to 7.2+; try to fix Travis tests on PHP 7.4

### LINKED ISSUE #920: Test failures on PHP 7.4
See the Travis build log: https://travis-ci.org/NatLibFi/Skosmos/jobs/645471054

1. Try to run Skosmos tests using PHP 7.4

The build fails with warnings like this:

    Error: implode(): Passing glue string after array is deprecated. Swap the parameters

There might be other errors as well but this was by far the most prominent.
Found while attempting to upgrade PHP, see #919

osma:
We have been using obsolete PHPUnit 7, which is not compatible with PHP 7.4 and is no longer supported.
This PR upgrades to PHPUnit 8, which is still supported (but only in Life Support mode).
It requires PHP 7.2, so this PR drops support for PHP 7.1 and attempts to get Travis CI tests working on PHP 7.4.

Some tests had to be modified (add void return types for setUp and tearDown methods) to keep PHPUnit 8 happy. There are warnings about deprecated PHPUnit features that should be addressed as well but they are not fatal (yet).

Fixes #920

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: required
 dist: trusty
 language: php
 php:
-  - 7.1
   - 7.2
   - 7.3
   - 7.4
@@ -35,6 +34,5 @@ env:
 matrix:
   exclude:
   allow_failures:
-  - php: 7.4
 notifications:
     slack: kansalliskirjasto:9mOKu3Vws1CIddF5jqWgXbli
--- a/composer.json
+++ b/composer.json
@@ -80,7 +80,7 @@
     "grimmlink/qtip2": "3.0.3"
   },
   "require-dev": {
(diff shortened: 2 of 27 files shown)


osma:
Tests are still failing on PHP 7.4. I think EasyRdf needs to be upgraded first...

osma:
EasyRdf update in PR #1128 needs to be merged first, then I can continue working on this PR.

osma:
Rebased on current `master` branch and force-pushed.

The PHPUnit 8 deprecation warnings should now be fixed; waiting for Travis CI tests to finish.

PR #1128, opened 2021-03-03
Upgrade EasyRdf to version 1.1.*

osma:
This PR upgrades EasyRdf to version 1.1.* (currently 1.1.1)
The only Skosmos code change, apart from changing the dependency, was to change two fields in SkosmosTurtleParser from `private` to `protected`, to avoid compiler errors.

Upgrading is required to make progress on issue #920 / PR #1127, since the old EasyRdf version we were using is not compatible with PHP 7.4.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
     "components/jqueryui": "1.12.*",
     "components/handlebars.js": "v4.7.6",
     "davidstutz/bootstrap-multiselect": "v0.9.13",
-    "easyrdf/easyrdf": "0.10.0-alpha.1",
+    "easyrdf/easyrdf": "1.1.*",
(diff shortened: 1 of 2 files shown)


osma:
Hmm, I think that actually SkosmosTurtleParser could now be simplified, since the changes to make the parser more efficient were merged into EasyRdf long ago and they are now in the 1.1.1 version we are upgrading to...

kouralex:
Perfect! It is a bit sad that we still need SkosmosTurtleParser as the parsed namespaces are not public. The only way I can think of is to save a reference to the parser object itself in GlobalConfig and call it directly but I guess that does not make much sense..

Anyways, will gladly merge! Very nice upgrade and needed, indeed! :)

Issue #875, opened 2019-05-22
We have spam


miguelahonen:
As we know, spambots are sending spam by making use of feedback form in Skosmos. Recently amount of spam has increased. We are very glad to hear good tips and suggestions to solve this issue :slightly_smiling_face:

*How to strengthen spam filtering*

Proposal 1:
In globalconfig adjust honeypot time

Proposal 2:

In most cases spambots read directly some URLs, I belive. People are more used to click or tab and “surf” on website and they do it mainly via frontpages.
Human user:
1) Goes to Finto.fi/fi (sv, en) frontpage
a. System: Sets up a cookie FBvisited=yes
b. System: Expiration = 3600 sec

2) By clicking the Feedback-link on the frontpage, the user is automatically redirected to the refresh page, which sets up a new cookie FeedVisited=yes and then redirects user to the real feedback page (by refresh action in some seconds).

3) On the real Feedback-page the form is not submitted before a function X has checked the values in the cookies. If both values are yes, the form can be submitted.

4) Result: Feedback is not sent if there is no visiting on the frontpage and Feedback-link is not tabbed or clicked in certain order. Spambot likely do not follow the order -> Front page -> Feedback refresh-page. To submit the form, the user have to visit firstly on Front page and then Feedback refresh-page.

Proposal 3:

1) Sets up a cookie visited=yes if user has been visiting on some concept page but the cookie is not set up on feedback page, the feedback page is excluded.

2) On the feedback-page the form is not submitted before a function X has checked the value in the cookie. If value is yes, the form can be submitted.

3) Result: Feedback is not sent if the visiting starts on the feedback page. We can assume that bots direct actions directly in some URL (page like http://finto.fi/fi/feedback).

miguelahonen:
If an action sending the feedback allowed only mouse clicks we could use jQuery to listen mouse click events?

"originalEvent.detail" is _true_ if the button is clicked by the mouse. Bots cannot click the mouse button.

---
```
<input id="this_does_something" type="button" value="Send feedback">
...
$('#this_does_something').click(function(e) {
  if (e.originalEvent.detail) {
    $(this).val('Clicked by mouse');
  } else {
    $(this).val('Not clicked by mouse!');
  }
});
```

osma:
My suggestion for some criteria for an acceptable solution:

[...]

PR #1131, opened 2021-03-04
Fix calculation of minimum time in feedback honeypot

osma:
While trying to get the Skosmos tests working under PHP 7.4 (#1127), I got some PHP errors from the code related to the honeypot designed to deter spammers from using the feedback form (related: #875). Digging further, I noticed that the calculations used to determine the minimum time between loading the feedback page and submitting the form were not correct; unfortunately the unit tests designed to check them were also buggy, so this was not discovered earlier! Probably the timer in the honeypot has never worked properly.

Anyway, this PR fixes several interrelated problems with the honeypot timer:

1. The parameters of Honeypot::validateHoneytime are now better specified; in particular, instead of a vague `$parameters` (which was assigned to an integer, but used as `$parameters[0]` - leading to the PHP7.4 error) it takes an integer parameter `$minDelta`
2. The Honeypot::decryptTime method now returns an integer, not a string. This is probably not crucial as PHP is pretty relaxed about typing, but it felt like a good idea anyway since this is used as an integer value.
3. The test method FeedbackTest::testHoneypotFieldsGenerated now properly mocks the `sendFeedback` method of the controller; it should always have done this, but not doing so didn't cause problems since other bugs obscured it.
4. Several methods in FeedbackTest where time offsets are used have been corrected. Previously they multiplied values by 60 (converted seconds to minutes) for no reason. I avoided values that were just at the limit of the uiHoneypotTime value, so I used 2 seconds and 6 seconds instead of 0 seconds and 5 seconds, which are special cases and could easily lead to wrong conclusions (and in fact were).

### DIFF
--- a/controller/Honeypot.php
+++ b/controller/Honeypot.php
@@ -11,14 +11,14 @@ class Honeypot
     /**
      * Enable the Honeypot validation
      */
-    public function enable()
+    public function enable() : void
     {
         $this->disabled = false;
     }
     /**
      * Disable the Honeypot validation
      */
-    public function disable()
+    public function disable() : void
     {
         $this->disabled = true;
     }
@@ -28,7 +28,7 @@ class Honeypot
      * @param  string $honey_time
      * @return string
      */
-    public function generate($honey_name, $honey_time)
+    public function generate($honey_name, $honey_time) : string
     {
         // Encrypt the current time
         $honey_time_encrypted = $this->getEncryptedTime();
@@ -43,7 +43,7 @@ class Honeypot
     * @param  mixed $value
     * @return boolean
     */
-    public function validateHoneypot($value)
+    public function validateHoneypot($value) : bool
     {
         if ($this->disabled) {
             return true;
@@ -57,7 +57,7 @@ class Honeypot
      * @param  int $minDelta minimum time difference in seconds
      * @return boolean
      */
-    public function validateHoneytime($value, $minDelta)
+    public function validateHoneytime($value, $minDelta) : bool
     {
         if ($this->disabled) {
             return true;
@@ -72,7 +72,7 @@ class Honeypot
      * Get encrypted time
      * @return string
      */
-    public function getEncryptedTime()
+    public function getEncryptedTime() : string
     {
         return base64_encode(time());
     }
@@ -82,7 +82,7 @@ class Honeypot
      * @param  mixed $time
      * @return string|null
      */
-    public function decryptTime($time)
+    public function decryptTime($time) : int
     {
         try {
             return intval(base64_decode($time));
--- a/controller/Honeypot.php
+++ b/controller/Honeypot.php
@@ -80,9 +80,9 @@ class Honeypot
      * Decrypt the given time
      *
      * @param  mixed $time
-     * @return string|null
+     * @return int|null
      */
-    public function decryptTime($time) : int
+    public function decryptTime($time) : ?int
     {
         try {
             return intval(base64_decode($time));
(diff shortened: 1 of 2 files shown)


controller/Honeypot.php
```diff
@@ -85,7 +85,7 @@ public function getEncryptedTime()
     public function decryptTime($time)
     {
         try {
-            return base64_decode($time);
+            return intval(base64_decode($time));
```
kouralex, line 88: Should this change be reflected at line 5?

osma: The explanation on line 5 is a bit vague perhaps, but I don't think it needs to be changed due to this. It's not more wrong than it used to be ;)

tests/FeedbackTest.php
```diff
@@ -68,7 +74,8 @@ public function testHoneypotAndHoneypot() {
     $this->request
         ->shouldReceive('getQueryParamPOST')
         ->with('user-captcha')
-        ->andReturn(base64_encode(time() - 5 * 60));
+        // 6 seconds ago is more than the default 5 seconds
+        ->andReturn(base64_encode(time() - 6));
```
kouralex, line 78: The base64_encode function will do nothing to the value here.

tests/FeedbackTest.php
```diff
@@ -21,6 +21,12 @@ protected function setUp() {
    * @covers Honeypot::getEncryptedTime
    */
   public function testHoneypotFieldsGenerated() {
+    $this->controller
+        ->shouldReceive('sendFeedback')
+        ->withAnyArgs()
+        ->once()
+        ->andReturn(true);
+
```
kouralex, line 29: Can your shortly explain the combination of methods used here, and why?

tests/FeedbackTest.php
osma: I don't understand?
Here the mocked `getQueryParamPOST` method is used to pass parameters, in a roundabout way, to the controller method being tested here. It is given an argument `time() - 6` i.e. an integer representing the Unix timestamp six seconds ago. PHP implicitly converts this to a string. Then base64_encode returns it as a base64 encoded string; and this is the "parameter" that eventually is passed to the controller.

tests/FeedbackTest.php
osma: This overrides the `sendFeedback` method of `$this->controller` (which is a partial mock class based on WebController) with a mock method that does nothing except returns `true`.

`withAnyArgs` means that any arguments are OK, the test will not fail no matter what was passed to the mock method.

`once` means that it is expected to be called exactly once; otherwise, the test will fail.

This is the same kind of mock that was already used in several other methods in this class, but it was previously forgotten for this method. What happened was that the original WebController::sendFeedback got called, and it tries to send an actual e-mail message, which fails since there is no recipient address. We don't want to call the true sendFeedback method in the test, just want to verify that it would have been called once.

kouralex (commented):
Nice catch! I have nothing to add except for the comments I left in the code + the common case of perhaps adding the return type to the function (new feature in PHP7) in case you are anyways changing it.

You may very well merge after evaluating those.

Let's hope for the best that this will stop the spam altogether. 😃

tests/FeedbackTest.php
kouralex: Oh, I tested badly, my bad! 😄

kouralex (approved):

kinow:
>unfortunately the unit tests designed to check them were also buggy, so this was not discovered earlier! Probably the timer in the honeypot has never worked properly.

Ah! My bad, and great job fixing it! Thanks for fixing it @osma!

PR #1132, opened 2021-03-04
Add a noscript tag to alert users when they have not enabled JS

kinow:
While testing the #700 PR, I refreshed the page a few times thinking there was something wrong in my container. Then I realized NoScript had prevented the JS of running in the Skosmos application.

Without JavaScript it's not able to search or navigate the application, so I have added in this PR the exact same `noscript` that comes by default in Vue.js applications (I believe React's message/placement is very similar).

That's useful for users with extensions such as NoScript. I know a few other friends using it here in New Zealand (even my wife now), but I guess in Europe there might be some too that are concerned with cookies/privacy/etc. So maybe it's worth alerting them to enable for Skosmos?

Cheers
Bruno

Screenshot from 2021-03-05 10-58-30

### DIFF
--- a/view/light.twig
+++ b/view/light.twig
@@ -20,6 +20,9 @@
 <title>{{ ServiceName }}{% block title %}{% endblock %}</title>
 </head>
 <body{% if request.vocabid == '' and request.page != 'feedback' and request.page != 'about' %} class="frontpage-logo"{% else %} class="vocab-{{ request.vocabid }}"{% endif %}>
+  <noscript>
+    <strong>We're sorry but Skosmos doesn't work properly without JavaScript enabled. Please enable it to continue.</strong>
+  </noscript>
   <a id="skiptocontent" href="{{ request.langurl }}#maincontent">{% trans "Skip to main" %}</a>
   {% if request.vocabid != '' or request.page == 'feedback' or request.page == 'about' %}<div class="topbar-white">{% endif %}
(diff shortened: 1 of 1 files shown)


osma:
Thanks @kinow! I think this needs i18n support though, so Skosmos can show the message in the current UI language.

kinow:
> Thanks @kinow! I think this needs i18n support though, so Skosmos can show the message in the current UI language.

Ah, true @osma. I've used the Twig macro to translate the string (is that all?). Then logged in to Transifex to see if I could add the entry there, but looks like someone else will have to do that first. I will add the Portuguese translation later once it's been added to Transifex.

Thanks!

osma:
I added the entries to the po files for English, Finnish and Swedish, which are maintained within the Skosmos repo (not Transifex), then merged this to `master`. Now Transifex should pick up the strings to translate from the skosmos_en.po file, so it should be possible to add the translations for e.g. Portuguese soon (I think that Transifex refreshes once per day).

Thanks again @kinow!

kinow:
Will refer to [your commit](https://github.com/NatLibFi/Skosmos/commit/20d994bdbe7851a5469659923f8be794ea5c690d#diff-750ce7e88dfc6cee6192b9c8e7ca956e4fe7f83dccd57b5ce131cd5090d91e12R852) next time I need to add a text string :+1: 

Thank you @osma!

PR #1133, opened 2021-03-09
A validation improved (solid red lines around the text fields in all …

miguelahonen:
A validation improved (solid red lines around the text fields in all validated and failed cases). In addition, WCAG 2.1 AA level related global style definition for HTML label tag now applies only to feedback form

### DIFF
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -774,7 +774,7 @@ span.xl-pref-label > img {
 /* feedback page
  ***************************/
 #feedback-fields > p .missing-value {
-  border: 1px solid red;
+  border: 2px solid #D95F8A;
 }
 
 #feedback-fields label {
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -5,10 +5,6 @@ li, td > a {
   font-size: 14px;
 }
 
-label {
-  font-weight: normal;
-}
-
 h1, .prefLabel, .prefLabelLang, .notation {
   font-family: 'Fira Sans', sans-serif;
   font-size: 24px;
@@ -777,8 +773,12 @@ span.xl-pref-label > img {
 
 /* feedback page
  ***************************/
-#feedback-fields > .missing-value {
+#feedback-fields > p .missing-value {
   border: 1px solid red;
+}
+
+#feedback-fields label {
+  font-weight: normal;
 }
 
 #message {
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -781,6 +781,12 @@ span.xl-pref-label > img {
   font-weight: normal;
 }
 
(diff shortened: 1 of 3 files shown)


resource/css/styles.css
```diff
@@ -777,10 +773,14 @@ span.xl-pref-label > img {
 
 /* feedback page
  ***************************/
-#feedback-fields > .missing-value {
+#feedback-fields > p .missing-value {
   border: 1px solid red;
```
osma, line 777: I think it would be better to use the color #D95F8A that Skosmos uses for deprecated concepts, see https://github.com/NatLibFi/Skosmos/blob/a201ace07d0aad5c3ddf4c2dbbbe045f4834ff4c/resource/css/styles.css#L1209-L1212
(I didn't check how that would look though)

osma (changes requested):
I tested this locally and it looks like this when I try to submit an empty form:

image

Comments/requests:

1. The color is now red. I suggest using the pinkish color Skosmos already uses to signal deprecated concepts (see specific comment on CSS file)
2. The form the text `* Value is required and can not be empty` at the bottom, but there are no asterisks next to the Subject and Message fields (or anywhere!). Could you please add asterisks after to the labels to mark the required fields?
3. The default text "Write a subject" in the subject field looks the same as normal text entered into the field. This is a bit confusing, especially since the field will be considered empty for validation purposes even if the default text is there (see screenshot above). Could you please adjust the color/style of the default text so that it doesn't look like regular text, for example by making it a lighter shade?

I noticed that Bootstrap 3 (which Skosmos still uses) has [special CSS classes for validation states](https://getbootstrap.com/docs/3.3/css/#forms-control-validation) and we could consider using them instead of a custom CSS class as is done here, but in practice, this would probably be more trouble than it's worth, as we would then have to override much of the Bootstrap styling anyway.

resource/css/styles.css
miguelahonen:
```
#feedback-fields > p .missing-value {
  border: 1px solid red;
}
```

Changed to ->

```
#feedback-fields > p .missing-value {
  border: 2px solid #D95F8A;
}
```

Contrast tests for text box with new border style passed
(https://contrastchecker.com/)

miguelahonen: Asterisks added

osma (changes requested):
I tested this locally.

1. Field outline color is now good!
2. Asterisks are OK!
3. The default text still looks too similar to user-entered text:

image

Can you please make the default text a lighter shade, so it doesn't look like the user has actually entered e.g. "Write a subject" into the subject field?

miguelahonen:
Placeholder texts in the input fields are now lighter and italic 
Lighter_and_italic

PR #1134, opened 2021-03-10
Automatically load rest of alphabetical index on vocabulary home page

### LINKED ISSUE #1108: Long lists not scrollable all way through
For example, when scrolling down YSO alphabetical tab with no letter selected, scroll starts from A and only reaches down to  less than midpoint of the list. See attached image: could only scroll down to YSO's agentiivisuus (kielitiede) and not further. I was using Mozilla Firefox. When a letter is clicked, however, the list is scrollable all way through.
yso-schreenshot-skosmos-scroll-problem-2020

osma:
This PR fixes #1108: it wasn't possible to scroll down to see the full alphabetical index of the first letter (typically A) on the vocabulary front page. The normal mechanism that triggers an AJAX request when scrolling down to the bottom of the first 250 results wasn't triggered on the vocabulary home page; it only worked on the alphabetical index page.

As analyzed in [this comment](https://github.com/NatLibFi/Skosmos/issues/1108#issuecomment-770938478) the reason is that the callback for loading more lines isn't properly activated. The `if` check is too strict, it only works on the `index` page. I avoided the `if` check altogether and simply targeted the `sidebar-grey-alpha` CSS class, which is specific to the alphabetical index listings, instead of the more generic `sidebar-grey` class.

### DIFF
--- a/resource/js/docready.js
+++ b/resource/js/docready.js
@@ -57,16 +57,14 @@ $(function() { // DOCUMENT READY
         });
         return removeThese.join(' ');
       });
-      if (settings.url.indexOf('index/') !== -1) {
-        $(".sidebar-grey").mCustomScrollbar({
-          alwaysShowScrollbar: 1,
-          scrollInertia: 0,
-          mouseWheel:{ preventDefault: true, scrollAmount: 105 },
-          snapAmount: 15,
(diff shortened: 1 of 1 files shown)


osma:
Thanks for testing @kinow!

PR #1135, opened 2021-03-11
Changed the description of search parameter maxhits

### LINKED ISSUE #1112: REST: incorrect swagger.json description for maxhits parameter
The current description for `maxhits` parameter seems to be outdated as it will return all values in case no parameter is set. For example, https://api.finto.fi/rest/v1/search?vocab=koko&query=n*&lang=fi will return whopping 2343 results instead of the 100 mentioned in swagger.json (see below).
https://github.com/NatLibFi/Skosmos/blob/254c2c66566527e873750fd36171086dc2d79417/swagger.json#L101-L107

However, if the parameter is used, it is honored as, for example, https://api.finto.fi/rest/v1/search?vocab=koko&query=n*&lang=fi&maxhits=100 will return exactly 100 results, as expected.

I believe that this is a result of issue #442 which was fixed by https://github.com/NatLibFi/Skosmos/commit/763b3f45e5941f9b5aacc76e705865640574a722. There seems to be mentions about the code change in https://github.com/NatLibFi/Skosmos/issues/524#issuecomment-225168890 and in https://github.com/NatLibFi/Skosmos/wiki/Configuration (`DEFAULT_SEARCH_LIMIT` parameter has been removed as of Skosmos 1.7).

The code responsible for setting the 'default' value is here:
https://github.com/NatLibFi/Skosmos/blob/254c2c66566527e873750fd36171086dc2d79417/model/ConceptSearchParameters.php#L149-L155

As one can see from the line numbers 151-153, empty value is treated as `0` (effectively the same as no maxlimit), consistent with the observed behavior.

Now, there are two possible fixes for this:
1) update the incorrect part in swagger.json description
2) support this behavior with yet another Skosmos configuration parameter.

The issue with the first fix is that long lists of results take some time to be generated, which may get halted due to a timeout and, thus, result in broken JSON to be returned to the user. However, I believe that #442 cannot truly be fixed without returning all values in some [visually] accessible fashion. One option is, of course, slicing (maxhits) with offset as is already done in alphabetical listing (still keeping in mind the reason why #442 was fixed the way it was). The other option is just to raise the timeout limits (if needed) and keep the behavior unchanged.

The second fix would introduce once-removed feature again in the code base.

joelit:
Fixes #1112

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -101,7 +101,7 @@
           {
             "name": "maxhits",
             "in": "query",
-            "description": "Maximum number of results to return. If not given, maxhits will default to 100. (default settable in config.ttl)",
+            "description": "Maximum number of results to return. If not given, all results will be returned.",
             "required": false,
             "type": "integer"
           },
@@ -485,7 +485,7 @@
           {
             "name": "maxhits",
             "in": "query",
-            "description": "Maximum number of results to return. If not given, maxhits will default to 100. (default settable in config.ttl)",
+            "description": "Maximum number of results to return. If not given, all results will be returned.",
             "required": false,
             "type": "integer"
           },


osma (approved):
Looks good! +1 for merging.

PR #1138, opened 2021-03-16
Add generator meta tag with Skosmos version

### LINKED ISSUE #1116: Add generator meta tag in HTML output to show Skosmos version
I tried to automatically detect whether a page is run by Skosmos. Could you please add a HTML header such as

~~~html
<meta name="generator" content="Skosmos 2.8" />
~~~

osma:
Fixes #1116 

Adds this kind of meta tag, as requested by @nichtich:

    <meta name="generator" content="Skosmos v.2.9-93-g20d994b" />

### DIFF
--- a/model/Request.php
+++ b/model/Request.php
@@ -266,4 +266,13 @@ class Request
         }
         return new PluginRegister($this->model->getConfig()->getGlobalPlugins());
     }
+
+    /**
+     * Return the version of this Skosmos installation, or "unknown" if
+     * it cannot be determined. The version information is based on Git tags.
+     * @return string version
+     */
+    public function getVersion() {
+        return $this->model->getVersion();
+    }
 }
--- a/model/Request.php
+++ b/model/Request.php
@@ -272,7 +272,8 @@ class Request
      * it cannot be determined. The version information is based on Git tags.
      * @return string version
      */
-    public function getVersion() {
+    public function getVersion() : string
+    {
         return $this->model->getVersion();
     }
 }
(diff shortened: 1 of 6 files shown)


model/Request.php
```diff
@@ -266,4 +266,13 @@ public function getPlugins() {
         }
         return new PluginRegister($this->model->getConfig()->getGlobalPlugins());
     }
+
+    /**
+     * Return the version of this Skosmos installation, or "unknown" if
+     * it cannot be determined. The version information is based on Git tags.
+     * @return string version
+     */
```
kouralex, line 274: Missing return type(s).

view/about.twig
```diff
@@ -15,6 +15,7 @@
     <p>{% if ServiceName != 'SERVICE_NAME' %}{{ ServiceName }}{% else %}Skosmos{% endif %} {% trans "layout designed by Hahmo" %}</p>
   </div>
   <div class="version">
+    {% set version = request.version %}
```
kouralex, line 18: One can set a variable or use it directly, OK anyways. 😄

kouralex (commented):
Does the job. Just minor comments, can be merged.

Good work (as always), @osma !

view/about.twig
osma: Only simple variables can be used with `trans`. And the variable name has to match what has been used before, otherwise all the translated strings would be invalidated.

Yes, a global variable would have been possible too. I decided to tack this into Request, since it's passed to all templates anyway and it already had a reference to Model, so it was trivial to delegate to Model->getVersion.

model/Request.php
osma: I added these in a new commit.

osma:
Waiting for Travis builds to finish, then I will merge this.

PR #1140, opened 2021-03-18
Fix sorting of concepts in hierarchy that broke in PR 1133

osma:
In PR #1133, the sorting of concepts in the hierarchy view was modified to take into account the domain (if any) of concept schemes. However, as a side effect, it broke the sorting in a frequently occurring situation: when sortByNotation is enabled (which is the default) but no notation codes are actually defined in the vocabulary. This happens e.g. in YSO.

Original situation (good), concepts are alphabetically ordered:

image

Situation after PR #1133 when sortByNotation is enabled (bad), not ordered:

image

The problem was that PR #1133 introduced an `if...else` structure but it didn't take into account the possibility of having sortByNotation enabled but not having any notations in the vocabulary data.

The solution is simply to remove the `else` block, so that the code that used to be in the `else` block is also executed if execution falls through the other `if` clauses. I've also added comments to highlight this somewhat unusual execution flow.

Thanks to @kouralex for originally reporting this!

### DIFF
--- a/resource/js/hierarchy.js
+++ b/resource/js/hierarchy.js
@@ -507,7 +507,7 @@ function getTreeConfiguration() {
         var aNode = this.get_node(a);
         var bNode = this.get_node(b);
 
-        // sort on notation if requested
+        // sort on notation if requested, and notations exist
         if (window.showNotation) {
             var aNotation = aNode.original.notation;
             var bNotation = bNode.original.notation;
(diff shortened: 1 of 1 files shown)


kouralex (approved):
This seems to do the trick! Good work, @osma !

For fun, one can check how this is done (except for domains) in PHP code here:
https://github.com/NatLibFi/Skosmos/blob/e21309631037401c4b5a9d171470655884eab678/model/ConceptProperty.php#L113-L146

As it is standing, there is no handling of similar labels but with different URIs but that is the case for above PHP code as well (see TODO comment above). This may result in different order in cases that are common in e.g.,  special ontologies that extend the more general one as in the original image. This can be seen below where the order of black-blue pairs are in inconsistent state. But this is how it already was so no harm done. Also, it seems to be happening in quite rare conditions now that I just tested (usually the black one is first, black meaning the more general ontology).
image 
That said, this is an improvement and very well meets Skosmos quality measures and can be merged as-is.

osma:
Thanks for the review @kouralex ! I'm merging this as is. If you want you can open a separate issue for the "similar labels but with different URIs" case.

Issue #727, opened 2018-03-06
Utilize typed parameters introduced in php7


kouralex:
Whilst the master branch already requires PHP 7.x, it is not coded using the PHP 7.x syntax. If we will rewrite our code to use the typed parameters introduced in php7, we would arguably improve our code base and make it a lot easier to counter bugs.

kinow:
+1, last I checked there were a couple of places that phpstorm was able to identify some typing issues, even though we lack complete annotation. Only minor concern is that I think we are stuck at PHP 7.0 until that JSON-LD issue with newer versions is sorted.

osma:
In many cases this is already done.

Closing this as stale. Please reopen if still relevant.

PR #1142, opened 2021-03-22
Fix typos in comments, add phpdocs, replace NBSP by space, simplify phpunit

kinow:
Happy to drop some commits if necessary :+1: Just syncing my local fork branch and taking a quick look at the current code base.


controller/Controller.php
```diff
@@ -69,7 +69,7 @@ public function setLanguageProperties($lang)
      * As a side effect, set the HTTP Vary header if a choice was made based on
      * the Accept header.
      * @param array $choices possible MIME types as strings
-     * @param string $accept HTTP Accept header value
+     * @param string $accept HTTP Accept header value
```
kinow, line 72: Hmm, doesn't show on the GitHub UI. But we have a NBSP (unicode U+00A0), not a space here (U+0020).

model/Concept.php
```diff
@@ -436,14 +436,14 @@ public function getMappingProperties(array $whitelist = null)
                             }
 
                             if ($response) {
-                                $ret[$prop]->addValue(new ConceptMappingPropertyValue($this->model, $this->vocab, $response, $this->resource, $prop), $this->clang);
```
kinow, line 439: Funny, these `clang` parameters are not in the `addValue` signature.

model/resolver/WDQSResource.php
```diff
@@ -6,7 +6,6 @@ class WDQSResource extends RemoteResource
     const WDQS_ENDPOINT = "https://query.wikidata.org/sparql";
 
     public function resolve(int $timeout) : ?EasyRdf\Resource {
-        $client = new EasyRdf\Sparql\Client(self::WDQS_ENDPOINT);
```
kinow, line 9: Re-declared in the code below

model/Concept.php
kouralex: Yes, seems like the signature was changed some five years ago: https://github.com/NatLibFi/Skosmos/commit/2ba0a48174df87965e7785c8196ce2b039a85050#

Good catch!

controller/WebController.php
```diff
@@ -588,10 +587,11 @@ public function getChangeList($request, $prop, $offset=0, $limit=200)
     }
 
     /**
-    * Formats the list of concepts as labels arranged by modification month
-    * @param Array $changeList
-    * @param string $lang the language for displaying dates in the change list
-    */
+     * Formats the list of concepts as labels arranged by modification month
+     * @param Array $changeList
```
kouralex, line 591: It seems like there are PHP types stated in both uppercase and lowercase (starting with) in PHPDocs all over the code base (array, for instance, here).

Well, I guess they do not matter. 😅

model/resolver/WDQSResource.php
kouralex: Good catch!

Now that I look at it, it is a bit questionable why lines https://github.com/NatLibFi/Skosmos/blob/4774577479d618932ba68c44c2e59ff1af050c9d/model/resolver/WDQSResource.php#L10-L44 are in the `try` block, though. I would expect them to succeed in any case.

controller/WebController.php
kinow: I noticed that too. I **think** in this case I opened another code in this function, that returned `Array`, and then I just typed `Array` too :smile: but normally I use `string`, `boolean`, `array`…

model/resolver/WDQSResource.php
kinow: Hmm, good point. We can move it out maybe in another PR? One thing that really bothers me in the code is the use of `return $this->showError…` functions, while these `showError`-etc functions actually do not return anything; but instead set HTTP `header()` values, print to the stdout with `echo()` and need the stream to be closed (with `return` or `exit`) somewhere.

e.g.

- https://github.com/NatLibFi/Skosmos/blob/c84dc970cea2954c76c98254c20f72d6da33756a/controller/RestController.php#L49 returns the value of https://github.com/NatLibFi/Skosmos/blob/c84dc970cea2954c76c98254c20f72d6da33756a/controller/Controller.php#L179-L184 (which doesn't actually return anything).

tests/GenericSparqlTest.php
```diff
@@ -1051,7 +1051,7 @@ public function testListConceptGroups()
     $voc = $this->model->getVocabulary('groups');
     $graph = $voc->getGraph();
     $sparql = new GenericSparql('http://localhost:13030/skosmos-test/sparql', $graph, $this->model);
-    $actual = $sparql->ListConceptGroups('http://www.w3.org/2004/02/skos/core#Collection', 'en', false);
+    $actual = $sparql->ListConceptGroups('http://www.w3.org/2004/02/skos/core#Collection', 'en');
```
kouralex, line 1054: Good fix!

Oh my, PHP truly is case-insensitive. Hurts my eyes! 😄

model/resolver/WDQSResource.php
kouralex: Yes, best to leave it to another PR, this one is rock solid as it is! 😃 

Heh, now that you pointed at it, it surely looks ugly. I also remember having some trouble with those things whilst trying to silence the extra output PHPUnit was printing whilst executing tests.

+1 for another PR.

controller/WebController.php
kouralex: I guess it is OK at this point.

We should anyways add type hints/declarations directly into function arguments and return values as requested in #727 but that can be done en masse, simultaneously double checking these PHPDocs.

kouralex (approved):
Very solid improvements in this PR, thanks @kinow !

I am more than happy to merge this as-is.

kinow:
Thanks for the review and helpful comments @kouralex !

kouralex:
I merged this now that I saw your last replies. 🎉

PR #1149, opened 2021-04-07
Avoid sending empty email when invoking the feedback form. 

### LINKED ISSUE #1139: Using the feedback form an empty email is sent
http://dev.finto.fi
http://dev.finto.fi/fi/feedback
http://dev.finto.fi/sv/feedback
http://dev.finto.fi/en/feedback

Entering the url http://dev.finto.fi/fi/feedback (including also sv and en addresses) or clicking the link Palaute/Feedback/Respons at url http://dev.finto.fi causes the sending of an empty email to the defined feedback recipient address.

## What is the expected output?
Only a normal feedback message with a topic and message body. 

## What do you see instead?
in addition to the normal message, it sends an empty and unnecessary message.

Firefox

osma:
There was a logic error in WebController. The feedback message is always sent even when just displaying the feedback form. This was previously hidden by the honeypot check, but the bug surfaced after the changes to honeypot code in PR #1131.

Fixes #1139

### DIFF
--- a/controller/WebController.php
+++ b/controller/WebController.php
@@ -206,22 +206,21 @@ class WebController extends Controller
         $vocab = $request->getVocab();
 
         $feedbackSent = false;
-        $feedbackMsg = null;
         if ($request->getQueryParamPOST('message')) {
             $feedbackSent = true;
             $feedbackMsg = $request->getQueryParamPOST('message');
-        }
-        $feedbackName = $request->getQueryParamPOST('name');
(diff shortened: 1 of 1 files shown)


miguelahonen (approved):

miguelahonen:
Relevant code changes :+1:

miguelahonen:
Changes tested and everything works well :+1: :1st_place_medal:

Issue #1156, opened 2021-04-08
docker fails to install php composer dependencies


jrvosse:
## At which URL did you encounter the problem?
root url
## What steps will reproduce the problem?
1. start with docker-compose up -d
2. visit localhost:9090
3.

## What is the expected output? What do you see instead?
Expected: home page of skosmos
Instead I get 
`Error: Dependencies managed by Composer missing. Please run "php composer.phar install".
`
Error goes away if I manually (by running bash in the skosmos container) install composer and run the command above.

## What browser did you use? (eg. Firefox, Chrome, Safari, Internet explorer)
Chrome

osma:
Hi @jrvosse , was this with the recent 2.10 release or an earlier version?

Docker support was completely redone in the 2.10 release. It's documented in the [wiki](https://github.com/NatLibFi/Skosmos/wiki/Install-Skosmos-with-Fuseki-in-Docker) and in the [dockerfiles](https://github.com/NatLibFi/Skosmos/tree/master/dockerfiles) directory.

osma:
Ah, now I think I understand what's wrong here. We changed the way the Docker image is built several times during the development of the new Dockerfile, and the current version copies all files from the Skosmos directory inside the image. This means that Composer must be run *outside* the container, before building the image. That's not ideal - I think Composer should be run *within* the container as it's being built, and probably the vendor directory etc. should be excluded while copying so that Composer has a clean starting point. @kinow do you have any ideas?

kinow:
Ah, your analysis is correct @osma ! And good catch @jrvosse , we missed that part in the review as we always had the composer dependencies installed, sorry! I think at one point we had composer running, but as @osma said changes during development removed it.

Let me take a look at #1157, I think that's safe to merge, but will confirm it won't change host files (it shouldn't I think, the composer dependencies should be installed in the container only).

Bruno

osma:
Fixed by PRs #1157 and #1162

PR #1157, opened 2021-04-08
Install composer.phar dependencies

jrvosse:
See issue #1156 
Added two lines to download composer.phar and install the dependencies.


osma:
I wanted to merge this but add a `.dockerignore` file in the same go. But since I don't have push access to the PR branch used here, I had to do it via an additional PR - #1162 . Since that PR is based on this PR, merging it will also merge the changes in this PR.

PR #1158, opened 2021-04-12
Fix: Sidebar height calculation

danmichaelo:
The sidebar is a few pixels too high when the alphabetical list is active, due to the padding of the div containing the letters. This should fix it.

### DIFF
--- a/resource/js/scripts.js
+++ b/resource/js/scripts.js
@@ -314,8 +314,8 @@ function countAndSetOffset() {
   /* calculates the sidebars content maximum height and sets it as an inline style.
      the .css() can't set important so using .attr() instead. */
   $('.sidebar-grey').attr('style', function() {
-    var pixels = $('.nav-tabs').height() + 2; // the 2 pixels are for the borders
-    if ($('#sidebar > .pagination').is(':visible')) { pixels += $('.pagination').height(); }
+    var pixels = $('.nav-tabs').outerHeight();
+    if ($('#sidebar > .pagination').is(':visible')) { pixels += $('.pagination').outerHeight(); }
     return 'height: calc(100% - ' + pixels + 'px) !important';
   });
   var $sidebar = $('#sidebar');


kinow (approved):
That's some sharp eye @danmichaelo :clap: 

Here's a zoomed-in view of https://dev.finto.fi/ysapaikat/en/page/Y169060?clang=fi (in XFCE/Ubuntu, ALT + Mouse scroll gives you control over monitor zoom).

image

Then applying your JavaScript changes locally with the following script in the browser console:

```js
var pixels = $('.nav-tabs').outerHeight();
if ($('#sidebar > .pagination').is(':visible')) { pixels += $('.pagination').outerHeight(); }
$('.sidebar-grey').attr('style', pixels)
```

image

Worked for me on Firefox + Ubuntu. Might be good to check the other browsers perhaps, but using `outerHeight` sounds like the right approach, instead of using a `2px` buffer :+1: 

Bruno

danmichaelo:
Actually I don't have *that* good eyes 😆 I realize that my description above was a bit brief – I forgot to mention that the problem was that the extra height causes a scrollbar on pages that should not have a scrollbar (only when using the alphabetical tab):

Screenshot 2021-04-14 at 19 53 44


That the patch also removes the extra padding on the hierarchical tab was just a lucky side effect 😄

I tested the following snippet in the browser console in Firefox, Chrome and Safari on Mac now:

```
$('.sidebar-grey').attr('style', function() {
    var pixels = $('.nav-tabs').outerHeight() ;
    if ($('#sidebar > .pagination').is(':visible')) { pixels += $('.pagination').outerHeight(); } 
    return 'height: calc(100% - ' + pixels + 'px) !important';
});
```

Seems to give a tight fit on the alphabetical and hierarchical tab 👍

kinow:
I cannot reproduce that on Firefox or Chromium on Ubuntu, but could be happening on other platforms/browsers.

image

danmichaelo:
> I cannot reproduce that on Firefox or Chromium on Ubuntu, but could be happening on other platforms/browsers. (screenshot without any changes to Skosmos).
> 
> image

You are on the hierarchy tab, could you check if you can reproduce in on the A-Z tab? For me, it's also not a problem on the hierarchy tab.

PR #1159, opened 2021-04-12
Add topbar-container css hook

danmichaelo:
At the moment I'm not sure if it's possible to make the frontpage topbar look the same as the topbar on other pages with just CSS, since the outer div is missing on the frontpage.

This PR adds an outer div to the frontpage topbar, which makes it easier to style the frontpage topbar as you like. I've added a class name "topbar-container", but otherwise kept the old class names, so I wouldn't expect this change to break anyone's CSS.

## Frontpage

```diff
+ <div class="topbar-container topbar-frontpage">
    <div class="topbar frontpage">
```

## Non-frontpage

```diff
- <div class="topbar-white">
+ <div class="topbar-container topbar-white">
    <div class="topbar topbar-white">
```


osma (approved):
Thanks! Looks good to me and doesn't break the Finto layout which is obviously a plus ;)

PR #1165, opened 2021-05-20
Show newly deprecated

joelit:
This is an otherwise ready PR - however, the labels for new concepts replacing deprecated ones is not implemented since it caused a significant regression in performance. A follow-up fix will come later.


controller/RestController.php
```diff
@@ -1104,9 +1104,18 @@ private function changedConcepts($request, $prop, $offset, $limit)
         $simpleChangeList = array();
         foreach($changeList as $conceptInfo) {
             if (array_key_exists('date', $conceptInfo)) {
-                $simpleChangeList[] =  array( 'uri' => $conceptInfo['uri'],
-                                               'prefLabel' => $conceptInfo['prefLabel'],
-                                               'date' => $conceptInfo['date']->format("Y-m-d\TH:i:sO") );
+                if (array_key_exists('replacedBy', $conceptInfo)) {
```
kouralex, line 1107: Duplicate code

model/sparql/GenericSparql.php
```diff
@@ -119,7 +119,6 @@ protected function generateFromClause($vocabs=null) {
         }
         return $clause;
     }
-
```
kouralex, line 122: Please remove

model/sparql/GenericSparql.php
```diff
@@ -2229,19 +2228,28 @@ public function listConceptGroupContents($groupClass, $group, $lang,$showDepreca
      * @param string $lang language of labels to return
      * @param int $offset offset of results to retrieve; 0 for beginning of list
      * @param int $limit maximum number of results to return
+     * @param boolean $showDeprecated whether to include deprecated concepts in the change list
      * @return string sparql query
      */
-    private function generateChangeListQuery($prop, $lang, $offset, $limit=200) {
+    private function generateChangeListQuery($prop, $lang, $offset, $limit=200, $showDeprecated) {
```
kouralex, line 2234: Mandatory parameter in signature after optional.

model/sparql/GenericSparql.php
```diff
@@ -2229,19 +2228,28 @@ public function listConceptGroupContents($groupClass, $group, $lang,$showDepreca
...
         $query = <<<EOQ
-SELECT DISTINCT ?concept ?date ?label $fcl
+SELECT DISTINCT ?concept ?date ?label $optionVar $fcl
 WHERE {
   ?concept a skos:Concept .
-  ?concept $prop ?date .
   ?concept skos:prefLabel ?label .
   FILTER (langMatches(lang(?label), '$lang'))
+  $optionStart ?concept $prop ?date .
+  MINUS{ ?concept owl:deprecated TRUE } $optionEnd
```
kouralex, line 2251: Please use `True` instead of `TRUE`.

model/sparql/GenericSparql.php
```diff
@@ -2229,19 +2228,28 @@ public function listConceptGroupContents($groupClass, $group, $lang,$showDepreca
...
-    private function generateChangeListQuery($prop, $lang, $offset, $limit=200) {
+    private function generateChangeListQuery($prop, $lang, $offset, $limit=200, $showDeprecated) {
         $fcl = $this->generateFromClause();
         $offset = ($offset) ? 'OFFSET ' . $offset : '';
 
+        //Additional clauses when deprecated concepts need to be included in the results
+        $optionVar = ($showDeprecated) ? '?replacedBy' : '';
+        $optionStart = ($showDeprecated) ? 'OPTIONAL {' : '';
+        $optionEnd = ($showDeprecated) ? '}' : '';
+        $deprecatedOption = ($showDeprecated) ? 'OPTIONAL { ?concept dc:modified ?date ; owl:deprecated TRUE ; dc:isReplacedBy ?replacedBy }' : '';
```
kouralex, line 2242: `?concept dc:modified ?date` and `?concept $prop ?date` both bind `?date`, seems incorrect?

kouralex: Also, `True` instead of `TRUE`

model/sparql/GenericSparql.php
```diff
@@ -2229,19 +2228,28 @@ public function listConceptGroupContents($groupClass, $group, $lang,$showDepreca
...
+
         $query = <<<EOQ
-SELECT DISTINCT ?concept ?date ?label $fcl
+SELECT DISTINCT ?concept ?date ?label $optionVar $fcl
```
kouralex, line 2245: I think you could just use ?replacedBy, which would yield null values if not bound?

model/sparql/GenericSparql.php
kouralex, line 2251: OPTIONAL + MINUS some of the very same matches does not seem right and may be the cause of performance regressions, especially combined with another OPTIONAL (OPTIONALs are not performance-friendly, at least not on Fuseki based on my own research). Please consider alternative approach, e.g., inner UNIONs can, in many cases, be applied instead and may even simplify the code. It should be an easy task for the triple store to retrieve the labels when the query is optimally constructed.

(Another performance-wise point regarding the whole code base: we should change from using the FROM clause into GRAPH based filtering.)

view/changes.twig
```diff
@@ -28,7 +28,11 @@
       {% for day,concepts in changeList %}
         <h5 class="versal versal-bold">{{day}}</h5>
         {% for concept in concepts %}
+          {% if concept.replacedBy is defined %}
+          <li><a href="{{ concept.uri | link_url(vocab,request.lang,'page', request.contentLang) }}"><del>{{ concept.prefLabel }}</del></a></li>
```
kouralex, line 32: I believe the `<s>` tag would be more appropriate here?

kouralex: Duplicate code, too.

One can either inline it, or even make a Twig filter or function.

tests/WebControllerTest.php
```diff
@@ -224,7 +224,7 @@ public function testFormatChangeList() {
         $months =$this->webController->formatChangeList($changeList, 'en');
 
         $expected = array ('hurr durr' => array ('uri' => 'http://www.skosmos.skos/changes/d3', 'prefLabel' => 'Hurr Durr', 'date' => DateTime::__set_state(array('date' => '2010-02-12 10:26:39.000000', 'timezone_type' => 3, 'timezone' => 'UTC')), 'datestring' => 'Feb 12, 2010'), 'second date' => array ('uri' => 'http://www.skosmos.skos/changes/d2', 'prefLabel' => 'Second date', 'date' => DateTime::__set_state(array('date' => '2010-02-12 15:26:39.000000', 'timezone_type' => 3, 'timezone' => 'UTC')), 'datestring' => 'Feb 12, 2010'));
-        $this->assertEquals(array('December 2011', 'February 2010', 'January 2000'), array_keys($months));
```
kouralex, line 227: ?

tests/GenericSparqlTest.php
```diff
@@ -1121,6 +1142,7 @@ public function testQueryChangeList()
     $this->assertEquals(array('Fourth date', 'Hurr Durr', 'Second date', 'A date'), $order);
   }
 
+
```
kouralex, line 1145: Please remove

model/Vocabulary.php
```diff
@@ -623,7 +623,8 @@ public function verifyVocabularyLanguage($lang)
      */
     public function getChangeList($prop, $clang, $offset, $limit)
     {
-      return $this->getSparql()->queryChangeList($prop, $clang, $offset, $limit);
+        $showDeprecated = $this->getConfig()->getShowDeprecatedChanges();
+        return $this->getSparql()->queryChangeList($prop, $clang, $offset, $limit, $showDeprecated);
```
kouralex, line 627: Now `getChangeList` will pass $showdeprecated on without it being able to be overriden in the signature. May be OK, but should give a thought for it before making the decision.

view/changes.twig
```diff
@@ -28,7 +28,11 @@
       {% for day,concepts in changeList %}
```
kouralex, line 28: Seems like GitHub is preventing me from commenting the appropriate spot, but one should check `trans` strings for the relevant `sr-only` parts as well as the relevant strings on the sidebar tab that are conditionally affected by the introduced functionality.

joelit:
For this to work as intended, finto.css needs to be modified by removing line 480 for .nav-tabs > li.active > a :
`  height: 38px;`

model/sparql/GenericSparql.php
```diff
@@ -2243,22 +2242,31 @@ public function listConceptGroupContents($groupClass, $group, $lang,$showDepreca
...
      * @return string sparql query
      */
-    private function generateChangeListQuery($prop, $lang, $offset, $limit=200) {
+    private function generateChangeListQuery($prop, $lang, $offset, $limit=200, $showDeprecated=false) {
         $fcl = $this->generateFromClause();
         $offset = ($offset) ? 'OFFSET ' . $offset : '';
 
+        //Additional clauses when deprecated concepts need to be included in the results
+        $deprecatedVars = ($showDeprecated) ? '?replacedBy ?deprecated ?replacingLabel' : '';
+        $deprecatedOptions = ($showDeprecated) ? 'UNION { ?concept dc:isReplacedBy ?replacedBy ; dc:modified ?date2 . BIND(COALESCE(?date2, ?date) AS ?date) OPTIONAL { ?replacedBy skos:prefLabel ?replacingLabel . FILTER (langMatches(lang(?replacingLabel), \''.$lang.'\')) }} OPTIONAL { ?concept owl:deprecated ?deprecated . }' : '';
```
kouralex, line 2254: This is (by far) the longest line in the file. Worth considering: split? (increases readability)

model/sparql/GenericSparql.php
```diff
@@ -2243,22 +2242,31 @@ public function listConceptGroupContents($groupClass, $group, $lang,$showDepreca
...
+            FILTER (langMatches(lang(?label), '$lang'))
+            {
+                ?concept $prop ?date .
+                MINUS { ?concept owl:deprecated True . }
+            }
+            $deprecatedOptions
+        }
+        ORDER BY DESC(YEAR(?date)) DESC(MONTH(?date)) LCASE(?label) DESC(?concept)
+        LIMIT $limit $offset
 EOQ;
```
kouralex, line 2270: This changes the indentation of the SPARQL query which has two effects:

1) Wherever it is logged/printed out, this will have extra whitespace in the beginning
2) It complete rewrites the query, thus, git diff here does not show immediately what has changed

Please consider reverting the indentation.

model/sparql/GenericSparql.php
```diff
@@ -2243,22 +2242,31 @@ public function listConceptGroupContents($groupClass, $group, $lang,$showDepreca
...
-  ?concept skos:prefLabel ?label .
-  FILTER (langMatches(lang(?label), '$lang'))
-}
-ORDER BY DESC(YEAR(?date)) DESC(MONTH(?date)) LCASE(?label)
-LIMIT $limit $offset
+        SELECT ?concept ?date ?label $deprecatedVars $fcl
+        WHERE {
+            ?concept a skos:Concept ;
+	        skos:prefLabel ?label .
+            FILTER (langMatches(lang(?label), '$lang'))
```
kouralex, line 2261: Even though it was already like this, this will filter out all concepts without a content language label, thus, resulting in incomplete listing. If one wants to fix it, one must use the fallback method found elsewhere in this file.

kouralex: Side note: in theory, this affects also replacedByLabel.

model/sparql/GenericSparql.php
```diff
@@ -2243,22 +2242,31 @@ public function listConceptGroupContents($groupClass, $group, $lang,$showDepreca
...
+            {
+                ?concept $prop ?date .
+                MINUS { ?concept owl:deprecated True . }
```
kouralex, line 2264: Shouldn't this be set only if `$showDeprecated` equals `false`?

tests/GenericSparqlTest.php
```diff
@@ -1121,12 +1121,33 @@ public function testListConceptGroupContentsIncludingDeprecatedConcept()
    * @covers GenericSparql::generateChangeListQuery
    * @covers GenericSparql::transFormChangeListResults
    */
-  public function testQueryChangeList()
+  public function testQueryChangeListCreatedNoDeprecated()
   {
     $voc = $this->model->getVocabulary('changes');
     $graph = $voc->getGraph();
+    $sparql = new GenericSparql('http://localhost:13030/skosmos-test/sparql', $graph, $this->model);
```
kouralex, line 1128: SPARQL endpoint address should be set via a variable, see #1166 for more information.

resource/css/styles.css
```diff
@@ -1553,10 +1553,21 @@ tr.replaced-by > td > ul > li > a {
 .nav-tabs {
   background-color: #e0e4e7;
   border-bottom: none;
+   display: flex;
```
kouralex, line 1556: Very small indentation issue 🙂

model/VocabularyConfig.php
```diff
@@ -440,7 +440,7 @@ public function getDefaultSidebarView()
         $defview = $this->resource->getLiteral('skosmos:defaultSidebarView');
         if ($defview) {
             $value = $defview->getValue();
-            if ($value === 'groups' || $value === 'hierarchy') {
+            if ($value === 'groups' || $value === 'hierarchy' || $value == 'new') {
```
kouralex, line 443: Please use "==="

PR #1166, opened 2021-05-25
Specify SPARQL endpoint URL via environment variable

osma:
I was trying to set up GitHub Actions CI (#1163, #1147) for running the tests but hit problems caused by hardcoded SPARQL endpoint URLs in the test suite (PHPUnit classes & Skosmos configuration files). Under GitHub Actions it would be easier to use a Fuseki running in a Docker container.

This PR makes the global `skosmos:sparqlEndpoint` and vocabulary-specific `void:sparqlEndpoint` settings default to the value of the `SKOSMOS_SPARQL_ENDPOINT` environment variable when no value is set in `config.ttl`.

The test suite makes use of this by relying on the environment variable instead of hardcoding the endpoint URL. The default (set in `tests/bootstrap.php`) is still to use Fuseki running on localhost:13030 if the environment variable hasn't been set, so the existing way of running unit tests on a developer machine (start Fuseki using init_fuseki.sh and leave it running in the background, then run vendor/bin/phpunit) should still work.

### DIFF
--- a/model/Vocabulary.php
+++ b/model/Vocabulary.php
@@ -31,14 +31,7 @@ class Vocabulary extends DataObject implements Modifiable
      */
     public function getEndpoint()
     {
-        $endpoint = $this->resource->get('void:sparqlEndpoint');
-        if ($endpoint) {
-            return $endpoint->getUri();
-        } elseif (getenv('SKOSMOS_SPARQL_ENDPOINT')) {
-            return getenv('SKOSMOS_SPARQL_ENDPOINT');
-        } else {
-            return 'http://localhost:3030/ds/sparql';
-        }
+        return $this->config->getSparqlEndpoint();
     }
 
     /**
@@ -48,12 +41,7 @@ class Vocabulary extends DataObject implements Modifiable
      */
     public function getGraph()
     {
-        $graph = $this->resource->get('skosmos:sparqlGraph');
-        if ($graph) {
-            $graph = $graph->getUri();
-        }
-
-        return $graph;
+        return $this->config->getSparqlGraph();
     }
 
     /**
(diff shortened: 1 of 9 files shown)


kinow (approved):
Still draft, but looking good and CI passing too, and I think it will solve the issue from GH actions, so +1 :grimacing:

osma:
Currently testing this on GitHub Actions. I will finalize this once I know this works with Actions...currently I'm not yet sure

osma:
It turned out that this wasn't necessary for GitHub Actions anyway, but configuration via environment variables still a nice feature and may be useful later. Probably other aspects should be configurable via environment variables as well, but that's for another PR or two...

After this PR, it's possible to

1. Omit the global SPARQL endpoint setting `skosmos:sparqlEndpoint`; the value of the `SKOSMOS_SPARQL_ENDPOINT` environment variable will be used if set, otherwise the default is `http://localhost:3030/ds/sparql` (typical for a local Fuseki)
2. Omit the vocabulary-specific SPARQL endpoint setting `skosmos:sparqlEndpoint`; if not set, the global endpoint setting (see above) will be used.

I did a little refactoring of the code, moving methods for reading the SPARQL endpoint configuration from the Vocabulary class to the VocabularyConfig class where they more properly belong.

model/Vocabulary.php
```diff
@@ -58,8 +57,8 @@ public function getSparql()
     {
         $endpoint = $this->getEndpoint();
         $graph = $this->getGraph();
-        $dialect = $this->resource->get('skosmos:sparqlDialect');
-        $dialect = $dialect ? $dialect->getValue() : $this->model->getConfig()->getDefaultSparqlDialect();
+        $dialect = $this->config->getSparqlDialect();
+        $dialect = $dialect ? $dialect : $this->model->getConfig()->getDefaultSparqlDialect();
```
osma, line 61: Use ?? operator

PR #1168, opened 2021-05-26
Switch to GitHub Actions CI

### LINKED ISSUE #1147: Switch from Travis CI to GitHub Actions

osma:
This PR switches to GitHub Actions CI instead of Travis CI. This is a rework of PR #1163, which got quite messy.

Fixes #1147 

TODO items:

- [x] Use the same PHPUnit version as declared in composer.json
- [x] Cache the Composer dependencies
- [x] Install Fuseki and load the test vocabularies into it
- [x] Cache the Fuseki installation
- [x] Run the tests on all supported PHP versions in parallel
- [x] Report code coverage to Code Climate
- [x] Report code coverage to Codecov
- [ ] Report passing/failing tests to Slack

There is a [networking issue](https://github.com/php-actions/phpunit/issues/28) in the php-actions/phpunit Action that this workflow relies on. This has been fixed in my fork but my [PR for fixing it](https://github.com/php-actions/phpunit/pull/29) hasn't yet been accepted upstream. So for now this CI configuration is relying on the forked version.

This is now using the PCOV PHP extension to collect coverage information, instead of phpdbg that was used on Travis CI. PCOV is both faster and more accurate so this is clearly a win. Also, the Fuseki version has been upgraded to 4.0.0.

A few times the Fuseki installation step has been stuck. I don't know what is causing it but [others](https://github.community/t/step-hangs-indefinitely/17720) [have](https://github.community/t/github-actions-stuck-on-step-and-timeout/137170) [reported](https://github.com/actions/runner/issues/571) similar problems so it could be a more general problem with Actions. If this happens the job can be restarted (though canceling also seems to get stuck).


.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
+name: CI
+
+on: [push]
+
+jobs:
+  build-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php_version: [7.2, 7.3, 7.4]
```
osma, line 10: Could try PHP 8 as well (not critical if it fails)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+
+    - name: Start up Fuseki
+      run: cd tests; sh ./init_fuseki.sh
+    
+    - name: Cache Composer dependencies
+      uses: actions/cache@v2
+      with:
+        path: |
+          /tmp/composer-cache
+          vendor
```
osma, line 31: Add components directory to the cache

PR #1174, opened 2021-05-31
Issue913 deprecated concepts number

### LINKED ISSUE #913: Exclusion/expression of deprecated concepts in ontology metadata
The numbers of concepts and terms in ontology metadata (Resource counts by type, Term counts by language) currently include deprecated concepts. They should be excluded from these numbers as they are not browsable. See: http://finto.fi/yso/en/

Is it possible to express the number of deprecated concepts as a separate figure in ontology metadata? 

More general question: Should the number of deprecated concepts be expressed as part of ontology metadata even if possible? As said, users cannot access these concepts through browsing. Are there standards in the field for deprecated/removed concepts in this respect?

Vainonen:
Fixes #913 by adding number of deprecated concepts into vocabulary info page under "Resource counts by type". Number of deprecated concepts by concept subtypes and number of deprecated collections is added to REST API responses.


osma (approved):
LGTM!

PR #1175, opened 2021-06-01
Replacing the malihu-custom-scrollbar plugin with some CSS

henriyli:
Doesn't look 100% the same but helps to get rid of one unmaintained plugin that's semi-useless.


osma (changes requested):
I tested this briefly and also looked at the code. Here is an illustration of how this affects the YSO front page:

image

Mostly this looks good. A few notes:

1. The waypoint callbacks are working on the vocabulary home page and the alphabetical index (new items are loaded when you scroll down enough). But it doesn't work on the New tab that shows new concepts. You can scroll all the way down the initial list but nothing happens. **This must be fixed.**
2. The scrollbar is a bit narrower than the old one, but that's OK. It's also quite a bit higher. I think this **should be adjusted closer to the original height**. The colors are the same as in the default Skosmos look and feel, which is OK; the above screenshot shows the Finto style, where the custom CSS should be adjusted to fix the colors (but we can do that after merging this).
3. The commits contained quite a bit of whitespace-only changes which make the reviewing of changes a bit harder than necessary. But I think we can live with those, as this is simple enough. But for future PRs it would be better either not to reformat existing code, or if you have to, then do it in a separate commit (or even a PR) so that it's possible to ignore when reviewing.

I have no idea why we were using a static copy of `resource/css/jquery.mCustomScrollbar.css` instead of relying on the one under the `vendor/` directory managed by Composer, but I'm glad we're getting rid of it :)

henriyli:
1) The new tab callback seems to be working for me. What do I need to do to reproduce the issue?

2) I adjusted it to be a bit wider and added a box shadow to the thumb to emulate shorter height (the thumb height cannot be made shorter otherwise). How does it look now?

3) Seems like I had auto formatting on save on in my editor but I've turned it off for now.

No idea why the CSS was a static import. Probably some brain fart from way back. :)

osma:
> The new tab callback seems to be working for me. What do I need to do to reproduce the issue?

I tested this again using both Firefox 88 and Chromium on Ubuntu 20.04. I see two issues with the callback on the New tab on both browsers:

1. If you open the A-Z tab, then click on the New tab, the UI performs a partial page load using AJAX. When you then scroll down the list of results, the callback isn't triggered at all. ISTR that the callback needs to be installed separately after the AJAX page update; probably this hasn't been done properly.
2. If you go to the New tab directly via URL and scroll down, the callback is triggered not just once, but many times; and eventually you get a list of results which may be much older than expected. See the Network tab in this screenshot:

image

Also, after the requests have completed, there is a gap in the list of new concepts (here, "compression garment" should have been followed by other concepts from November 2020, e.g. "compression treatment", not concepts from April 2018):

image

> I adjusted it to be a bit wider and added a box shadow to the thumb to emulate shorter height (the thumb height cannot be made shorter otherwise). How does it look now?

I didn't have time to look at this in very much detail but it seems to me that the height of the new scrollbar is still very different from the old one. It is also much higher on Firefox than Chromium. I'll look at this again once the callback issues have been resolved.

PR #1178, opened 2021-06-12
Update Dutch language strings

redmer:
Hi y’all 👋

I’ve improved upon the Dutch language translation. I hope this is in the correct way, as I begun from the `.po` file and not from Transifex.

- correct spelling in places (by removing spaces in compounds)
- use same terms more consistently (_concept_, _subvocabulair_, _woordenlijst_, imperative for actions, etc.)
- adds untranslated strings

This PR includes a compiled `.mo` file.


osma:
Thanks for the PR @redmer ! However, we use Transifex as the master data source for translations (except fi, sv, en). So these changes should be done in Transifex, not directly in a PR; otherwise, they will be overwritten the next time we pull translations from Transifex. Do you think you could do the edits there instead?

redmer:
No problem, I updated the translations over at Transifex!

redmer:
Because I also spreak Frysk (`fy`) that project was somehow requested to the project, so I added a full translation of it as well.

pedropaulofb:
@redmer , I received some suggestions from dutch colleagues regarding the dutch translation. Can you please evaluate and consider them?

breder concept/concepten -> algemener begrip
nauwere concept/concepten -> specifieker begrip
is concept van -> referentie
download dit concept -> download dit begrip

PR #1182, opened 2021-06-27
Update to Bootstrap 5

### LINKED ISSUE #1045: Upgrade UI to use Bootstrap 5
The current Skosmos UI uses Bootstrap CSS version 3.

We should upgrade to Bootstrap 4, which among other things uses flexbox for layout. This should make it easy to fix alignment issues such as those described in issue #1006.

Upgrading Bootstrap is a large undertaking as all the HTML templates and CSS files are affected.

Bootstrap 4 drops support from some older browsers. We should verify that this doesn't cause problems for users of important Skosmos installations such as Finto.

kinow:
Closes #1045 

- [x] ported Bootstrap 3 to Bootstrap 5
- [x] updated dependencies fixing the stylesheet and JS accordingly
- [x] added CSS variables
- [x] tested
- [ ] addressed feedback & fixed issues (testing again)
- [ ] wrote a changelog suggested entry
- [ ] tested autocomplete (https://github.com/NatLibFi/Skosmos/issues/1246)


composer.json
```diff
@@ -62,8 +62,8 @@
     "etdsolutions/waypoints": "4.0.0",
     "twig/twig": "2.13.*",
     "twig/extensions": "1.5.*",
-    "twitter/bootstrap": "3.3.*",
-    "twitter/typeahead.js": "v0.10.5",
+    "twbs/bootstrap": "5.0.*",
+    "twitter/typeahead.js": "v0.11.*",
```
kinow, line 66: Note to self: before making it a final PR, confirm whether there's a new version of twbs/bootstrap and update here if necessary.

dockerfiles/Dockerfile.ubuntu
```diff
@@ -67,13 +67,13 @@ RUN echo "ServerName localhost" >> /etc/apache2/apache2.conf
...
-RUN curl -sS https://getcomposer.org/installer | php
-RUN php composer.phar install --no-dev
+# COPY . /var/www/html
+# RUN curl -sS https://getcomposer.org/installer | php
+# RUN php composer.phar install --no-dev
 
 
 # Configure Skosmos
-COPY dockerfiles/config/config-docker.ttl /var/www/html/config.ttl
+# COPY dockerfiles/config/config-docker.ttl /var/www/html/config.ttl
```
kinow, line 76: Note to self: this is to speed up development, mapping the Docker volume. Undo all the changes to the Dockerfiles when the PR is ready for review.

view/concept-shared.twig
```diff
@@ -75,9 +75,8 @@
         {% endif %}
         {% endif %}
         &nbsp;
-        <button type="button" data-toggle="tooltip" data-placement="button" title="Copy to clipboard" class="btn btn-default btn-xs copy-clipboard"
-          for="#{% if concept.notation %}notation{% else %}pref-label{% endif %}">
-          <span class="glyphicon glyphicon-copy" aria-hidden="true"></span>
+        <button type="button" data-bs-toggle="tooltip" data-bs-placement="button" title="Copy to clipboard" class="btn btn-default btn-xs copy-clipboard" for="#{% if concept.notation %}notation{% else %}pref-label{% endif %}">
+            <svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 0 24 24" width="24px" fill="#00748f"><path d="M0 0h24v24H0z" fill="none"/><path d="M16 1H4c-1.1 0-2 .9-2 2v14h2V3h12V1zm3 4H8c-1.1 0-2 .9-2 2v14c0 1.1.9 2 2 2h11c1.1 0 2-.9 2-2V7c0-1.1-.9-2-2-2zm0 16H8V7h11v14z"/></svg>
```
kinow, line 79: Glyphicons is not free anymore, and Bootstrap dropped it in Bootstrap 4. I've used this SVG from Google Material Icons, but we can use another one instead too :+1:

kinow:
@osma I believe this PR is ready for an initial round of reviews. Here's how I tested:

```
# one terminal for:
cd skosmos/dockerfiles
docker-compose rm -f && docker-compose up --force-recreate

# another terminal with files for UNESCO and STW vocabs
curl': curl -I -X POST -H Content-Type:text/turtle -T unescothes.ttl -G http://localhost:9030/skosmos/data --data-urlencode graph=http://skos.um.es/unescothes/ && ^Crl -I -X POST -H Content-Type:text/turtle -T stw.ttl -G http://localhost:9030/skosmos/data --data-urlencode graph=http://zbw.eu/stw/
```

You'll notice I've changed the `dockerfiles` contents in this PR, but that's only for this review. I'll remove/edit commits later as that's only useful for development. I'm basically binding a volume so I can quickly develop the changes in the container and watch the changes in the browser.

In one monitor then I have http://localhost:9090 with Skosmos + Bootstrap v5, and in another monitor I have skosmos.dev.finto.fi/en/ with Skosmos + Bootstrap v3.

But this is not the only way to review. If you have a working PHP installation and you are more comfortable using it, feel free to check out this branch and try it. The docker changes should not interfere with your review :+1: 

Thanks!
Bruno

osma:
Thanks a lot @kinow , I will review this and get back to you soon!

osma (changes requested):
I tested this locally (no docker involved). I was surprised about how smooth the upgrade seems to be, even though this is skipping directly from Bootstrap 3 to 5!

That said, I found some issues. I used Firefox 90 on Ubuntu 20.04 for testing.

Functional:

1. (This could be related to the Typeahead upgrade you did in this PR.) When inside a vocabulary, the search box is doing an unrestricted global search instead of searching within the vocabulary and the chosen language. This means the search is slow and returns results also from other vocabularies, e.g. 
image
I checked the AJAX query and it looks like this: `rest/v1/search?query=aardva*` while in the old version it looks like this: `rest/v1/search?query=aardv*&vocab=yso&lang=en&labellang=en` - so clearly the URL parameters are not being set correctly. The same happens also on the front page where the search is (by default) global; the query is just `rest/v1/search?query=aardva*` when it should be e.g. `rest/v1/search?query=aardv*&vocab=&lang=en&labellang=en`

Display-related:

1. The Fira font is not being applied to many elements. Here are some examples where I'm seeing another font (Arial?) instead:

image
image
image
image
image
These were just some obvious ones, there could be more.

2. I tested how the layout works on narrower devices using Responsive Design mode. There seems to be a lot of empty space on e.g. the vocabulary information page:

image

3. In the search bar, the font size has increased a little bit.

4. The copy to clipboard icon - I understand the issue with Glyphicons. I think we may want to choose a different icon than the one you selected though. I will get back about this. Also, I think it would be better to avoid embedding the SVG icon in the HTML code. The icon is already used in a couple of places and there could be more in e.g. Skosmos plugins, so either a CSS rule (like glyphicons) or a separate image file would be better.

Keep up the good work!

kinow:
Hi @osma ! Thanks for the excellent feedback (as always!). I was looking only at CSS changes and didn't see the URL query parameters change, nice catch! Will go through your feedback in the next days and update this PR.

kinow:
>(This could be related to the Typeahead upgrade you did in this PR.) When inside a vocabulary, the search box is doing an unrestricted global search instead of searching within the vocabulary and the chosen language. This means the search is slow and returns results also from other vocabularies

Fixed. You were correct @osma , Typeahead changed its syntax, so I had to read their [docs](https://github.com/twitter/typeahead.js/blob/master/doc/bloodhound.md#remote) and update the code. I beleive now the result should be the same as before :+1:

kinow:
> The Fira font is not being applied to many elements. Here are some examples where I'm seeing another font (Arial?) instead:

Bootstrap 5 has a CSS variable that is setting the font to the OS default sans-serif font. There's also a reset.css or reboot.css I think, that resets other font attributes. The issue is that in Bootstrap 3 we didn't have that, so now Bootstrap 5 is trying to be helpful by resetting font attributes, but overriding some of what we had defined.

I've added the `font-family: inherit` and then it got the font-family from that CSS variable. Also had set the font-family for h2 and h3 only, now p, h_ are all using Fira. But will have to have another look later to confirm I didn't miss any HTML elements.

kinow:
Three more to go! Will have to stop for today, but will pick it up in the next day(s).

kinow:
>In the search bar, the font size has increased a little bit.

Fixed setting the document root font size. Usually I try to avoid touching the root properties, but Bootstrap is doing the same, but now with a bigger font. We don't have SASS or SCSS in our build toolchain, otherwise I could alter Bootstrap's variable. For now I think this will do (later if the UI build changes it should be simpler to update it).

kinow:
> The copy to clipboard icon - I understand the issue with Glyphicons. I think we may want to choose a different icon than the one you selected though. I will get back about this. 

:+1: 

> Also, I think it would be better to avoid embedding the SVG icon in the HTML code. The icon is already used in a couple of places and there could be more in e.g. Skosmos plugins, so either a CSS rule (like glyphicons) or a separate image file would be better.

Good idea. Moved it to an external file, so we can just swap the file and it should be updated.

kinow:
>    I tested how the layout works on narrower devices using Responsive Design mode. There seems to be a lot of empty space on e.g. the vocabulary information page:

Hmm, you are right. I was using Firefox `CTRL+M` to test the responsive mode, but with a very small viewport. Looks like the medium viewport breaks the UI. Bootstrap 4 & 5 have a `container-fluid` exactly for cases like that, so I switched our `.content` by `.content-fluid`, and added a few more rules to try to match the Bootstrap 3 rules, keeping as close as possible to the old layout.

I think I addressed your initial feedback @osma. Let me know if there are other items pending, otherwise I will have another look at the differences one day of this week (so I stop comparing it for a while; will choose a day when :brain: is happy and fresh, so I'm able to spot more differences).

Cheers
Bruno

osma:
Thanks for the fixes! I did another round of testing, using Firefox 91 on Ubuntu. This time the issues are more minor, in many cases just small adjustments of a few pixels. Mind you, not everything has to be pixel-perfect in the upgrade - I think it's OK to change things in the layout as long as they don't look bad or cause usability or accessibility issues.

1. The hover color on the sidebar tab looks wrong (I couldn't get the mouse cursor included in the screenshot):
image

2. In the sidebar, the tabs are a couple of pixels higher than they used to be. Also, the alphabetical letters are spaced further apart. In this screenshot (old on the left, new on the right) you can see that L is the first letter in the second row, when it used to be the last one in the first row:
image

3. The search bar is a few pixels wider than it used to be, but the search box itself is actually a bit narrower. The down arrow in the content language selection is further away from the language name and higher up. The X in the search box is lower down. In the screenshot, the old is on top, new on the bottom:
image

4. On the vocabulary front page (but not other page types), the left margin on the content area is narrower than it used to be. Also the gap between the columns is narrower and the column widths are different. I think it would be important to maintain the margins and the width of the gap; the exact column widths are not so important as they are affected by the content anyway.
image

5. In the content area of the concept page, there is more vertical space between lines than before, especially just above and below PREFERRED TERM but also between later rows. For example the screenshots below cover the exact same elements, but the new one (on the right) is 30 pixels higher:
image

6. The search suggestion box that pops up is a couple pixels too wide on the left side:
image

7. The feedback page is missing the white background it used to have:
image

8. The download links on the vocabulary front page are not shown using the Fira Sans font - only Fira should be used in Skosmos. This is actually a bug in the current version as well, so not your fault! ;) But maybe we could fix this in this Bootstrap upgrade.
image

Phew, that's what I found for now. Keep up the good work!

kinow:
Woah! That's an impressive review  @osma !! I'll try to update the branch again fixing these bugs. The ones where items are not aligned really bug me, but I missed them sorry. At least I think we are getting close to a working Skosmos+Bootstrap5, and this PR appears to be going in the right direction :tada: :man_dancing: thanks!!!

kouralex:
To start, I just want to say that this is an awesome WIP @kinow and I just wanted to give it a try!

Some remarks not yet covered by @osma :

* Autocomplete is currently limited to 5 results only
* There are couple margin/lining/missing information (type) issues in the dropdown, see image below:
image
*Aligning, font size and background color on front page:
image
*Alphabetical index does not get initialized on vocabulary front page load

Keep up the good work! :)

kinow:
Thanks for the review @kouralex ! The more the merrier. Noted the issues found, will start fixing them slowly soon :nerd_face:

kinow:
@osma, thank you for the feedback.

>1. The hover color on the sidebar tab looks wrong (I couldn't get the mouse cursor included in the screenshot):

Good spot! Bootstrap 5 defines the link hover color for navigation links. Had to override that one.

>2. In the sidebar, the tabs are a couple of pixels higher than they used to be. Also, the alphabetical letters are spaced further apart. In this screenshot (old on the left, new on the right) you can see that L is the first letter in the second row, when it used to be the last one in the first row:

This was the hardest to tackle. I had noticed this regression, but forgot to point it out, sorry. Basically, Bootstrap 5 uses flex elements to create the pagination. Browsers implement grid & flex nowadays, and it's a lot easier to work with these than trying to align items ourselves (main justifications for moving to grid or flex).

image

Looks like the flex model adds some kind of virtual border, as the width of the elements was the same in old & new. I reduced the width of the element from 26px to 24px, and now the pagination layout is looking a bit closer to the old design. WDYT @osma?

>3. The search bar is a few pixels wider than it used to be, but the search box itself is actually a bit narrower. The down arrow in the content language selection is further away from the language name and higher up. The X in the search box is lower down. In the screenshot, the old is on top, new on the bottom:

Tried to match the previous layout. It's tricky because of the structure changes from Bootstrap 3 to 5. But I think I got the elements re-aligned. Not identical, but the down-arrow should look better aligned. I spent some time trying to understand why I was getting different spacing for the two down-arrows/carets in the screen. Turns out the one in the top had the text "English", and the one next to the search field had "English " with an extra space.

>4. On the vocabulary front page (but not other page types), the left margin on the content area is narrower than it used to be. Also the gap between the columns is narrower and the column widths are different. I think it would be important to maintain the margins and the width of the gap; the exact column widths are not so important as they are affected by the content anyway.

Ah, I hadn't noticed it! I caused this bug when using `container-fluid` which has its own gutter values (padding/margin). I've reverted to `.container` only, but added `width: 100%` and `padding: 0 15px` to match the `concept-info`.

>5. In the content area of the concept page, there is more vertical space between lines than before, especially just above and below PREFERRED TERM but also between later rows. For example the screenshots below cover the exact same elements, but the new one (on the right) is 30 pixels higher:

I think I got it fixed now. The clipboard icon is getting in the way of the row element. I've changed it a bit, but I think we will revisit it later when we decide which icon to use as replacement for glyphicon (now paid).

>6. The search suggestion box that pops up is a couple pixels too wide on the left side:

I couldn't find what change caused that, so my best guess is that typeahead changed something in its CSS and caused that? I reduced the width of the suggestion box in 2 pixels, and then moved the box 1 pixel to the right. The border has 1 pixel to each side (top, right, bottom, left), so that should position the suggestion box exactly below with the border and no content overflowing the search input.

>7. The feedback page is missing the white background it used to have:

Fixed by one of the changes above. Couldn't reproduce it on Chromium or Firefox. Could you test it again, please @osma?

>8. The download links on the vocabulary front page are not shown using the Fira Sans font - only Fira should be used in Skosmos. This is actually a bug in the current version as well, so not your fault! ;) But maybe we could fix this in this Bootstrap upgrade.

Oh, interesting. @osma, I've applied the Fira font to the `html` and `body` elements now, as well as the font-size `14px`. So it should apply to all elements by default. I think that's the intention? To use that font everywhere? If so, it should fix all font-family issues.

Going to work on @kouralex feedback now. Thanks!

kinow:
Hi @kouralex,

>Autocomplete is currently limited to 5 results only

Turns out there's a limit now in Typeahead, that defaults to 5. I've added a variable in the `config.js` file set to 100. @osma, @kouralex, should we increase it to something like 1000, or does 100 sound reasonable?

>There are couple margin/lining/missing information (type) issues in the dropdown, see image below:

I think @osma had provided a similar feedback. I changed the border/width of the tt-menu, and it looks aligned on chromium/ffox. Could you check if that looks better now? If not, could you share your environment settings, please? Then I will try to reproduce locally.

> *Aligning, font size and background color on front page:

I zoomed in and opened two browser windows with old & new skosmos to compare the dropdowns. I tried to match the layout & behaviour. And I must admit it was really different, thanks for pointing the differences @kouralex! Could you take another look and see if it's better, or at least going in the right direction?

> *Alphabetical index does not get initialized on vocabulary front page load

I just tested with YSO, STW, and UNESCO vocabularies, and the alphabetical index loaded correctly, I think. Could you confirm if that's still an issue, please, @kouralex?

Thanks a lot for the review!

kinow:
Ready for review again :tada:

osma:
Thanks @kinow, looking at this again...

I noted this above:
> Note to self: before making it a final PR, confirm whether there's a new version of twbs/bootstrap and update here if necessary.

It seems that 5.1.1 is now [available on Packagist](https://packagist.org/packages/twbs/bootstrap). Of course 5.0.x is already a huge improvement over the current version so updating to 5.1.1 could reasonably be left for later.

There's a little conflict in `docready.js` unfortunately, do you think it would be easy to fix @kinow?

I will review this once more Very Soon (tm) and get back if there are any further issues. I think it would be good to merge this pretty soon to `master` so it gets more testing and we can work out any issues that turn up after merging.

kinow:
Ah, good timing @osma, just found some spare time now- waiting for a webinar. I'll fix the conflict and update to 5.1.1.

> Very Soon (tm)

:laughing:

kinow:
SonarCloud security issue is related to `resource/js/scripts.js`, but that file was not changed in this PR? Fixing multiselect CSS regressions after updating to 1.1.1.

kinow:
Done @osma. Found a few glitches in the new CSS of multiselect. Overrode it with `! important` (in one case they were using it, had to match the CSS class exactly to override it). Did a quick test with some vocabularies, and everything seems to be in order. Feel free to review it any time :+1: 

> I think it would be good to merge this pretty soon to master so it gets more testing and we can work out any issues that turn up after merging.

+1, happy to keep the linked issue open for a few weeks/months (maybe until next sprint?), to gather feedback and fix any regressions.

Thanks!

kinow:
Rebased onto `master`, and also removed the Docker modifications (they were only to ease the upgrade). But in case I need to run this again and forget what I changed, here's a screenshot :sweat_smile: 

docker-skosmos

kinow:
I started rebasing this PR, but there were some changes in the layout that will take a little longer to fix the conflicts :cold_sweat: . There are some changes that won't cause conflicts, but still need to be addressed (e.g. some HTML elements were added somewhere that didn't cause a conflict, but using Bootstrap 3; git might merge it without conflicts, but then this branch would have a bug).

Will try to rebase and fix conflicts later when I have more time :+1:

osma:
Thanks for the update @kinow ! And sorry about the conflicts - these are inevitable when there is work going on in parallel on many branches.

I'm also preparing a set of feedback for this PR - @kouralex started collecting cases where the layout changes due to this PR, but then he had to switch to other tasks. I will post the notes soon, but I need to go through it all first.

osma:
Here are some notes about unintended layout changes. In the screenshots, old (current master) is on the left, new (this PR branch) on the right.

# Front page headings

The headings on the front page used to be bold, now they are not:

image

# UI language dropdown

When using the dropdown style UI language selection, the menu doesn't look quite right. When hovering the mouse over the current selection (here "English"), its background turns dark blue. When clicking, the menu opens too far down, and the background color of the current selection then switches to a wrong shade of gray.

image

# Highlights when using the keyboard to select language

The content language selection menu has different colors and highlights, especially when using the keyboard to navigate.

First selection by tabbing (thick white border is missing):

image

Opening the language menu and moving a couple items down the list using Enter + Down + Down (the text color changed from black to blue and the dotted selection indicator is missing, making it impossible to see which entry is selected; also, the background color of the menu element now stays light blue, but IMHO this is actually good):

image

# Highlights when choosing multiple vocabularies

Like languages above, some of the highlight effects are wrong or missing when choosing multiple vocabularies for searching. When first selecting the menu by tabbing, there is no highlight at all (used to be a thick white border):

image

Opening the menu and moving down the list by Enter + Down + Down (the dotted highlight is missing - and the menu itself changes back to gray; also, spacing between items has increased a bit):

image

When performing selections, there are also some differences. The blue checkmark background is a bit darker than it used to be, reducing contrast:

image

The tabbing behavior of the vocabulary selection is now also different. Before, if you had opened the menu using the keyboard, clicking Tab would close the menu. Now if you click on Tab while inside the menu, it will select each checkbox and item in turn (checkbox - item - checkbox - item) which is a bit strange. I don't know what would be the best solution for accessibility here, but at least it seems wrong to tab through both the checkboxes and the whole items in turns.

# Vocabulary front page

The separators between fields now look a bit funny. Not sure what happened here, because this didn't happen in earlier iterations.

image

# Vocabulary selection on feedback page

When selecting the target vocabulary on the feedback page, the selection list looks different. The gray background of the previously selected item is darker. The selection list opens lower down than it used to. The text color of items is now blue, when it used to be black. Items with long names are wrapped incorrectly.

image

Phew, that's it for now. I'm sure there are more details like this but at least here's something for you to work on ;) (in addition to the conflicts, rebases etc)

osma:
One more thing - the spinner shown when the concept page is loaded via AJAX is wrongly positioned:

image

kinow:
@osma I've finished rebasing it (after squashing commits). I will go through the issues you listed above in the next days. Thanks!!!

kinow:
Rebased fixing conflicts from https://github.com/NatLibFi/Skosmos/pull/1226.

I used the same variables created by @schlawiner, so users currently relying on those variables to customize their Skosmos installations shouldn't have a problem when upgrading.

However, there will be CSS class names that differ since Bootstrap structure/layout isn't exactly the same when comparing 3 with 5.

Finally, we have two extra variables in this PR, for `font-family` and `font-size` (didn't create different variables for font sizes, but that'd be a good follow-up improvement too).

Now will start rebuilding the environment so I can test the changes after the rebase, and address @osma 's feedback.

Bruno

kinow:
@osma, @kouralex,

Excellent feedback, thanks for testing it (again!).

>Front page headings
>The headings on the front page used to be bold, now they are not:

Ah! I looked at that landing page so many times comparing old and new, and just never noticed that font difference. Fixed!

>UI language dropdown
>When using the dropdown style UI language selection, the menu doesn't look quite right. When hovering the mouse over the current selection (here "English"), its background turns dark blue. When clicking, the menu opens too far down, and the background color of the current selection then switches to a wrong shade of gray.

I spotted that one yesterday while rebasing. Also thought odd and wondered if it was caused during my rebase. Looks like it was not, and instead I simply overlooked it. Fixed!

>Highlights when using the keyboard to select language
>First selection by tabbing (thick white border is missing):

Huh, I must have fixed this one yesterday? I just clicked on the search input box, then CTRL+TAB (select backward on my Linux) and the border was displayed alright.

image

>Highlights when choosing multiple vocabularies
>

This was the hardest part. The drop downs in Skosmos are not uniform, and had some style specific for them in the old styles.css, mixed with Bootstrap 3, which made it really hard to port it to Bootstrap 5. I ended up adding an HTML ID to the drop down button to select vocabularies, and applying the styles individually.

@osma, one thing that I couldn't find an easy solution is the tab order. At http://skosmos.dev.finto.fi/en/, when I expand the drop down to select the vocabularies, and press tab, it goes to the next drop down. I tried fiddling with the `tabindex` but that didn't work. I would have to dig a little deeper and possibly change the structure a little to match the previous behavior.

But since other dropdowns allow the user to navigate using the tab, I decided to check here before altering the code. Is that required/intentional, or was the tab navigation disabled in that element accidentally?

>Vocabulary front page
>The separators between fields now look a bit funny. Not sure what happened here, because this didn't happen in earlier iterations.

Also noticed this one yesterday after the rebase. Now, this one I'm quite sure we didn't have before. My best guess is that perhaps this was caused after upgrading to the latest version of Bootstrap? But fixed anyway!

>Vocabulary selection on feedback page
>When selecting the target vocabulary on the feedback page, the selection list looks different. The gray background of the previously selected item is darker. The selection list opens lower down than it used to. The text color of items is now blue, when it used to be black. Items with long names are wrapped incorrectly.

I didn't do a good job testing the dropdown with navigation keys, and also didn't pay attention to form elements modifiers like active and focus. Sorry.

I tried to match the previous feedback page, both the layout, elements, and also how they behave when navigated with mouse or keyboard. Hopefully I got this right now.

>One more thing - the spinner shown when the concept page is loaded via AJAX is wrongly positioned:

Ah! Great catch. I enabled throttling on Firefox network utilities, setting to slow 3G and could reproduce it.
To make it easier, here's a snippet that shows the spinner without having to reload a concept.

```js
var $delayedSpinner = $("<p class='concept-spinner center-block'>Loading more items &hellip;</p>");
$('.content').empty().append($delayedSpinner.hide());
$('.concept-spinner').show();
```

There was a CSS class from Bootstrap3 that was centralizing the spinner. Completely missed that, sorry.

image

Fixed!
---

Phew, I think that's that. Ready for review again. Thanks @osma and @kouralex!

Bruno

osma:
Great work as always @kinow ! I will take a look, but it might take a while - right now our focus is to release 2.13 with the changes that were already merged into `master` - after some additional testing to make sure nothing broke.

Anyway I will get back once I have tested the changes. Maybe we could then merge this for the 2.14 release, if there are no big problems left.

kinow:
Not a problem @osma. If there are any issues I might be able to fix them really quickly if it's nothing major.

Thanks!

kinow:
Will rebase and fix conflicts this week. Should be done by the weekend 👍

kinow:
Rebased! :tada:

kinow:
Rebased.

osma:
Thanks for the rebase @kinow ! I'm looking at this again, with the hope of being able to actually merge this if there are no big showstoppers.

Can you explain the changes in `Dockerfile.ubuntu` and `docker-compose.yml`? It seems you've changed the way the Skosmos code is included into the image. Are these related to the Bootstrap upgrade? Did you intend to include them in this PR?

kinow:
Hi @osma

> Thanks for the rebase @kinow ! I'm looking at this again, with the hope of being able to actually merge this if there are no big showstoppers.

:partying_face: 

> Can you explain the changes in `Dockerfile.ubuntu` and `docker-compose.yml`? It seems you've changed the way the Skosmos code is included into the image. Are these related to the Bootstrap upgrade? Did you intend to include them in this PR?

Ah! My note-to-myself failed me. I was supposed to remove these changes :smile: I use those changes for developing Skosmos as I don't have PHP + composer + etc on my notebook.

Edited the old commits, there should be no more Docker changes here :+1: 

Thanks!
Bruno

osma:
I'm going to add a few small commits to this PR branch, fixing issues that I've found that are easier to just fix myself than to explain :) I already did one.

osma:
The `css_class` mechanism, used to add `active` class to some elements in Twig templates, is a bit broken. The variable is initialized to `['nav-link']` and then an `active` class is conditionally added. But the `active` class is never removed, so the class will be applied also to subsequent elements...

I'm trying to think of a quick fix to make it work better.

kinow:
Thank you for reviewing and fixing the PR too @osma ! I'll keep an eye open for Bootstrap & layout issues in Finto & `master`.

Issue #1186, opened 2021-08-20
JS code executed before a resource is loaded causes hierarchy to not be loaded


kinow:
## At which URL did you encounter the problem?

Running Skosmos from `master` (from some time ago, but also the latest version) with Docker.

## What steps will reproduce the problem?
1. Start the Fuseki+Skosmos with `docker-compose up`
2. Ingest data
3. Visit Skosmos at localhost:9090 or another URL you have configured
4. Now you should be able to reproduce on the first load a scenario where the hierarchy is empty, and you have a JS error in the browser console, similar to (copied Henk's snippet reported via mailing list):

```
Uncaught TypeError: $(...)[0] is undefined
$('script[type="application/ld+json"]')[0].innerHTML = data.graph;
```

## What is the expected output? What do you see instead?

No errors.

## What browser did you use? (eg. Firefox, Chrome, Safari, Internet explorer)

Firefox on Ubuntu LTS.

kinow:
I had seen this error before, but never created an issue. Henk posted on the mailing list and I recognized the issue. I have a theory on why that happens, but never stopped to investigate it. I never saw it happen on the dev instance of Skosmos.

kouralex:
Duplicate of #1016

Whilst a bit different set up and end result (different parts of code are being loaded), I believe that both can be tackled with the same fix.

kinow:
> Duplicate of #1016
> 
> Whilst a bit different set up and end result (different parts of code are being loaded), I believe that both can be tackled with the same fix.

Ah thanks @kouralex ! Let me close this one.

PR #1188, opened 2021-08-21
Modify code to always have an ld json element, may be empty

### LINKED ISSUE #1016: Unhandled JavaScript error causes mapping properties to be unable to load
http://dev.finto.fi/juho/en/groups
1. Go to URL above
2. Click any group in the `Groups` section
3. See Developer Tools -> Console output

I expect no errors to be raised. Instead, I get the following JS error:
`TypeError: $(...)[0] is undefined                     scripts.js:414:8`

It seems like there is an unhandled exception happening at https://github.com/NatLibFi/Skosmos/blob/258c04a4dd9e19991a8178b666c866ea011a0079/resource/js/scripts.js#L414 that causes the mapping properties to be unable to load.

kinow:
I think this issue happens due to a mix of Twig and JS code. When you visit a vocabulary, the value of `request.page == 'page'` is false, since `request.page` has the value "vocab".

But if you click on a concept, it will trigger the code to load concept mappings without changing the value of `request.page`. That code won't find the JSON-LD script element, and result in a JS error. If you reload the page, now the `request.page` will be `page` and now the concept mappings will be displayed correctly.

This pull request simply modifies the template to always have a JSON-LD script element, that may be empty (`{}`). This way the code should work no matter what page is trying to load concept mappings. Tested locally with YSO vocabulary.

Closes #1016

### DIFF
--- a/view/scripts.twig
+++ b/view/scripts.twig
@@ -41,11 +41,13 @@ var pluginParameters = {{ plugin_params|raw }};
 {% endif %}
 </script>
 
-{% if request.page == 'page' and search_results and search_results|length == 1 %}
 <script type="application/ld+json">
+{% if search_results and search_results|length == 1 %}
 {{ search_results|first.dumpJsonLd|raw }}
-</script>
+{% else %}
+{}
 {% endif %}
+</script>
 <script src="vendor/components/jquery/jquery.min.js"></script>
 <script src="vendor/components/jqueryui/jquery-ui.min.js"></script>
 <script src="vendor/components/handlebars.js/handlebars.min.js"></script>


kouralex:
Perfect! Thank you @kinow for implementing this.

PS. I assume this also fixed #1186 ? 😃

kinow:
> 
> 
> Perfect! Thank you @kinow for implementing this.
> 
> PS. I assume this also fixed #1186 ? 😃

Yup, should have fixed both. Now after the release we can ask Henk to confirm it's working OK 🎉 thanks @kouralex !

PR #1033, opened 2020-07-16
Issuexxxx improve http error handling

kouralex:
This PR fixes a bunch or problems in current code.

Fixes the following listed issues:
Fixes #693
Fixes #721
Fixes #722
Fixes #723
Fixes #1011 

Addresses #759 in the sense that it could be implemented in a similar way (add a retry button/handle ajax failure)

Adds PHP fatal error checks for various easyrdf http client induced errors. E.g., doesn't crash the UI in the usual cases for a non-responding SPARQL endpoint (some may still exist).
Global search: adds a warning for every selected, non-default endpoint vocabularies.

Searches may now return 500 Internal Server Error, which on turn is utilized by Ajax queries. Please note that not every failure results in an HTTP 500 Error.
Waypoint system in searches is fixed; no more extra requests.

TODO/open questions:
* Should global search warnings lie above the number of results?
* Translations for various new JS and Twig template variables
* Tests? (and for which part)
* Is the way errors are handled here the best possible?
* Add a retry button to other parts of UI?
* Is the usage of HTTP 500 good as it is?
* Sidebar could be separated for good from the content area, this PR only has started the process.
* Also, it is possible to merge the functionality of Skosmos into a single page app by developing this a bit further.

PR #1195, opened 2021-09-07
Improve search error handling

### LINKED ISSUE #693: Navigation links do not work on a search result page when the SPARQL query returns a 404 error

http://erato.sparna.fr/authority/legal-subject-theme/fr/search?clang=fr&q=this.returns.no.results

1. Go to http://erato.sparna.fr/authority
2. Select any vocabulary
3. Make a full text search for a dummy keyword that returns no results
4. Click on any navigation link in the navigation bar (Vocabularies, About, Feedback)

1.
The links start with "/", like "/about", and are not pointing to the correct path. They sould be prepended by the request language, I think.
It works when the search return some results : http://erato.sparna.fr/authority/legal-subject-theme/fr/search?clang=fr&q=service

However I don't have the same behavior when looking at other skosmos deployments, like vocabularies.unesco.org or finto.fi.

What am I missing ?

osma:
Fixes #693 

This fix has been split off from PR #1033 by creating a new branch from master and cherry-picking two commits by @kouralex . Some merge conflicts were fixed.


osma:
I was able to trigger/simulate the problem described in #693 by editing config.ttl and changing the SPARQL endpoint URL for a vocabulary to a different URL that returns 404. The PR does seem to handle this more gracefully and shows an error like this:

image

However, the message is missing translations. In the above screenshot, the UI language was Finnish but the error still shows up in English.

PR #1196, opened 2021-09-07
Fix sidebar navigation after erroneous page

osma:
Fixes #721
Fixes #722
Fixes #723

These fixes have been split off from PR #1033 by creating a new branch from master and cherry-picking two commits by @kouralex . Some merge conflicts were fixed.

PR #1194, opened 2021-09-07
error-resistant Vocabulary->getInfo()

osma:
This fix has been split off from PR #1033 by creating a new branch from `master` and cherry-picking a single commit by @kouralex . A merge conflict in concept-shared.twig was fixed.

Issue #1170, opened 2021-05-28
PHP fatal error if the requested vocabulary id is not found


kouralex:
Whilst looking at apache2 error log, I noticed that Skosmos throws uncaught PHP fatal errors if the requested vocabulary id is not found. See the following call stack:

```PHP Fatal error:  Uncaught ValueError: Vocabulary id 'xyz' not found in configuration. in model/Model.php:420
Stack trace:
#0 model/Request.php(246): Model->getVocabulary('xyz')
#1 index.php(51): Request->setVocab('xyz')
#2 {main}
thrown in model/Model.php on line 420
```

These should be caught and handled appropriately.

hekl:
I have something like this, these errors are thrown almost every minute. Sometimes with the last referer statement, mostly not. Everything seems fine, except that the loading of the last section of equivalent concepts seems to stall. Watch it: https://classificaties.cbs.nl/geo/def/gemeenten/GM0359

`PHP Fatal error:  Uncaught Error: Class 'ValueError' not found in /var/www/Skosmos/model/Model.php:420\nStack trace:\n#0 /var/www/Skosmos/model/Request.php(246): Model->getVocabulary()\n#1 /var/www/Skosmos/index.php(51): Request->setVocab()\n#2 {main}\n  thrown in /var/www/Skosmos/model/Model.php on line 420, referer: http://classificaties.cbs.nl/sess-bin/login_session.cgi
`

osma:
@hekl It seems that your PHP installation doesn't know about the ValueError class that is used in some parts of Skosmos code since PR #1127 which, among other things, upgraded PHPUnit to version 8. I started investigating why this could happen, and it turns out ValueError was only introduced in PHP 8. But it is also defined by [symfony/polyfill-php80](https://github.com/symfony/polyfill-php80) which Skosmos depends on indirectly through twig-gettext-extractor -- but only if development dependencies are installed! So that explains why you are getting this error (you've used `composer install --no-dev` as instructed in the InstallTutorial) while Skosmos developers usually have the development dependencies, obviously :)

I will open a separate bug report about this "missing ValueError" problem.

osma:
This was mostly fixed in PR #1215 (and the separate but similar problem with the ValueError class, encountered by @hekl, was fixed in PR #1200).

However, looking at Finto server logs, there are still problems in the REST API; uncaught ValueErrors appear for REST API calls with an unknown vocabulary ID.

PR #1197, opened 2021-09-08
Fixes for searches, ajax, waypoints, easyrdf HTTP client errors

osma:
This is the last part of fixes split off from PR #1033. I created a new branch from master (after merging PRs #1194 #1195 #1196) and cherry-picking the last commit in PR #1033 originally created by @kouralex . Some merge conflicts were fixed, hopefully correctly - in `headerbar.twig` I simply discarded the change as the template had been subsequently modified in PR #1061 (adding accessibility features) and it didn't seem to apply anymore.

Original description of changes from the commit message:

> fixes fore failing vocabulary/global search; fixes erroring ajax search queries; fixes waypoint system used in search; added checks for various easyrdf http client errors


controller/WebController.php
```diff
@@ -329,23 +329,47 @@ public function invokeGlobalSearch($request)
         $vocabObjects = array();
         if ($vocids) {
             foreach($vocids as $vocid) {
-                $vocabObjects[] = $this->model->getVocabulary($vocid);
+                try {
+                    $vocabObjects[] = $this->model->getVocabulary($vocid);
+                } catch (Exception $e) {
```
osma, line 334: An empty catch block is bad practice. What happens with the $vocabObjects variable that is used below?
At a minimum, it could be set to null here, or array() if that makes more sense.

resource/js/docready.js
```diff
@@ -917,7 +921,20 @@ $(function() { // DOCUMENT READY
             $('.search-result-listing').append($ready);
             return false;
           }
-          $('.search-result:nth-last-of-type(4)').waypoint(function() { waypointCallback(); }, options );
+          waypoint.destroy();
+          $('.search-result:nth-last-of-type(4)').waypoint(function() { waypointCallback(this); }, options );
+        },
+        error: function(jqXHR, textStatus, errorThrown) {
+          $loading.detach();
+          var $failedSearch = $('<div class="alert alert-danger"><h4>Error: Loading for more items failed!</h4></div>');
```
osma, line 929: We need translations for this new UI string

resource/js/docready.js
```diff
@@ -917,7 +921,20 @@ $(function() { // DOCUMENT READY
...
+          var $retryButton = $('<button class="btn btn-default" type="button">Retry</button>');
```
osma, line 930: "Retry" should be translated as well

view/search-result.twig
```diff
@@ -2,7 +2,15 @@
...
 </div>{% if limit_type or limit_parent or limit_group or limit_scheme %}<button type="button" class="btn btn-default" id="remove-limits">{% trans "Clear limitations" %}</button>{% endif %}
 {% endif %}
-{% if search_results is defined and search_results|length == 0 %}<p class="no-results">{% trans 'The search provided no results.' %}</p>{% endif %}
+{% if global_search and not search_failed and skipped_vocabs|length > 0 %}
+<div class="alert alert-warning">
+{% for voc in skipped_vocabs %}
+{% set vocTitle = voc.title %}
+<h4>{% trans %}Warning: Skipped searching from '{{vocTitle}}' vocabulary!{% endtrans %}</h4>
```
osma, line 9: Message needs to be translated

osma (commented):
Added a few notes on TODO items.

More generally, we need a way to test the changes in this PR, to verify that the fixes are indeed working as they should.

controller/WebController.php
```diff
@@ -329,23 +329,47 @@ public function invokeGlobalSearch($request)
         $vocabObjects = array();
         if ($vocids) {
             foreach($vocids as $vocid) {
-                $vocabObjects[] = $this->model->getVocabulary($vocid);
+                try {
+                    $vocabObjects[] = $this->model->getVocabulary($vocid);
```
osma, line 333: How to trigger the problem addressed here:
http://localhost/Skosmos/fi/search?clang=fi&q=kissa&vocabs=ysa+yso+xyzzy

(here, xyzzy is a vocabulary ID not found in configuration)

Results in:
[Wed Sep 08 14:26:37.964415 2021] [php7:error] [pid 193585] [client 127.0.0.1:35836] PHP Fatal error:  Uncaught ValueError: Vocabulary id 'xyzzy' not found in configuration. in /var/www/html/Skosmos/model/Model.php:420\nStack trace:\n#0 /var/www/html/Skosmos/controller/WebController.php(332): Model->getVocabulary()\n#1 /var/www/html/Skosmos/index.php(44): WebController->invokeGlobalSearch()\n#2 {main}\n  thrown in /var/www/html/Skosmos/model/Model.php on line 420

controller/WebController.php
```diff
@@ -329,23 +329,47 @@ public function invokeGlobalSearch($request)
...
+                } catch (Exception $e) {
+                    // skip vocabularies not found in configuration
+                    // please note that this may result in global search
+                    // NB: should not happen in normal UI interaction
+                }
             }
         }
         $parameters->setVocabularies($vocabObjects);
 
+        $nondefaultEndpointVocs = array();
```
osma, line 343: The idea here is to report to the user (via `skipped_vocabs` passed to Twig template) that actually not all vocabularies were searched in the global search, since only the default SPARQL endpoint is used in global search, but not other vocabulary-specific endpoints.

model/DataObject.php
```diff
@@ -46,10 +46,15 @@ public function __construct($model, $resource)
     protected function getExternalLabel($exvoc, $exuri, $lang)
     {
         if ($exvoc) {
-            $exsparql = $exvoc->getSparql();
-            $results = $exsparql->queryLabel($exuri, $lang);
-
-            return isset($results[$lang]) ? $results[$lang] : null;
+            try {
```
osma, line 49: Here the idea is to catch errors accessing labels from another vocabulary.

model/DataObject.php
```diff
@@ -63,9 +68,15 @@ protected function getExternalLabel($exvoc, $exuri, $lang)
     protected function getExternalNotation($exvoc, $exuri)
     {
         if ($exvoc) {
-            $exsparql = $exvoc->getSparql();
-            $results = $exsparql->queryNotation($exuri);
-            return isset($results) ? $results : null;
+            try {
```
osma, line 71: Same as above, but accessing notations from external vocabularies.

model/Model.php
```diff
@@ -483,8 +490,16 @@ private function disambiguateVocabulary($vocabs, $uri, $preferredVocabId = null)
 
         // no preferred vocabulary, or it was not found, search in which vocabulary the concept has a label
         foreach ($vocabs as $vocab) {
-            if ($vocab->getConceptLabel($uri, null) !== null)
-                return $vocab;
+            try {
```
osma, line 493: Same as above, catch errors that happen when disambiguating URIs

resource/js/scripts.js
```diff
@@ -210,7 +210,8 @@ function loadLimitedResults(parameters) {
   clearResultsAndAddSpinner();
   $.ajax({
     data: parameters,
-    success : function(data) {
+    complete : function(jqXHR, textStatus) {
```
osma, line 213: This callback will be executed in any case, both for successful and unsuccesful requests.

controller/WebController.php
osma: This fix may address #1170

controller/WebController.php
osma: Might be better to fail fast here when any vocabulary ID is not known. At least then we wouldn't accidentally trigger global search.

resource/js/docready.js
```diff
@@ -844,7 +844,7 @@ $(function() { // DOCUMENT READY
       $('.search-result-listing').append($ready);
     }
     else {
-      $trigger.waypoint(function() { waypointCallback(); }, options);
+      $trigger.waypoint(function() { waypointCallback(this); }, options);
```
osma, line 847: The waypoint fixes here address issues when loading more search results on the search result page (possibly also on the alphabetical list in the sidebar) and the AJAX query for fetching more results fails for some reason.

controller/WebController.php
osma: This change fixes one particular path that could lead to the error seen #1170 (and copied above), but there are probably many other ways of triggering the same ValueError.

controller/WebController.php
osma: Switched to fail fast operation.

resource/js/docready.js
osma: Done.

resource/js/docready.js
osma: Done.

resource/js/docready.js
osma: Tested this and the fix seems to be working.

view/search-result.twig
osma: I removed this part of the PR. The message is very frightening and there's nothing the user can do about the situation anyway.

osma:
I think this is good enough now - tests are missing, but handling errors like this are a bit challenging to test anyway. I'll merge this and if anything turns up it can be fixed in subsequent PRs.

PR #1198, opened 2021-09-08
Displaying skos notation with a data type and label in the properties on the concept page

miguelahonen:
If a notation is shown in the properties on the concept page, also a datatype and datatype label are shown.

In the vocabulary config:

skosmos:showNotationAsProperty "true"; -> The notation in the properties is shown

skosmos:showNotationAsProperty "false"; -> The notation in the properties is not shown

commented: #skosmos:showNotationAsProperty "true"; -> The notation in the properties is shown


miguelahonen:
If you need test data...

...a datatype and label can be created like this:

PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>

INSERT DATA
  {
    GRAPH <http://www.yso.fi/onto/juho/>
      {
        <http://www.yso.fi/onto/juho/dt1>  rdf:type rdfs:Datatype ;
        rdfs:label "DatatyyppiTestaus"@fi .
        rdfs:label "DatatypTestning"@sv .
        rdfs:label "DataTypeTesting"@en .
      }
  }
  
  ___
  
 
Adding a notation and datatype:
  
 INSERT DATA
  {
    GRAPH <http://www.yso.fi/onto/juho/>
      {
        <http://www.yso.fi/onto/juho/p5436> skos:notation "Notaatiotesaus"^^<http://www.yso.fi/onto/juho/dt1> .
      }
  }

view/concept-shared.twig
```diff
@@ -96,7 +97,8 @@
           {% for language,labels in concept.allLabels(property.type) %}
             {% for value in labels %}
             <div class="row other-languages{% if loop.first %} first-of-language{% endif %}">
-            <div class="versal col-xs-6{% if value.type == "skos:altLabel" %} replaced{% endif %}">{% if value.containsHtml %}{{ value.label|raw }}{% else %}{{ value.label }}{% endif %}</div><div class="versal col-xs-6">{% if loop.first %} {{ language }}{% endif %}</div></div>
+                <div class="versal col-xs-6{% if value.type == "skos:altLabel" %} replaced{% endif %}">{% if value.containsHtml %}{{ value.label|raw }}{% else %}{{ value.label }}{% endif %}</div><div class="versal col-xs-6">{% if loop.first %} {{ language }}{% endif %}</div>
```
osma, line 100: What's the reason for this change? Seems to be just useless adjusting of whitespace, but hard to tell for sure.

model/Concept.php
```diff
@@ -631,6 +631,7 @@ public function getProperties()
                 $arrayPropObj->addValue($propVal);
             }
         }
+
```
osma, line 634: Please avoid unnecessary whitespace changes

model/Concept.php
```diff
@@ -640,6 +641,9 @@ public function getProperties()
         }
 
         $ret = $this->removeDuplicatePropertyValues($ret, $duplicates);
+
+        $ret = $this->vocab->getConfig()->getShowNotationAsProperty() !== null ? $this->removeNotationFromProperties($ret, $this->vocab->getConfig()->getShowNotationAsProperty()) : $ret ;
```
osma, line 645: Rather than first adding and later removing the notation properties, I think it would be better to avoid adding them to `$ret` in the first place (if the configuration setting is false). That should be done within the foreach block the starts on line 578. So I suggest doing that there (something like: if getShowNotationAsProperty() returns false, and the property type is skos:notation, skip that property) and removing this line.

model/Concept.php
```diff
@@ -680,6 +684,14 @@ public function removeDuplicatePropertyValues($ret, $duplicates)
         return $ret;
     }
 
+    public function removeNotationFromProperties($ret, $isShowNotationAsPropertySet = true)
```
osma, line 687: Also, this method is unnecessary if you're not adding the notation properties to `$ret` at all, as suggested above.

osma (changes requested):
I gave a few suggestions for improvements.

Also, unit tests are failing and that should be investigated. My guess is that this could have been caused by the changes to the SPARQL query, but it's hard to tell and needs to be tested locally.

view/concept-shared.twig
miguelahonen: I was looking for a fancier indentation. Something according this: https://www.granneman.com/webdev/coding/formatting-and-indenting-your-html

Should I revert it to its original?

model/Concept.php
miguelahonen: Done

view/concept-shared.twig
osma: Please revert. It's okay to clean up indentation, but let's keep that separate from changes in functionality if possible. In this case there was (AFAICT) no functional change in the template so this is just adding noise to the PR.

model/Concept.php
miguelahonen: I will fix this too

view/concept-shared.twig
miguelahonen: Okay, I will do..

model/Concept.php
miguelahonen: I think we should reconsider the construction of long conditional statements

Prefer option 1 to option 2 or something else?

https://github.com/NatLibFi/Skosmos/blob/fba94659c41bec4a466c182c8713333f2cb4a00d/model/Concept.php#L578-L595

Guidelines:
https://www.php-fig.org/psr/psr-12/
_Expressions in parentheses MAY be split across multiple lines, where each subsequent line is indented at least once. When doing so, the first condition MUST be on the next line. The closing parenthesis and opening brace MUST be placed together on their own line with one space between them. Boolean operators between conditions MUST always be at the beginning or at the end of the line, not a mix of both._

```
<?php

if (
    $expr1
    && $expr2
) {
    // if body
} elseif (
    $expr3
    && $expr4
) {
    // elseif body
}

```

Or should we keep it separate from changes in functionality as @osma you mentioned here: https://github.com/NatLibFi/Skosmos/pull/1198#discussion_r704414499

osma: I'm sorry, I now see that I gave you bad advice yesterday. Even better than adjusting the foreach loop on line 578 (as you've now done) would be to do the check even earlier, around line 511. Currently there's a check to see if the property is in $this->DELETED_PROPERTIES and if it is, the entire property is skipped.

Perhaps the most elegant option would be doing something like this:

```php
$deleted = $this->DELETED_PROPERTIES;
if (!$this->vocab->getConfig()->getShowNotationAsProperty()) {
  // delete notations as well unless configured to show them
  $deleted[] = 'skos:notation';
}

// adjust the check on line 511 to use $deleted instead of $this->DELETED_PROPERTIES
if (!in_array($prop, $this->DELETED_PROPERTIES) || ($this->isGroup() === false && $prop === 'skos:member')) {
  ...
```

miguelahonen: I suppose it is ready now:

https://github.com/NatLibFi/Skosmos/blob/f0a932db45c114309565e5ba1d25aaaf7a3acc59/model/Concept.php#L512-L517

model/Concept.php
miguelahonen: Done

view/concept-shared.twig
miguelahonen: Is it okay if I skip this now and I will take your recommendation on account from now on

miguelahonen: I had an issue with my IDE but now this is corrected

model/Concept.php
```diff
@@ -508,7 +508,13 @@ public function getProperties()
                 $sprop = "<$prop>";
             }
 
-            if (!in_array($prop, $this->DELETED_PROPERTIES) || ($this->isGroup() === false && $prop === 'skos:member')) {
+            // delete notations unless configured to show them
```
osma, line 511: This block should be moved outside the foreach loop, i.e. around line 499. There is no need to create `$deleted` for every single property value; it can be done just once.

model/Concept.php
```diff
@@ -508,7 +508,13 @@ public function getProperties()
...
+            $deleted = $this->DELETED_PROPERTIES;
+            if ($sprop === 'skos:notation' && !$this->vocab->getConfig()->getShowNotationAsProperty()) {
```
osma, line 513: No reason to check `$sprop === 'skos:notation'` here - and if the whole block is moved, as suggested above, it cannot be done anyway as `$prop` is only set within the foreach loop.

model/VocabularyConfig.php
```diff
@@ -388,6 +388,15 @@ public function getMarcSourceCode($lang = null)
         return $this->getLiteral('skosmos:marcSourceCode', false, $lang);
     }
 
+    /**
+     * Returns a boolean value set in the config.ttl config.
```
osma, line 392: The docstring should be changed to reflect the meaning of the method - something like "Returns the boolean value of the skosmos:showNotationAsProperty setting".

(I see now that there are other methods with similarly uninformative docstrings in the same class - that's bad, but not a reason to not do it right this time)

osma (changes requested):
Starting to look good, I gave a few comments for improvement.

model/Concept.php
miguelahonen: Fixed with this: https://github.com/NatLibFi/Skosmos/pull/1198#pullrequestreview-750353342

model/Concept.php
miguelahonen: It think it is fixed now

miguelahonen: Done

model/Concept.php
miguelahonen: Done

model/VocabularyConfig.php
miguelahonen: Done

model/Concept.php
```diff
@@ -581,7 +587,6 @@ public function getProperties()
                     if (isset($ret[$prop]) && ($literal->getLang() === $this->clang || $literal->getLang() === null) || $this->vocab->getConfig()->hasMultiLingualProperty($prop)) {
                         $ret[$prop]->addValue($literal);
                     }
-
```
kouralex, line 584: Please remove

model/sparql/GenericSparql.php
```diff
@@ -429,6 +430,10 @@ private function generateConceptInfoQuery($uris, $arrayClass, $vocabs) {
   UNION
   {
    ?uri ?p ?o .
+   OPTIONAL {
+     BIND(datatype(?o) AS ?dt)
+     ?dt rdfs:label ?dtlabel
```
osma, line 435: I think this block should check that there is a datatype. Otherwise `?dt` remains unbound and then the query will pick up all `rdfs:label` triples, which wasn't intended.

So add this clause inside the OPTIONAL block:

    FILTER(bound(?t))

osma (changes requested):

model/sparql/GenericSparql.php
miguelahonen: fixed

model/Concept.php
miguelahonen: Fixed

osma (approved):

model/VocabularyConfig.php
```diff
@@ -388,6 +388,15 @@ public function getMarcSourceCode($lang = null)
         return $this->getLiteral('skosmos:marcSourceCode', false, $lang);
     }
 
+    /**
+     * Returns the boolean value of the skosmos:showNotationAsProperty setting.
+     * @return boolean
+     */
+    public function getShowNotationAsProperty()
```
kouralex, line 395: Return type could be added as this is a new function.

model/ConceptPropertyValueLiteral.php
```diff
@@ -31,6 +34,17 @@ public function getLang()
         return $this->literal->getLang();
     }
 
+    public function getDatatype()
```
kouralex, line 37: Docstring + return type missing

PR #1199, opened 2021-09-08
Cleanup translation files and rename trans_script to compile-translations

osma:
There are 3 cleanup style changes to the translation files under `resource/translations`:

1. Remove unnecessary `.mo` files (the real ones are under language specific directories)
2. Remove an unnecessary execute bit on one of the `.po` files
3. Rename the strangely named script `trans_script` to `compile-translations` to better reflect what it does

I don't expect 1. and 2. to be controversial at all, but 3 could be. Also need to update the [wiki documentation on translations](https://github.com/NatLibFi/Skosmos/wiki/Translation#using-the-updated-translations) accordingly.


osma:
Merged the changes to `master` and updated the Translations wiki page accordingly.

PR #1201, opened 2021-09-09
add order of plugins

### LINKED ISSUE #1148: Order of vocabulary-specific plugins is not configurable
At the moment, global plugins are declared in a list (which is – by default – ordered). However, this is not the case for vocabulary-specific plugins, please see https://github.com/NatLibFi/Skosmos/wiki/Plugins#activating-plugins. This will cause trouble as I can not guarantee/change their respective execution order.

I think implementing this will request two kinds of changes to the code base:
1) Deprecate `skosmos:usePlugin` in favor of `skosmos:usePlugins` (the name can be different, that was just an example) that takes a list as a parameter
2) In case there are similarly named plugins in `skosmos:usePlugins` and `skosmos:globalPlugins`, use the order for them declared in `skosmos:usePlugins`. For a reasonable default order, it should be decided whether to append or prepend `skosmos:globalPlugins` plugins with `skosmos:usePlugins` in order to determine their respective execution order.

Alternatively, one could just add a specific `skosmos:executionOrder/Priority` value for each plugin. However, this would make it impossible to have plugins in different order in different vocabularies.

Regarding plugins: it would be better to require a true object relation instead of plain strings. This would enable us to mix parameterized plugins with 'normal' ones in the list. Actually, this way we could still use `skosmos:usePlugin` in the same clever way it is used in parameterized plugin and just shift its use from vocabulary level (deprecate it there) to plugin level.

Vainonen:
Changes into function and variable names ii VocabularyConfig.php :
- $plugins --> $pluginRegister
- function getPlugins --> function getPluginRegister
- json_encode function moved to WebController.php and test below omitted (new tests for encoding in webcontroller?)

Fixes #1148


model/VocabularyConfig.php
```diff
@@ -428,6 +484,14 @@ public function getLanguages()
...
      */
     public function getPluginParameters() {
+        return $this->pluginParameters;
+    }
+
+    /**
+     * Returns the plugin parameters
+     * @return string plugin parameters or null
+     */
+    public function getEncodedPluginParameters() {
```
kouralex, line 494: It seems like this is a new method that is never used? If so, please remove.

model/VocabularyConfig.php
```diff
@@ -31,41 +31,97 @@ class VocabularyConfig extends BaseConfig
...
+     * Get an ordered array of plugin names with order configured in skosmos:vocabularyPlugins
+     * @return array|null of plugin names
+     */
+    public function getPluginArray() {
+
         $pluginArray = array();
+
+        $vocabularyPlugins = $this->resource->getResource('skosmos:vocabularyPlugins');
+        if ($vocabularyPlugins) {
+            // $vocabularyPlugins has all resources
```
kouralex, line 50: ? A bit misleading comment

model/VocabularyConfig.php
```diff
@@ -31,41 +31,97 @@ class VocabularyConfig extends BaseConfig
...
+
+        return $pluginArray;
+
+    }
+
+    /**
+     * Sets array of parameterized plugins
+     * @param Easyrdf\Resource $pluginResource
+     */
+    private function setParametrizedPlugins() {
```
kouralex, line 81: one could add `@return void` and `:void` here

model/VocabularyConfig.php
```diff
@@ -31,41 +31,97 @@ class VocabularyConfig extends BaseConfig
...
-        $plugins = $this->resource->allLiterals('skosmos:usePlugin');
+        $this->globalPlugins = $globalPlugins;
+        $this->setParametrizedPlugins();
+        $pluginArray = $this->getPluginArray();
+        $this->pluginRegister = new PluginRegister($pluginArray);
+    }
+
+    /**
+     * Get an ordered array of plugin names with order configured in skosmos:vocabularyPlugins
+     * @return array|null of plugin names
```
kouralex, line 42: Only `@return array` is possible, isn't it so?

model/VocabularyConfig.php
```diff
@@ -31,41 +31,97 @@ class VocabularyConfig extends BaseConfig
...
+     */
+    public function getPluginArray() {
```
kouralex, line 44: Return type needed here

controller/WebController.php
```diff
@@ -178,7 +178,7 @@ public function invokeVocabularyConcept(Request $request)
         if ($this->notModified($results[0])) {
             return;
         }
-        $pluginParameters = $vocab->getConfig()->getPluginParameters();
+        $pluginParameters = json_encode($vocab->getConfig()->getPluginParameters(), true);
```
kouralex, line 181: What is the purpose on `true` parameter used here and elsewhere? See https://www.php.net/manual/en/function.json-encode.php

model/VocabularyConfig.php
```diff
@@ -31,41 +31,97 @@ class VocabularyConfig extends BaseConfig
...
             }
         }
-        $this->plugins = new PluginRegister(array_merge($globalPlugins, $pluginArray));
-        // Get parameterized plugins defined as resources and their respective parameters
-        $pluginResources = $this->resource->allResources('skosmos:useParamPlugin');
+
+        $pluginArray = array_values(array_unique(array_merge($pluginArray, array_keys($this->pluginParameters))));
+
+        return $pluginArray;
+
```
kouralex, line 74: Extra whitespace

model/VocabularyConfig.php
```diff
@@ -31,41 +31,97 @@ class VocabularyConfig extends BaseConfig
...
+     */
+    private function setParametrizedPlugins() {
+
         $this->pluginParameters = array();
+
+        $vocabularyPlugins = $this->resource->getResource('skosmos:vocabularyPlugins');
+        if ($vocabularyPlugins) {
+            // $vocabularyPlugins has all resources
+            foreach ($vocabularyPlugins as $plugin) {
+                if (!$plugin instanceof EasyRdf\Literal) {
```
kouralex, line 89: Maybe better to use without negation? i.e., `EasyRdf\Resource`

model/VocabularyConfig.php
kouralex: Also `paramaterizedPlugins` instead of `parametrizedPlugins` (need to be fixed elsewhere too)

model/VocabularyConfig.php
```diff
@@ -31,41 +31,97 @@ class VocabularyConfig extends BaseConfig
...
+                    $pluginName = $plugin->getValue();
+                    array_push($pluginArray, $pluginName);
+                }
+                else {
+                    array_push($pluginArray, $plugin->getLiteral('skosmos:usePlugin')->getValue());
+                }
+            }
+        }
+        $pluginArray = array_merge($pluginArray, $this->globalPlugins);
+
```
kouralex, line 62: Rest could be sorted, now they are not. However, this is not required, just a note.

Implementation in short:

Use new array variable
Collect them there
Sort
Then merge and unique

model/VocabularyConfig.php
```diff
@@ -31,41 +31,97 @@ class VocabularyConfig extends BaseConfig
...
-                $pluginArray[] = $pluginlit->getValue();
+                $pluginName = $pluginlit->getValue();
+                array_push($pluginArray, $pluginName);
             }
         }
-        $this->plugins = new PluginRegister(array_merge($globalPlugins, $pluginArray));
-        // Get parameterized plugins defined as resources and their respective parameters
-        $pluginResources = $this->resource->allResources('skosmos:useParamPlugin');
+
+        $pluginArray = array_values(array_unique(array_merge($pluginArray, array_keys($this->pluginParameters))));
```
kouralex, line 71: Why is there a merge with `array_keys($this->pluginParameters)`?

kouralex (changes requested):
Some minor comments on files, looks otherwise very good!

Ps. the starting post could be updated to link to the issue this is fixing.

model/VocabularyConfig.php
```diff
@@ -428,6 +484,14 @@ public function getLanguages()
      * @return string plugin parameters or null
      */
     public function getPluginParameters() {
+        return $this->pluginParameters;
```
kouralex, line 487: Update docstring + return type

model/VocabularyConfig.php
```diff
@@ -31,41 +31,97 @@ class VocabularyConfig extends BaseConfig
...
+    public function getPluginArray() : array
+    {
         $pluginArray = array();
+
+        $vocabularyPlugins = $this->resource->getResource('skosmos:vocabularyPlugins');
+        if ($vocabularyPlugins) {
+            foreach ($vocabularyPlugins as $plugin) {
+                if ($plugin instanceof EasyRdf\Literal) {
+                    $pluginName = $plugin->getValue();
+                    array_push($pluginArray, $pluginName);
```
osma, line 53: Why not simply do

    $pluginArray[] = $plugin->getValue();

instead of using `array_push` and defining a single-use variable? Using `$array[] = ...` is what the [PHP documentation for array_push](https://www.php.net/manual/en/function.array-push.php) recommends when you're only adding a single value:

> Note: If you use array_push() to add one element to the array, it's better to use $array[] = because in that way there is no overhead of calling a function.

model/VocabularyConfig.php
```diff
@@ -31,41 +31,97 @@ class VocabularyConfig extends BaseConfig
...
+                }
+                else {
+                    array_push($pluginArray, $plugin->getLiteral('skosmos:usePlugin')->getValue());
```
osma, line 56: Same here, `$pluginArray[] = ...` would be better.

model/VocabularyConfig.php
```diff
@@ -31,41 +31,97 @@ class VocabularyConfig extends BaseConfig
...
+                }
+            }
+        }
+        $pluginArray = array_merge($pluginArray, $this->globalPlugins);
+
+        $plugins = $this->resource->allLiterals('skosmos:usePlugin');
         if ($plugins) {
             foreach ($plugins as $pluginlit) {
-                $pluginArray[] = $pluginlit->getValue();
+                $pluginName = $pluginlit->getValue();
```
osma, line 65: I think the previous old single line code was good, why change it to make it more complicated?

model/VocabularyConfig.php
```diff
@@ -31,41 +31,97 @@ class VocabularyConfig extends BaseConfig
...
+                array_push($pluginArray, $pluginName);
             }
         }
-        $this->plugins = new PluginRegister(array_merge($globalPlugins, $pluginArray));
-        // Get parameterized plugins defined as resources and their respective parameters
-        $pluginResources = $this->resource->allResources('skosmos:useParamPlugin');
+
+        $pluginArray = array_values(array_unique($pluginArray));
+
+        return $pluginArray;
```
osma, line 72: There's no need for a new variable, you could just use

    return array_values(array_unique($pluginArray));

model/VocabularyConfig.php
```diff
@@ -31,41 +31,97 @@ class VocabularyConfig extends BaseConfig
...
+     * @param Easyrdf\Resource $pluginResource
+     * @return void
+     */
+    private function setParameterizedPlugins() : void
+    {
         $this->pluginParameters = array();
+
+        $vocabularyPlugins = $this->resource->getResource('skosmos:vocabularyPlugins');
+        if ($vocabularyPlugins) {
+            // $vocabularyPlugins has all resources
```
osma, line 86: The comment is a bit unclear. Perhaps you mean that "$vocabularyPlugins only contains resource values" or something along those lines? Is the comment even necessary?

model/VocabularyConfig.php
```diff
@@ -31,41 +31,97 @@ class VocabularyConfig extends BaseConfig
...
+            foreach ($vocabularyPlugins as $plugin) {
```
osma, line 87: The code seems to assume that `$vocabularyPlugins` is an RDF list. What happens if it isn't?

For example:

    skosmos:vocabularyPlugins "plugin1";  # single literal value without list
    skosmos:vocabularyPlugins "plugin1", "plugin2";  # multiple literal values without list
    skosmos:vocabularyPlugins :paramPlugin;  # single resource value
    skosmos:vocabularyPlugins :paramPlugin1, :paramPlugin2;  # multiple resource values

It's quite likely that users will try this kind of configuration, whatever the documentation says...

osma (changes requested):
LGTM, mostly. I gave a few suggestions for minor improvements

tests/VocabularyConfigTest.php
```diff
@@ -515,6 +515,16 @@ public function testGetOrderedPlugins() {
...
   }
 
+  /**
+   * @covers VocabularyConfig::getPluginArray
+   */
+  public function testGetUnorderedVocabularyPlugins() {
+    $vocab = $this->model->getVocabulary('paramPluginOrderTest');
+    $plugins = $vocab->getConfig()->getPluginArray();
+    $arrayElements = ["plugin2", "Bravo", "imaginaryPlugin", "plugin1", "alpha", "charlie", "plugin3"];
+    $this->assertEquals(sort($arrayElements), sort($plugins));
```
kouralex, line 525: Doesn't this double sort (sorting both sides) kind of defy the purpose of the issue this PR is supposed to fix?

Vainonen:
This fixes only a part of the issue #1148. Vocabulary-specific plugins can be configured and listed with skosmos:vocabulyPlugins, but the order listed in that property is changed when plugins are loaded. Therefore the issue is reopened.

PR #1205, opened 2021-09-14
Add "lexical" and "natural" sort strategies for notation codes

### LINKED ISSUE #937: The way concepts with notation codes are sorted should be described in config.ttl
At the moment, there are different types of classification systems published on finto.fi, however, they all must now follow the same sorting setting for notation codes, namely `strnatcasecmp` method. This causes issues whilst sorting e.g., related concepts for http://finto.fi/ykl/en/page/49.262 that has related concepts such as

30.1 Sociology
30.8 Good manners and etiquette
30.11 Social systems
30.12 Sociology of lifestyles. Cultural sociology
30.81 Arranging parties and celebrations
(in this particular order).

Note that for the decimal classification system used for this vocabulary (PCL) is not interoperable with the wanted ordering of related concepts, which would look like this:

30.1 Sociology
30.11 Social systems
30.12 Sociology of lifestyles. Cultural sociology 
30.8 Good manners and etiquette
30.81 Arranging parties and celebrations

That said, there should be a way to describe the way sorting of notation codes is done by Skosmos as some classification systems prefer natural sorting and others do not. Also, there may be other cases (undocumented here) that should be taken into consideration - the implementation should allow user-defined sorting.

Please, have a look at related issues (#265 , #556, #737 and #889) for background details.

osma:
This PR makes it possible to select either "lexical" or "natural" sort strategy for sorting notation codes (see #937 for more background). The sorting is implemented both on the PHP and JS sides. The `skosmos:sortByNotation` configuration setting, which previously was just a boolean value, can now given a literal value to select the appropriate sort strategy:

    skosmos:sortByNotation "false";    # disabled, i.e. sort by labels, not notations
    skosmos:sortByNotation "lexical";  # lexical sorting
    skosmos:sortByNotation "natural";  # natural sorting
    skosmos:sortByNotation "true";     # default sorting i.e. decimal, as it's the most common

TODO:

- [x] more testing - ~~it seems that sorting by notation is not always working on the JS side because notations are missing for the jsTree concept objects~~ (fixed, the problem was in Finto layout footer.inc)
- [x] more unit tests to check that the sort order is as expected
- [ ] update wiki documentation (Configuration)

Fixes #937

### DIFF
--- a/model/ConceptPropertyValue.php
+++ b/model/ConceptPropertyValue.php
@@ -33,12 +33,7 @@ class ConceptPropertyValue extends VocabularyDataObject
 
     public function __toString()
     {
-        $label = is_string($this->getLabel()) ? $this->getLabel() : $this->getLabel()->getValue();
-        if ($this->vocab->getConfig()->sortByNotation()) {
-            $label = ltrim($this->getNotation() . ' ') . $label;
-        }
-
-        return $label;
+        return is_string($this->getLabel()) ? $this->getLabel() : $this->getLabel()->getValue();
     }
 
     public function getLang()
--- a/model/VocabularyConfig.php
+++ b/model/VocabularyConfig.php
@@ -179,20 +179,20 @@ class VocabularyConfig extends BaseConfig
 
     /**
      * Returns the sorting strategy for notation codes set in the config.ttl
-     * config: either "decimal" (default), "natural", or null if
+     * config: either "lexical" (default), "natural", or null if
      * sorting by notations is disabled.
      * @return string|bool
      */
     public function getSortByNotation(): ?string
     {
         $value = $this->getLiteral('skosmos:sortByNotation');
-        if ($value == "decimal" || $value == "natural") {
+        if ($value == "lexical" || $value == "natural") {
             return $value;
         }
         // not a special value - interpret as boolean instead
         $bvalue = $this->getBoolean('skosmos:sortByNotation');
-        // default sorting strategy is "decimal"
-        return $bvalue ? "decimal" : null;
+        // default sorting strategy is "lexical"
+        return $bvalue ? "lexical" : null;
     }
 
     /**
--- a/model/VocabularyConfig.php
+++ b/model/VocabularyConfig.php
@@ -179,8 +179,9 @@ class VocabularyConfig extends BaseConfig
 
     /**
      * Returns the sorting strategy for notation codes set in the config.ttl
-     * config: either "lexical" (default), "natural", or null if
-     * sorting by notations is disabled.
+     * config: either "lexical", "natural", or null if sorting by notations is 
+     * disabled. A "true" value in the configuration file is interpreted as
+     * "lexical".
      * @return string|bool
      */
     public function getSortByNotation(): ?string
@@ -191,7 +192,7 @@ class VocabularyConfig extends BaseConfig
         }
         // not a special value - interpret as boolean instead
         $bvalue = $this->getBoolean('skosmos:sortByNotation');
-        // default sorting strategy is "lexical"
+        // "true" is interpreted as "lexical"
         return $bvalue ? "lexical" : null;
     }
 
--- a/model/VocabularyConfig.php
+++ b/model/VocabularyConfig.php
@@ -178,12 +178,21 @@ class VocabularyConfig extends BaseConfig
     }
 
     /**
-     * Returns a boolean value set in the config.ttl config.
-     * @return boolean
+     * Returns the sorting strategy for notation codes set in the config.ttl
+     * config: either "decimal" (default), "natural", or null if
+     * sorting by notations is disabled.
+     * @return string|bool
      */
-    public function sortByNotation()
+    public function getSortByNotation(): ?string
     {
-        return $this->getBoolean('skosmos:sortByNotation');
+        $value = $this->getLiteral('skosmos:sortByNotation');
+        if ($value == "decimal" || $value == "natural") {
+            return $value;
+        }
+        // not a special value - interpret as boolean instead
+        $bvalue = $this->getBoolean('skosmos:sortByNotation');
+        // default sorting strategy is "decimal"
+        return $bvalue ? "decimal" : null;
     }
 
     /**
--- a/resource/js/scripts.js
+++ b/resource/js/scripts.js
@@ -326,7 +326,14 @@ function countAndSetOffset() {
   }
 }
 
+// return -1 if the value is negative, 1 otherwise
+// used to coerce sort values so they are compatible with the jsTree sort plugin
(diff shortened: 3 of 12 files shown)


model/ConceptPropertyValue.php
```diff
@@ -33,12 +33,7 @@ public function __construct($model, $vocab, $resource, $prop, $clang = '')
 
     public function __toString()
     {
-        $label = is_string($this->getLabel()) ? $this->getLabel() : $this->getLabel()->getValue();
-        if ($this->vocab->getConfig()->sortByNotation()) {
-            $label = ltrim($this->getNotation() . ' ') . $label;
```
osma, line 38: Adding the notation here doesn't seem to have any effect on the sort order so I just removed it.

model/VocabularyConfig.php
```diff
@@ -178,12 +178,21 @@ public function getTitle($lang = null)
...
     /**
-     * Returns a boolean value set in the config.ttl config.
-     * @return boolean
+     * Returns the sorting strategy for notation codes set in the config.ttl
+     * config: either "decimal" (default), "natural", or null if
+     * sorting by notations is disabled.
+     * @return string|bool
      */
-    public function sortByNotation()
+    public function getSortByNotation(): ?string
```
osma, line 186: I renamed the method for better consistency with other getters, and redefined its return value.

resource/js/scripts.js
```diff
@@ -326,7 +326,14 @@ function countAndSetOffset() {
...
 }
 
+// return -1 if the value is negative, 1 otherwise
+// used to coerce sort values so they are compatible with the jsTree sort plugin
+function negVsPos(val) {
+  return (val < 0) ? -1 : 1;
+}
+
 // Natural sort from: http://stackoverflow.com/a/15479354/3894569
+// adapted to return only -1 or 1 using negVsPos function above
```
osma, line 336: Looking at the documentation of the jsTree sort plugin, it expects that the return value of the sort function is either -1 or 1 (anything other than -1 will be interpreted as 1, AFAICT). To make sure that the return values meet those expectations, I added a coercing function.

osma:
> it seems that sorting by notation is not always working on the JS side because notations are missing for the jsTree concept objects

Finally found the cause of the missing notation information: the `footer.inc` customizations in the Finto layout were overriding key JS methods, and they were not in sync with changes that happened in the original Skosmos JS code (that they are based on) a long time ago. Fixed in [this Finto-data commit](https://github.com/NatLibFi/Finto-data/commit/59e5be65ca913bfda1054ab1d80fa6519ad5c584). Now the notation information is available and sorting by notation code works (both for decimal and natural sort order) in the hierarchy tree.

However, it turns out that the natural sort order is a bit different for PHP vs JS. On the PHP side, numbers with leading zeros are always first (PHP uses [Martin Pool's algorithm](http://sourcefrog.net/projects/natsort/)) The current JS implementation of naturalCompare ignores leading zeros. The differences can be seen in this screenshot, where the notation codes 33.03 and 33.04 are placed differently in the hierarchy sidebar (JS) and the concept view (PHP):

image

Not sure if this is a big issue - the example is a bit artificial, because YKL/PLC shouldn't use natural sorting but decimal, where this doesn't happen. And for vocabularies where natural sorting is preferred (i.e. PTVL), leading zeros probably shouldn't be used anyway.

osma:
I added unit tests that verify the sort order is correct on the PHP side in both "decimal" and "natural" cases. Ready for review.

I will update the wiki documentation after merging.

kouralex:
Looks great!

Three remarks, one that you already pinpointed:

Here (look at the image below) the natural order faces the same problem. I am not completely sure, if this vocabulary (HKLJ/HCLCS) should be using the "decimal" order. If that is the case (which may very well be), there is no problem as you mentioned and explained. (URL for your convenience: https://finto.fi/hklj/fi/page/001)
image 

Here, a "real" problem where irregardless of the setting, this always shows up like this: (URL for your convenience: https://finto.fi/hklj/fi/page/691.1)
image

It seems like the sorting is done by the label only. For mapping properties one should probably also sort by its (if vocabulary known to Skosmos) sortByNotation setting. I am not sure if this complicates the issue a bit too much - maybe (I think) it is OK to use the same value.

Here it seems like the search is done using always the "decimal" order, even though one changes the value to "natural". This would mean that no vocabulary may use true "natural" sort strategy in search queries, am I right? The same naturally applies to the typeahead component, too (they use the same implementation under the hood).
image

kouralex (changes requested):
I already commented in the PR where I had three remarks (of which at least two require some action or explanation).

kouralex:
As a fourth item, I also noticed that with "false" value only affects concept info part - it does not affect search nor JS (this may or may not be justified):
image

osma:
Thanks for the review @kouralex . You're right, I didn't consider ordering of search results at all in this PR. I will take a closer look at that and the other things you mentioned.

osma:
Renamed the "decimal" strategy to "lexical". It's not really appropriate for decimal numbers (e.g. "110" sorts before "13") but is just based on lexical string sorting. Amended the PR title and OP accordingly.

osma:
Regarding the four comments:

# 1. Different natural sort orders in PHP and JS

Already discussed above, not a big issue IMHO.

# 2. Search result order

I looked at the ordering of search results by notation in more detail and opened an issue #1209. But I don't think there's anything worth changing in this area so I set that issue to the Blue Sky milestone.

# Mapping property order

> For mapping properties one should probably also sort by its (if vocabulary known to Skosmos) sortByNotation setting. I am not sure if this complicates the issue a bit too much - maybe (I think) it is OK to use the same value.

I opened a separate issue #1208 about ordering mapping property values. I think that's out of scope for this PR and issue.

# 4. Setting "false" has no effect on hierarchy

> As a fourth item, I also noticed that with "false" value only affects concept info part - it does not affect search nor JS (this may or may not be justified):

You are right, I'm currently investigating this. I think this happens because when sorting by notation is disabled (set to false), the sorting of jsTree nodes (representing concepts) falls back to comparison of node text. But that text includes the notation code, so notation codes will affect the sorting anyway.

osma:
I fixed the JS side sorting in the hierarchy when the setting is `false`. I already responded to the three other comments above.
Ready for another round of review and (hopefully) merging @kouralex

kouralex (approved):
Seems to be working now and the issues I pinpointed have been addressed. Great job @osma !

I noticed that the default value is "false" if the property is not set at all, which kind of makes sense, however, the comments in the code imply it could be "lexical" (which is true if the value is "true"!).

Apart from that, ready for merging! 🙂

osma:
I fixed a minor JS issue and clarified the comments about default values. Will merge after tests have passed.

Issue #1189, opened 2021-08-21
Add e2e tests


kinow:
New feature request, so removed the bug template.

At $work we use [Cypress](https://docs.cypress.io/), which while similar to Selenium, is way more stable and less error-prone (i.e. you won't have to add a lot of `setTimeout`s in your code to handle dynamically loaded resources, it uses Promises instead).

Skosmos has several cases that are not easy to be covered by unit tests.

The #1016 bug, for example, requires a test vocabulary (which we have, with the `initFuseki.sh` script starting a Fuseki instance), and requires the user to visit a vocabulary page, then click on a concept. The error can be verified by inspecting the browser console message. These last steps are all easily achieved with a tool like Cypress.

The #1182 issue, another example, has a case where when you start typing a search term, it will create a URL concatenating certain parameters. When I updated Typeahead/Bloodhound, the parameters were not concatenated anymore due to changes in the JS third-party API. While @osma spotted it during the review, the code worked (slower) and it could have resulted in a performance regression. We can use a Cypress to not only drive the UI, but also act as a middleware capturing [JS requests](https://docs.cypress.io/api/commands/server#Syntax) and either forwarding to real Fuseki or ending the test asserting that the URL contains certain parameters. That would prevent the regression @osma spotted.

In order to use Cypress, we could create an `e2e` folder with a `package.json`, `cypress.json`, and the tests. It wouldn't require any changes to Skosmos, only to the GitHub Action workflow.

There are other tools too, like [Nightwatch](https://nightwatchjs.org/), [cavy](https://github.com/pixielabs/cavy), etc.

Another benefit of adding a tool like this with tests now, is that it would future-proof Skosmos should it ever moves to a different UI architecture (SPA, PWA, jQuery-only without Twig, Vue, React, etc). We would be able to modify the UI and just wait for the tests to run and verify that we didn't break anything.

It would have been helpful to have a suite of e2e tests for the #1182 PR.

osma:
Hi @kinow , I'm belatedly replying to this as we are currently setting up the development processes for Skosmos 3 (see the [skosmos-3](https://github.com/NatLibFi/Skosmos/tree/skosmos-3) branch and the [project board](https://github.com/NatLibFi/Skosmos/projects/16)). End-to-end testing is something that we would like to implement, although we haven't made a firm decision yet on how and whether to do this.

[...]

PR #1206, opened 2021-09-15
Use BCP-47 tags for locale in JS

### LINKED ISSUE #1204: [JS] Error when comparing with invalid locale (Incorrect locale information provided)

Running Skosmos locally, with the locale `en_US`. In JavaScript, the locale needs - I think - to be replaced by en-US or en, otherwise we get a `Incorrect locale information provided` from `String.localeCompare`.

```
scripts.js:339 Uncaught RangeError: Incorrect locale information provided
    at String.localeCompare (<anonymous>)
    at naturalCompare (scripts.js:339)
    at a.jstree.plugins.sort.sort (hierarchy.js:534)
    at Array.sort (<anonymous>)
    at a.jstree.plugins.sort.sort (jstree.min.js:5)
    at a.jstree.plugins.sort.sort (jstree.min.js:5)
    at a.jstree.plugins.sort.<anonymous> (jstree.min.js:5)
    at HTMLDivElement.dispatch (jquery.min.js:3)
    at HTMLDivElement.r.handle (jquery.min.js:3)
    at Object.trigger (jquery.min.js:4)
naturalCompare @ scripts.js:339
sort @ hierarchy.js:534
sort @ jstree.min.js:5
sort @ jstree.min.js:5
(anonymous) @ jstree.min.js:5
dispatch @ jquery.min.js:3
r.handle @ jquery.min.js:3
trigger @ jquery.min.js:4
triggerHandler @ jquery.min.js:4
trigger @ jstree.min.js:2
k @ jstree.min.js:2
(anonymous) @ jstree.min.js:2
```

1. Visit any concept of STW, for example
2. Switch to a locale with underscore (e.g. `en_US`, `de_DE` [is this correct?], etc)
3. Open browser console

No errors, `String.localeCompare` uses a valid locale.

Firefox & Chromium, Ubuntu LTS.

kinow:
Prevents errors due to invalid locales with `_` (as per BCP-47).

Closes #1204

### DIFF
--- a/view/scripts.twig
+++ b/view/scripts.twig
@@ -19,7 +19,8 @@ var sr_only_translations = {
 };
 
 <!-- variables passed through to javascript -->
-var lang = "{{ request.lang }}";
+// We replace underscores by dash so the locale complies with BCP-47 used by ECMAScript.
+var lang = "{{ request.lang }}".replace("_", "-");
 var content_lang = "{{ request.contentLang }}";
 var vocab = "{{ request.vocabid }}";
 {% if parameters %}


kinow:
@osma fixing it was super easy thanks to your pointers on where to look at the code. However, not sure how to write a test for this one... our PHPUnit are mainly RDFGraph models, PHP controllers, request/response. I think we have no tests for JS yet? (kinda related to #1189 I think)

Thanks!

osma:
The fix looks reasonable - except for the question I made in [this comment](https://github.com/NatLibFi/Skosmos/issues/1204#issuecomment-919863186): does it even make sense to use underscores for language tags in config.ttl? We know they cause problems for JS, and this fixes that issue, but are they sensible on the PHP side? If not, the real problem is elsewhere and perhaps they should be fixed (underscores converted to dashes) already when reading config.ttl (perhaps emitting a warning if underscores are used).

kinow:
Replied to comment in the issue :+1:

PR #1207, opened 2021-09-15
Replace NBSP by spaces

kinow:
Was reading the code, searching for language label/value, and noticed the `NBSP` character. It's not really a space char, although most browsers/IDE's support it, it can still cause issues when refactoring with a regex like `\s` for instance.

The GitHub UI will show like a space. But if you open on PhpStorm, or `vim` (click `G` followed by `A` to show the current char code in the status bar of `vim`) you should be able to see the char.

### DIFF
--- a/model/DataObject.php
+++ b/model/DataObject.php
@@ -50,7 +50,7 @@ class DataObject
                 $exsparql = $exvoc->getSparql();
                 $results = $exsparql->queryLabel($exuri, $lang);
                 return isset($results[$lang]) ? $results[$lang] : null;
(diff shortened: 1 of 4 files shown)


osma:
Good catch @kinow ! I take it that you searched for NBSP characters in the full codebase and these were the only occurrences?

kinow:
> Good catch @kinow ! I take it that you searched for NBSP characters in the full codebase and these were the only occurrences?

There were two other, but in those i18n files (.po?). Couldn't tell if the files were generated via that transifex scropt or if I could change them.

osma:
OK, let's leave those for later. Merged, thanks!

PR #1212, opened 2021-09-16
Fix notation data type tests

osma:
A couple of minor fixes after merging PR #1198:

- fix the coverage annotations (so that the test coverage is properly assessed for the getDatatype method)
- change getDatatype to return `null` if a datatype exists without a label (and modify corresponding test)

ping @miguelahonen


osma:
Hooray, coverage increased by 9 lines! (+0.22%)
Will merge.

Issue #1222, opened 2021-10-15
Accessibility improvement: Remove hardcoding of allcaps label styling, move to CSS


schlawiner:
There are several places in Skosmos where text (mostly labels for concepts) are printed in allcaps, which is known to be  hard to read for many people (dyslexics and others) and can confuse screenreader apps used by blind people. 

I tried to change this for my own instance of Skosmos and found that it has been hardcoded into the twig files - using `|upper` in most places, so easy to search and replace, and in one place as `toUpperCase`. 

My suggestion is to remove these uppercasing filters and re-add them into the CSS stylesheet, where they can be easily overridden. This is what I used in my own custom stylesheet after removing the twig uppercasing:

```
.property-label, .logo-header > h2, .vocab-category > h3 {
    text-transform: uppercase;
}
```

All instances of |upper and toUppercase are related to this and can safely be deleted, I found, and replaced with above CSS snippet.

This is what it looks like with the CSS property activated (identical to before, when it was in twig): 
<img width="992" alt="Bildschirmfoto 2021-10-15 um 11 55 46" src="https://user-images.githubusercontent.com/9462565/137469763-204fc7d5-aa73-4d54-9716-6c433c67960e.png">

And this is what it looks like when it is turned off on the client side (or via custom stylesheet):
<img width="980" alt="Bildschirmfoto 2021-10-15 um 11 56 56" src="https://user-images.githubusercontent.com/9462565/137469924-87f96c69-3172-430c-aab6-78f63753e695.png">

osma:
Thanks for the suggestion! This makes a lot of sense.

Since you've already made the changes locally, do you think you could contribute a pull request with those changes?

schlawiner:
Done :). Some trepidation, as this is my first pull request for a public/open source project.

PR #1223, opened 2021-10-15
Move allcaps in property-labels (and vocabulary-category) from twig  to css

schlawiner:
Making the changes discussed in issue #1222: removing `|upper` and `toUpperCase` from twig files to make these labels regular case again, then adding the relevant classes to styles.css and using `text-transform: uppercase`on them.

### DIFF
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -751,6 +751,10 @@ span.xl-pref-label > img {
   padding: 15px 0 15px 10px;
 }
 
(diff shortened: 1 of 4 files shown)


miguelahonen:
Tested with a local installation of Skosmos and looks to be working fine. Thank you very much @schlawiner :+1:

miguelahonen:
Screenshot from 2021-11-03 12-02-22

Issue #1221, opened 2021-10-14
[JS] jstree error when top concept's preLabel contains special characters


schlawiner:
## At which URL did you encounter the problem?
Locally (I use the docker-compose version)

## What steps will reproduce the problem?
1. Load a vocabulary with Concepts whose prefLabel contains "()". Example: 
```patype:bachelorThesis a skos:Concept ;
    skos:inScheme patype:; skos:topConceptOf patype:;
    skos:notation "723";
    skos:prefLabel "Thesis (Bachelor)"@en, "Abschlussarbeit (Bachelor)"@de;
    skos:altLabel "bachelor's thesis"@en, "bachelor thesis"@en, "bachelor dissertation"@en, "Bachelor-Arbeit"@de;
    skos:scopeNote "theses to earn a bachelor's degree or equivalent"@en;
    skos:exactMatch fabio:BachelorsThesis, crt:c_7a1f;
    skos:broadMatch fabio:Thesis, eprint:Thesis, schema:Thesis
.
```

2. Start on the Vocabulary overview page (for me: http://localhost:9090/de/)
3. Click the offending vocabulary and wait for it to open. This specific vocab is set to display "Hierarchy" view by default.

## What is the expected output? What do you see instead?

Expected: the hierarchy view of my vocabulary, listing all the concepts. 
Instead: A "loading" spinner in the sidebar that never finished. 

<img width="1402" alt="Bildschirmfoto 2021-10-14 um 14 17 51" src="https://user-images.githubusercontent.com/9462565/137319743-2e754b8a-bc7e-419d-8afe-8dacf81f29e3.png">

In the console:
```
Uncaught Error: Syntax error, unrecognized expression: thesis (bachelor)
    at Function.fa.error (jquery.min.js:2)
    at fa.tokenize (jquery.min.js:2)
    at fa.select (jquery.min.js:2)
    at Function.fa [as find] (jquery.min.js:2)
    at n.fn.init.find (jquery.min.js:2)
    at new n.fn.init (jquery.min.js:2)
    at n (jquery.min.js:2)
    at nodeLabelSortKey (hierarchy.js:437)
    at a.jstree.plugins.sort.sort (hierarchy.js:550)
    at Array.sort (<anonymous>)
```

[...]

PR #1227, opened 2021-10-21
Fix hierarchy js error for special characters in prefLabels

schlawiner:
related to #1221.

The trouble was that in one place when building the tree, the getLabel function wasn't used to create the html for notation and preflabel. The getLabel function unescapes special characters in labels and also adds a .tree-label class to the label span, which is later filtered for when sorting the tree in naturalCompare - if it doesn't exist, the variable it depends on isn't filled and the whole thing breaks down.

Lines 221 and 222 are removed because a) the same thing is done in getLabel, and 2) because it didn't escape special chars and doesn't add the .tree-label class that getLabel does.

### DIFF
--- a/resource/js/hierarchy.js
+++ b/resource/js/hierarchy.js
@@ -210,7 +210,7 @@ function vocabRoot(topConcepts) {
   for (var i = 0; i < topConcepts.length; i++) {
     var conceptData = topConcepts[i];
     var childObject = {
-      text: conceptData.label, 
+      text: getLabel(conceptData), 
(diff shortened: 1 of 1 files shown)


osma:
Thanks a lot @schlawiner - this is perfect. I tested it with the UNESCO Thesaurus and this indeed fixes the problems with the hierarchy display of top level concepts.

There may still be additional cases that are broken, where a similar `getLabel()` call needs to be added, but those can be handled in separate PRs.

PR #1234, opened 2021-11-05
Fix language switcher "eating" part of vocids that end in a langcode (e.g. "interessen" and "en")

schlawiner:
I happend to have a vocabulary named (vocid) "interessen", which ends in "en", one of the available ui languages. 
When switching the UI to German, the "en" ending in the url (**/interessen/en/**) was replaced with the "de" code, as well as appending another /de. So en was (re)placed twice, leading to broken urls such as /interess**de**/de/.

I fixed this for now by making the "| replace" filter that rewrites the url not as greedy: instead of any langcode-looking text, it now looks for a slash preceding it (/en instead of just en). This will probably not fix instances where a vocid is exactly the same as one of the langcodes (vocabularies named "en", "de", "fr" etc.) but is good enough for me at the moment.

### DIFF
--- a/view/topbar.twig
+++ b/view/topbar.twig
@@ -25,7 +25,7 @@
   <ul class="dropdown-menu dropdown-menu-right">
   {% for langcode, langdata in languages %}
   {% if request.lang != langcode %}
-  <li><a id="language-{{ langcode }}" class="versal" href="{{ request.langurl|replace({("#{request.lang}/"): "#{langcode}/"}) }}"> {{langdata.lemma}}</a></li>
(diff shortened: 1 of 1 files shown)


osma:
Thanks! Looks good but needs to be tested.

You are right that's it's possible to end up in trouble by having a vocabulary id that is identical to a language code. The URL pattern for Skosmos doesn't support that case very well.

osma:
I tested this both with and without the dropdown-style UI language selector. It does exactly what it should, so I'm merging it now. Thanks a lot @schlawiner !

PR #1235, opened 2021-11-06
Faster docker build.

pulquero:
Moved the config into a layer post "composer install" to speed up build when only changing config.

### DIFF
--- a/dockerfiles/Dockerfile.ubuntu
+++ b/dockerfiles/Dockerfile.ubuntu
@@ -67,8 +67,13 @@ RUN echo "ServerName localhost" >> /etc/apache2/apache2.conf
 WORKDIR /var/www/html
 RUN rm index.html
 
-COPY . /var/www/html
+# composer and packages layer
 RUN curl -sS https://getcomposer.org/installer | php
+COPY composer.json /var/www/html
+RUN php composer.phar install --no-dev --no-autoloader
+
+# skosmos layer
+COPY . /var/www/html
 RUN php composer.phar install --no-dev
 
 


kinow (commented):
Looks Ok to me. Only issue is remembering to update image should directory layout changes. I'm not sure which option I prefer, faster build or simpler maintenance.

pulquero:
Potentially you could have both if nested config and/or the other directories under different parents.Sent from my Galaxy
-------- Original message --------From: "Bruno P. Kinoshita" ***@***.***> Date: 06/11/2021  19:28  (GMT+00:00) To: NatLibFi/Skosmos ***@***.***> Cc: pulquero ***@***.***>, Author ***@***.***> Subject: Re: [NatLibFi/Skosmos] Faster docker build. (PR #1235) 
@kinow commented on this pull request.

Looks Ok to me. Only issue is remembering to update image should directory layout changes. I'm not sure which option I prefer, faster build or simpler maintenance.

kinow:
@pulquero I think I see what you mean. If the source lived in a directory like `src`, and the config elsewhere, maybe in a `config` folder, then we could have two distinct layers, building or re-building the image would be much faster, and maintenance of source & docker files would also be quite easy. Let's wait and see what others think too :+1:

pulquero:
Here, found a way to get the php packages into their own layer.

kinow (commented):
This looks like a great fix @pulquero ! Not approving because I didn't have time to try it out locally, but in principle looks like a great improvement, with little to no difference in maintenance for devs :clap:

miguelvaara:
Tested carefully with comparison between NatLibFi:master and pulquero:docker. Pulquero's changes results the faster building while changing only the configuration. Nice work!

Thanks @pulquero for the PR and thanks @kinow for your commenting.

PR #1236, opened 2021-11-09
Added support for scoping vocabularies by skos:inScheme.

pulquero:
New vocabulary config property skosmos:conceptSchemes to optional scope queries using skos:inScheme. This is an alternative to using GRAPHs or different endpoints, particularly when dealing with highly interconnected vocabularies.


kinow (commented):
Had a quick look, but not reviewing it :+1: one minor comment that caught my eye. Great work @pulquero!

PR #1239, opened 2021-11-09
Normalize the search term

### LINKED ISSUE #1184: Searching with decomposed unicode characters
http://finto.fi/finaf/fi/page/000185275

1. Copy a keyword used in search from a text with decomposed unicode characters 

Search will not find similar prefLabel if the decomposed unicode character in keyword is precomposed in the vocabulary.

joelit:
Fixes #1184

### DIFF
--- a/model/ConceptSearchParameters.php
+++ b/model/ConceptSearchParameters.php
@@ -71,6 +71,7 @@ class ConceptSearchParameters
         if (!$term && $this->rest)
             $term = $this->request->getQueryParamRaw('label');
         $term = trim($term); // surrounding whitespace is not considered significant
+        $term = Normalizer::normalize( $term, Normalizer::FORM_C ); //Normalize decomposed unicode characters #1184
         if ($this->rest) {
             return $term;
         }


osma:
I tested this and it works as it should. The only issue is that highlighting (with bold text) the matching parts in the autocomplete box is not working when the input string is decomposed (composed on the left, decomposed on the right):

kuva

I tried to fix this by adding a `.normalize('NFC')` call here: https://github.com/NatLibFi/Skosmos/blob/30ce5d0de7e9b886caab6bb8c6db728156f1e128/resource/js/docready.js#L804
...but it didn't help; I think it would require changes within typeahead.js as it seems to read the search string directly from the text field, so there is no easy opportunity to normalize it. Anyway this is no big deal; there are other similar problems with missing highlights (e.g. in case of accent folding) and these should be pretty rare cases anyway.

I think it's reasonable to place the normalization call in ConceptSearchParameters.getSearchTerm(), as that method also performs other types of search term normalization such as stripping whitespace.

osma (approved):

kinow:
>I think it would require changes within typeahead.js

I think #1182 is also updating typeahead. I can try to test it with the latest version and comment if we have a follow-up issue for that @osma :+1: (I will see if I have some text with surrogate characters or just manually edit a dataset and upload to my test fuseki)

osma:
>  will see if I have some text with surrogate characters or just manually edit a dataset and upload to my test fuseki

You don't need very special data for this, just text with non-ASCII characters (e.g. åäöéñ) that are stored as composed Unicode characters (NFC), which is the usual case for RDF data.

The problem was when the user enters a search string which contains decomposed (NFD) characters (typically copied and pasted from some other system - in our case an Aleph ILS). [This comment](https://github.com/NatLibFi/Skosmos/issues/1184#issuecomment-964471499) in the original issue shows how to trigger it with Linux command line tools and the KANTO/FINAF data set in Finto.

joelit:
Thanks for spotting this defect, and thanks for the valuable input. :)  I'll make a new issue regarding the autocomplete highlighting bug, and call this one an incremental improvement on the situation (and I'll link this discussion there).

PR #1241, opened 2021-11-10
Remove float property of reified property element

### LINKED ISSUE #1238: Layout bug in display of notes
https://skosmos.dev.finto.fi/stw/en/page/?uri=http%3A%2F%2Fzbw.eu%2Fstw%2Fdescriptor%2F10286-6

1. Go to a concept which has reified notes in many languages, e.g. STW concept "Accelerators" (above URL)

The notes in both English and German are displayed on the same line:

kuva

Expected them to be on separate lines; though the German note wouldn't need to be displayed at all in this case since the content and UI languages are set to English.

Firefox

kinow:
Closes #1238

### DIFF
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -658,10 +658,6 @@ ul.dropdown-menu > li:last-child > input {
 
 }
 
-.reified-property-value {
-  float: left;
-}
-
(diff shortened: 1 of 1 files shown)


osma:
Thanks @kinow !

PR #1245, opened 2021-11-10
GraphDB text-search dialect.

### LINKED ISSUE #628: Provide a "GraphDB" dialect for full-text search
I like GraphDB. It would be nice if Skosmos can talk to GraphDB using its full-text operators. 
See the documentation at http://graphdb.ontotext.com/documentation/free/lucene-graphdb-connector.html and also https://stackoverflow.com/questions/43517064/how-to-create-a-custom-analyzerfactory-in-graphdb-full-text-search.
I will investigate and provide a patch by looking at what already exists for JenaText.

pulquero:

- Closes #628 

## Checklist


joelit:
Thanks, this is mighty interesting - especially the lucene connectors you included. We'll have to try this out to see how it works.

charlycou:
Any reason for not merging this one?
I just try to set up Skosmos 2.18.11 with GraphDB 10.6.2. 

image

From GraphDB Logs : `org.eclipse.rdf4j.http.server.ClientHTTPException: MALFORMED QUERY: projection alias 'distLabels' was previously used at org.eclipse.rdf4j.http.server.repository.handler.DefaultQueryRequestHandler.getQuery(DefaultQueryRequestHandler.java:185)`

osma:
Thanks for raising this again @charlycou !

We are just about to release Skosmos 3.0. It's very unlikely that we will do any further Skosmos 2.x releases. See #1473 for details.

As this PR is targeting the old Skosmos 2 branch, we will probably not merge it, as we only do very critical fixes in that branch and no new features. Also, it would need some documentation on how to test it and use it, for example as part of the PR in a comment. Automated tests that can be run under GitHub Actions would also be a big plus...

So as it stands we don't have the capacity to process this. But if this would be re-targeted against the `main` branch, ideally with the above enhancements, then we could consider it for Skosmos 3.x.

PR #1251, opened 2021-12-01
Redirect error log to stderr for docker logs.

pulquero:

Have errors printed straight to docker logs to avoid having to exec into the container to check them.

- Closes #

## Checklist

### DIFF
--- a/dockerfiles/Dockerfile.ubuntu
+++ b/dockerfiles/Dockerfile.ubuntu
@@ -62,9 +62,9 @@ COPY dockerfiles/config/000-default.conf /etc/apache2/sites-available/000-defaul
 RUN a2enmod rewrite
 RUN a2enmod expires
 
-RUN echo "ServerName localhost" >> /etc/apache2/apache2.conf
-# redirect error log to stderr for docker logs
-RUN sed -ri \
+# set ServerName & redirect error log to stderr for docker logs
+RUN echo "ServerName localhost" >> /etc/apache2/apache2.conf && \
+        sed -ri \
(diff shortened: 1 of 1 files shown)


dockerfiles/Dockerfile.ubuntu
```diff
@@ -63,6 +63,10 @@ RUN a2enmod rewrite
 RUN a2enmod expires
 
 RUN echo "ServerName localhost" >> /etc/apache2/apache2.conf
+# redirect error log to stderr for docker logs
+RUN sed -ri \
+        -e 's!^(\s*ErrorLog)\s+\S+!\1 /proc/self/fd/1!g' \
+        "/etc/apache2/apache2.conf"
```
kinow, line 69: The two lines above could be combined to keep it a single layer modifying `/etc/apache2/apache2.conf`.

There's a configuration for the error log in the virtual host of Skosmos too here: https://github.com/NatLibFi/Skosmos/blob/a770d4c8bcf8407ebae5f43ea70310f614cfc1cb/dockerfiles/config/000-default.conf#L7

But I think this one would be the catch-all error handler. If we adopt the official PHP image as base layer it won't be necessary (and I think it could be helpful for other things too), for a follow-up issue/discussion.

osma: I combined the RUN directives into a single line, thanks for the suggestion @kinow!

As for adopting the official PHP image, I think that's a different discussion...

osma:
I combined the RUN directives (thanks for the suggestion @kinow!) and pushed the commit directly into the PR branch.
Thanks for the contribution @pulquero and sorry for the long delay!

Issue #1247, opened 2021-11-11
Upgrade Skosmos to a new translation component


joelit:
This is an upgrade request for getting rid of the abandoned `twig/extensions` package. 

## What steps will reproduce the problem?
1. Currently we are using the `I18nExtension` for creating the translation strings from .mo files
2. The library this extension is part of is abandoned and sould not be used https://packagist.org/packages/twig/extensions
3. The recommendation is to use the `trans` filter in `symfony/twig-bridge` instead

## What is the expected output? What do you see instead?
The new package has a different set-up for loading the translation files, as documented in https://symfony.com/doc/current/translation.html
Also the often used `{% trans ID_STRING %}` syntax does not work with the extension in symfony/twig-bridge
* It shoud be refactored to `{{ ID_STRING | trans }}` or `{% trans %}ID_STRING{% endtrans %}`
 * All three are valid syntaxes in the current `I18nExtension` set-up

osma:
This has been done already. We have switched to Symfony Translations for Skosmos 3.0. Closing the issue.

PR #1257, opened 2021-12-08
Remove the dependency to deprecated gettext-extractor package

### LINKED ISSUE #1153: Updating twig-gettext extractor (and possibly changes for translating Finnish, Swedish and English UI)
The [Poedit](https://poedit.net) used for syncing the translation UI templates is used with twig-gettext-extractor tool.
 - Package twig/extensions is abandoned, you should avoid using it. No replacement was suggested.
nat-lib-fi/skosmos               dev-master  requires  twig/extensions (1.5.*)  
umpirsky/twig-gettext-extractor  v1.3.1      requires  twig/extensions (~1.0) 

There are several branches available for [twig-gettext-extractor](https://packagist.org/packages/roydejong/twig-gettext-extractor?query=twig-gettext-extractor), but none have been updated recently. If we want to drop this dependecy, we should consider the current setup of Poedit or whether we can continue updating the translation templates manually without Poedit.

joelit:

This PR removes the dependency to twig-gettextExtractor, as it is a package no longer maintained

- Closes #1153

This PR will not get rid of related and abandoned twig-extensions package, as it is used by the twig translation filters in Skosmos. There is a separate issue of that in #1247

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -82,7 +82,6 @@
   },
   "require-dev": {
     "phpunit/phpunit": "8.5.*",
-    "umpirsky/twig-gettext-extractor": "1.3.*",
     "symfony/dom-crawler": "3.4.3",
     "mockery/mockery": "1.3.1"
   },


kinow (approved):
Looks good to me. No code updated, so I assume it was not in use. CI green so shouldn't break anything I guess.

osma:
There is also a static, modified copy of `twig-gettext-extractor` under [tools/twig-gettext-extractor](https://github.com/NatLibFi/Skosmos/blob/master/tools/twig-gettext-extractor). If we want to excise it completely, that should probably be removed too (which would also remove the `tools` directory as it's the only file there).

osma:
I just stumbled on this old open PR and concluded that it is no longer relevant. Closing to reduce clutter.

PR #1259, opened 2021-12-22
The vocabulary list on the frontpage shows the vocabulary names in th…

miguelvaara:
…e default language of the vocabulary

Thanks @henriyli, @osma and @kouralex for your contribution :-)

To follow wishes mentioned in the PR #747 

[#731](https://github.com/NatLibFi/Skosmos/tree/henriyli-henriyli-fix-731)
- Closes #

The vocabulary list on the front page shows the vocabulary names in the default language of the vocabulary

## Checklist
1. For the vocabulary of your choice, point out the default language in the config.ttl (make sure the dc:title is also defined with respect to the default language)
2. In Skosmos, select the content language not used in the vocabulary
3. On the Skosmos front page check if the vocabulary name in the list is shown in the default language  

When it comes to the phpUnit tests, at the moment I am investigating ways to run PHPUnit tests for Symphony Twig templates. Tests are coming a bit later.

### DIFF
--- a/view/vocabularylist.twig
+++ b/view/vocabularylist.twig
@@ -15,7 +15,14 @@
       <h3>{{ vocabClassName }}</h3>
       <ul>
       {% for vocab in vocabArray %}
-        <li><a class="navigation-font" href="{{ vocab.id }}/{{ request.lang }}/{% if request.contentLang != request.lang and request.contentLang != '' and request.contentLang in vocab.config.languages %}?clang={{ request.contentLang }}{% endif %}">{{ vocab.title }}</a></li>
+        {% set useDefaultLanguage = null %}
+        {% if request.contentLang not in vocab.config.languages and request.contentLang != '' %}
+          {% set useDefaultLanguage = true %}
+        {% endif %}
+        <li><a class="navigation-font" href="{{ vocab.id }}/{{ request.lang }}/{% if request.contentLang not in vocab.config.languages and request.contentLang != '' %}?clang={{ vocab.config.defaultLanguage }}{% endif %}">
+          {% if useDefaultLanguage is not null %} {{ vocab.title(vocab.config.defaultLanguage) }}
+            {% else %} {{ vocab.title(request.contentLang) }}
+          {% endif %} </a></li>
       {% endfor %}
       </ul>
     </div>


view/vocabularylist.twig
```diff
@@ -15,7 +15,14 @@
...
       {% for vocab in vocabArray %}
-        <li><a class="navigation-font" href="{{ vocab.id }}/{{ request.lang }}/{% if request.contentLang != request.lang and request.contentLang != '' and request.contentLang in vocab.config.languages %}?clang={{ request.contentLang }}{% endif %}">{{ vocab.title }}</a></li>
+        {% set useDefaultLanguage = null %}
+        {% if request.contentLang not in vocab.config.languages and request.contentLang != '' %}
+          {% set useDefaultLanguage = true %}
+        {% endif %}
+        <li><a class="navigation-font" href="{{ vocab.id }}/{{ request.lang }}/{% if request.contentLang not in vocab.config.languages and request.contentLang != '' %}?clang={{ vocab.config.defaultLanguage }}{% endif %}">
+          {% if useDefaultLanguage is not null %} {{ vocab.title(vocab.config.defaultLanguage) }}
+            {% else %} {{ vocab.title(request.contentLang) }}
+          {% endif %} </a></li>
```
kinow, line 25: Couldn't we use something like (not tested) this?

```suggestion
        {% set useDefaultLanguage = request.contentLang not in vocab.config.languages and request.contentLang != '' %}
        <li><a class="navigation-font" href="{{ vocab.id }}/{{ request.lang }}/{% if useDefaultLanguage %}?clang={{ vocab.config.defaultLanguage }}{% endif %}">
          {% if useDefaultLanguage %} {{ vocab.title(vocab.config.defaultLanguage) }}
          {% else %} {{ vocab.title(request.contentLang) }}
          {% endif %} </a></li>
```
I think we have the same condition in two places, so maybe storing the `true` or `false` value of that condition in that variable would simplify it?

miguelvaara: The way you proposed was clearly more elegant, I changed the code. Thanks a lot @kinow

miguelvaara:
Another question is where the business logic is located. For instance, I think it is also possible to modify a request according to the needs mentioned in the original issue. The configuration and therefore the default language of the vocabulary could be checked before accessing the data depending on the request.

PR #1261, opened 2022-01-04
Fix REST search when using zero ('0') as the query parameter

### LINKED ISSUE #1260: Vocabulary search using "0" as the query argument throws an error
https://api.finto.fi/rest/v1/ykl/search

1. GET https://api.finto.fi/rest/v1/ykl/search?query=0

As searching by notation is a thing and 0 is a valid notation I would expect the search to work and find this item: http://urn.fi/URN:NBN:fi:au:ykl:0, just like it works with https://api.finto.fi/rest/v1/ykl/search?query=1.

Currently it throws `400 Bad Request : query parameter missing`. I'd assume this comes from php evaluating "0" to false

kinow:

Prevent a string value being treated as a boolean, causing a valid search query to fail.

- Closes #1260

1. added a unit test with some cases where it should pass, and where it should fail. It's using parameters with phpunit, so this can be extended later if needed to cover more cases.
2. fixed the code by using `isset()` and also trimming the value, and checking if the `strlen` returned is `zero`.
3. happy to revert, but the invalid `return`'s have always bothered me in the controller methods :disappointed_relieved: so I also fixed that by using `return`, and annotated the params & method changed with types (I think our minimum version of PHP accepts it? if not, and CI fails, I'll revert this part)

## Checklist

### DIFF
--- a/controller/RestController.php
+++ b/controller/RestController.php
@@ -169,27 +169,30 @@ class RestController extends Controller
      * Performs the search function calls. And wraps the result in a json-ld object.
      * @param Request $request
      */
-    public function search($request)
+    public function search(Request $request): void
     {
         $maxhits = $request->getQueryParam('maxhits');
         $offset = $request->getQueryParam('offset');
         $term = $request->getQueryParamRaw('query');
 
-        if (!$term && (!$request->getQueryParam('group') && !$request->getQueryParam('parent'))) {
-            return $this->returnError(400, "Bad Request", "query parameter missing");
+        if ((!isset($term) || strlen(trim($term)) === 0) && (!$request->getQueryParam('group') && !$request->getQueryParam('parent'))) {
+            $this->returnError(400, "Bad Request", "query parameter missing");
+            return;
         }
         if ($maxhits && (!is_numeric($maxhits) || $maxhits <= 0)) {
-            return $this->returnError(400, "Bad Request", "maxhits parameter is invalid");
+            $this->returnError(400, "Bad Request", "maxhits parameter is invalid");
+            return;
         }
         if ($offset && (!is_numeric($offset) || $offset < 0)) {
-            return $this->returnError(400, "Bad Request", "offset parameter is invalid");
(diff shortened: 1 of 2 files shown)


kinow:
SoanrCloud analysis is failing due to duplicated code, but as far as I can tell, none is in this PR:

image

osma:
>    added a unit test with some cases where it should pass, and where it should fail. It's using parameters with phpunit, so this can be extended later if needed to cover more cases.

Great work!

>    fixed the code by using isset() and also trimming the value, and checking if the strlen returned is zero.

Sounds like a belt-and-suspenders approach but this seems to work fine!

>    happy to revert, but the invalid return's have always bothered me in the controller methods disappointed_relieved so I also fixed that by using return, and annotated the params & method changed with types (I think our minimum version of PHP accepts it? if not, and CI fails, I'll revert this part)

Yes, these are excellent improvements! Adding type annotations is good whenever you are editing Skosmos code and these are not problematic - the minimum PHP version is currently 7.2 and it already has good support for type annotations (7.0 was a bit lacking IIRC).

I'm not sure why the GitHub Actions CI jobs don't seem to have run for this PR. I can try to execute them some other way.

As pointed out in [this comment](https://github.com/NatLibFi/Skosmos/issues/1260#issuecomment-1017419719), a very similar problem affects also searches done via the web UI. This PR takes care of the REST side, but not the web UI searches. So I think we need another PR to fix that, but this is already a great improvement!

osma:
I still don't understand why CI actions aren't running. I reviewed the settings for this repository and there should be an approval process for first-time contributors, but I'm not seeing the approval button...

I added a very minor typo fix comment to @kinow's PR branch, but that didn't trigger it either. Then I pushed the same branch into NatLibFi/Skosmos:kinow-fix-rest-search-zero . That triggered a [CI run](https://github.com/NatLibFi/Skosmos/actions/runs/1723312843) which finished successfully.

All seems to be good, I'm merging this PR. Thanks again @kinow!

PR #1262, opened 2022-01-05
Allow plugin robloach/component-installer, because it is required by components/handlebars.js

janvanmansum:
When running `composer install --no-dev` I get a question about whether to trust the [robloach/component-installer plug-in](https://packagist.org/packages/robloach/component-installer). In `composer.json` it is mentioned as a requirement for `components/handlebars.js`. `robloach/component-installer` is deprecated ("abandoned") but I know too little about PHP to solve this in a cleaner way. Hopefully, a new version of handlebars will use the new plug-in https://packagist.org/packages/oomphinc/composer-installers-extender.

The interactive part of the installation is incovenient because I am using Ansible to do the installation.

n/a

* Adds robloach/component-installer as an allowed plug-in to `composer.json`
* Unrelated: adds `/.idea` to `.gitignore` as a convenience for people like myself who edit the project in IntelliJ.

As far as I understand the only change that this PR applies is that the user who executes the `composer install` command is not longer asked whether to trust the `robloach/component-installer` plug-in. However, currently there is no real choice.

## Checklist

don't know how to run these, but I haven't changed any PHP code, so I would be suprised if they don't pass anymore.
not sure how to do that 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@ components
 plugins/*
 !README
 composer.phar
-composer.lock
 tests/jena-fuseki*
 .vagrant
 .project
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ tests/jena-fuseki*
 .project
 .DS_Store
 .phpunit.result.cache
+/.idea
+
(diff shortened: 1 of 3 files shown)


joelit:
Thanks for the PR! I will add this to the next release. :)

osma:
This PR was merged to `v2.13-maintenance` branch instead of `master`. New PRs should always target `master`. This needs to be reapplied so it gets into the new release.

PR #1266, opened 2022-01-20
Support PHP 8.0

### LINKED ISSUE #1243: Tests fail and CI hangs on PHP 8

Happens with GitHub Actions CI builds, for example [this build](https://github.com/NatLibFi/Skosmos/runs/4156908570) for a recent PR.

1. Open a PR or perform some other action that will trigger a GitHub Actions CI build

Expected that CI test jobs on PHP8 should either pass (preferred) or at least fail gracefully, as I think they used to do earlier when we switched to GitHub Actions. Instead they get stuck apparently at 189/469 (with a great many errors at that point) and time out after 6 hours, so the whole build gets stuck for a very long time.

osma:

Trying to get Skosmos working on PHP 8.0

- Closes #1243

* upgrade willdurand/negotiation to 3.0.* which supports PHP8
* reenable GitHub Actions CI jobs for PHP 8.0
* upgrade php-actions/phpunit to v3 which is slightly newer than what we used before

* CI jobs for PHP 8.0 seem to be hanging (again, see #1243) so creating this as a draft PR for now

## Checklist

No new tests are necessary since there is no new functionality.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on: [push]
 jobs:
   build-test:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     continue-on-error: ${{ matrix.experimental }}
     strategy:
       fail-fast: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php_version: [7.2, 7.3, 7.4]
+        php_version: [7.2, 7.3, 7.4, 8.0]
         experimental: [false]
 
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,10 @@ jobs:
         CC_TEST_REPORTER_ID: fb98170a5c7ea9cc2bbab19ff26268335e6a11a4f8267ca935e5e8ff4624886c
       with:
         prefix: /app
+      # PHPUnit 8.5 doesn't support code coverage for PHP 8
+      if: ${{ matrix.php_version < "8.0" }}
 
     - name: Publish code coverage to Codecov
       uses: codecov/codecov-action@v1
+      # PHPUnit 8.5 doesn't support code coverage for PHP 8
+      if: ${{ matrix.php_version < "8.0" }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,6 @@ jobs:
         php_extensions: gettext intl xsl pcov
         memory_limit: 512M
         configuration: phpunit.xml
-        args: --verbose --debug
 
     - name: Publish code coverage to Code Climate
       uses: paambaati/codeclimate-action@v2.7.5
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
         php_extensions: gettext intl xsl pcov
 
     - name: Run PHPUnit tests
-      uses: php-actions/phpunit@91ff02a58932e63741157ea8c04e7e7077233537
+      uses: php-actions/phpunit@v3
       env:
         LANGUAGE: fr
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,9 @@ jobs:
       matrix:
         php_version: ["7.2", "7.3", "7.4", "8.0"]
         experimental: [false]
+        include:
+          - php_version: "8.1"
+            experimental: true
 
     steps:
     - name: Check out repository code
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php_version: [7.2, 7.3, 7.4, 8.0]
+        php_version: [7.2, 7.3, 7.4, 8.0.15]
         experimental: [false]
 
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,9 +61,9 @@ jobs:
       with:
         prefix: /app
       # PHPUnit 8.5 doesn't support code coverage for PHP 8
-      if: ${{ matrix.php_version < "8.0" }}
+      if: ${{ matrix.php_version < '8.0' }}
 
     - name: Publish code coverage to Codecov
       uses: codecov/codecov-action@v1
       # PHPUnit 8.5 doesn't support code coverage for PHP 8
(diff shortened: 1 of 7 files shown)


osma:
This is quite difficult to untangle - there are many problems here and some unfortunate interactions between them that make debugging very hard. Things I've found out so far:

- The PHPUnit configuration under GitHub Actions CI is not quite the same as when running PHPUnit locally. The CI environment wasn't using the `phpunit.xml` configuration file (now fixed in this PR) and also the `error_reporting` settings seem to be different. Basically, under CI the deprectation errors (E_DEPRECATED) were breaking tests, but this didn't happen when running test locally. These should be aligned - either deprecations should cause test failures everywhere (preferred) or in neither environment, if we decide so. Probably `error_reporting` should be set in `tests/bootstrap.php` so that it will be applied every time PHPUnit is run, regardless of environment.
- there are some deprecation errors in the Skosmos codebase, as well as in the Mockery library, because PHP 8.0 [deprecates the use of required parameters after optional parameters](https://php.watch/versions/8.0/deprecate-required-param-after-optional). These should probably be fixed in Skosmos code and the dependencies should be upgraded if possible.
- The CI environment was using an old version of PHP (8.0.6) because it had created the php-build images (for Composer and PHPUnit separately) in May 2021 and cached them as packages in the Skosmos repository. I just deleted the PHP 8 ones (but not php 7.x which are similarly old), to force the use of a newer PHP. Locally I'm using PHP 8.0.14 and 8.0.15 seems to be the newest available 8.0 release. **Update**: Now the CI environment seems to be using PHP 8.1.1, which is not what I wanted at all! The `ci.yml` requests `8.0` which should use the newest available 8.0.* version.
- The Http304 tests (which use ob_start) seem to be the ones getting stuck under CI. I disabled them (by skipping in setUp) in this PR for now.

osma:
> Now the CI environment seems to be using PHP 8.1.1, which is not what I wanted at all! The ci.yml requests 8.0 which should use the newest available 8.0.* version.

This can be fixed by requesting a more specific version such as 8.0.15, [see here](https://github.com/php-actions/composer/issues/73)

osma:
I think I've fixed most of the above:

- PHP version numbers specified with quotes got me the right PHP 8.0 version (YAML parsing issue)
- tests are running with the same error_reporting and deprecation policy (strict!) both locally and in CI
- fixed a few cases of deprecated method parameters in Skosmos code
- upgraded Mockery and DomCrawler to versions that work on PHP 8.0 (but also still support 7.2)
- reenabled Http304 tests that were no longer getting stuck (but I never found out why)
- disabled debug output for phpunit

The PHPUnit step is now passing under PHP 8.0 CI :tada: - but there are still problems with the Code Climate test reporter on PHP 8.0, which causes CI jobs to fail.  I'll leave sorting that out for later!

osma:
I disabled Code Climate and Codecov reporters for PHP 8, since PHPUnit 8.5 won't collect coverage information for PHP 8 anyway. Now the PHP 8.0 CI jobs are succesful :tada: 

I also added PHP 8.1 CI tests in experimental mode (permitted to fail). There are a lot of deprecation warnings causing nearly all tests to fail on PHP 8.1. The main culprit seems to be EasyRdf, which isn't yet compatible with PHP 8.1, but there's an [open PR](https://github.com/easyrdf/easyrdf/pull/384) which should hopefully fix this before long.

osma:
All seems well now looking at the test results, but I still need to test actually **running** Skosmos on PHP 8.0, because test coverage is far from 100%. That might uncover some more problems.

kinow:
Tried running with Docker but something went wrong when starting the server.

```diff
diff --git a/dockerfiles/Dockerfile.ubuntu b/dockerfiles/Dockerfile.ubuntu
index d834ce42..cdf6600c 100644
--- a/dockerfiles/Dockerfile.ubuntu
+++ b/dockerfiles/Dockerfile.ubuntu
@@ -8,6 +8,8 @@ ARG SKOSMOS_VERSION=v.2.9
 
 ARG DEBIAN_FRONTEND=noninteractive
 
+RUN apt-get update && apt-get install -y software-properties-common && add-apt-repository -y ppa:ondrej/php
+
 # git is necessary for some composer packages e.g. davidstutz/bootstrap-multiselect
 # gettext is necessary as php-gettext was available in 18.04, but not in 20.04
 RUN apt-get update && apt-get install -y \
@@ -15,13 +17,13 @@ RUN apt-get update && apt-get install -y \
     curl \
     gettext \
     git \
-    libapache2-mod-php7.4 \
+    libapache2-mod-php8.0 \
     locales \
-    php7.4 \
-    php7.4-curl \
-    php7.4-xsl \
-    php7.4-intl \
-    php7.4-mbstring \
+    php8.0 \
+    php8.0-curl \
+    php8.0-xsl \
+    php8.0-intl \
+    php8.0-mbstring \
     php-apcu \
     php-zip \
     unzip \
@@ -55,7 +57,7 @@ ENV LC_ALL=en_US.UTF-8
 ENV LANG=en_US.UTF-8  
 
 # timezone
-RUN sed -i 's/;date.timezone =/date.timezone = "UTC"/g' /etc/php/7.4/apache2/php.ini
+RUN sed -i 's/;date.timezone =/date.timezone = "UTC"/g' /etc/php/8.0/apache2/php.ini
 
 COPY dockerfiles/config/000-default.conf /etc/apache2/sites-available/000-default.conf
 

```

osma:
@kinow Thanks for testing! It seems to me based on the error messages that your Composer was running under PHP 8.1.1, not 8.0.x. Also it complained a lot about missing PHP extensions even though you clearly had those installed.

My guess is that installing `php-apcu` and `php-zip` will install PHP 8.1.1 and make that the default PHP version. Switching to `php80-apcu` and `php80-zip` should help in that case.

kinow:
>My guess is that installing php-apcu and php-zip will install PHP 8.1.1 and make that the default PHP version. Switching to php8.0-apcu and php8.0-zip should help in that case.

You are right @osma. Changed that, and now I have Skosmos and PHP 8.0 :-) :tada: 

image

osma:
You beat me to it @kinow ! Do tell if you find problems. I will also try running Skosmos itself under PHP 8.0 (not just the tests), but I haven't yet found the time to do it.

osma:
I tested this briefly on a VM running PHP 8.0.15. Everything seems to be working correctly and there are no suspicious warnings in the Apache error log. So I think it's pretty safe to merge this - if new problems turn up later we can deal with them separately.

osma:
@kinow Now that we have PHP 8.0 support for Skosmos, should the Dockerfile be updated to use PHP 8.0 as well, the way you already tested it? Care to open a PR?

kinow:
> @kinow Now that we have PHP 8.0 support for Skosmos, should the Dockerfile be updated to use PHP 8.0 as well, the way you already tested it? Care to open a PR?

Good idea, will try to send a quick PR this week. Excited about learning php8 with skosmos! Thanks Osma!

PR #1267, opened 2022-01-26
Fix vocabulary search using "0" as the search string

### LINKED ISSUE #1260: Vocabulary search using "0" as the query argument throws an error

miguelvaara:

Fix the remaining cases in #1260 ; the REST API search using "0" has been fixed in #1261 and #1276, but searches for "0" in the web UI are still not working until this PR.
 
- Closes #1260 

- fix ConceptSearchParameters.getSearchTerm() so it looks at the right URL parameter `q` even when its value is `"0"`
- rename the test provider method `testSearchWithZeroData` into `provideSearchWithZeroData` to avoid PHPUnit considering it a test case and complaining that it's "Risky" because it doesn't perform any assertions

n/a

## Checklist


osma:
Can you please update the PR title and description so it corresponds to current status?

kinow:
Hadn't seen this PR, I think it's also related to the zeroes in the beginning of the search, sorry! Feel free to close mine and centralize on this one if you prefer, or use anything from that PR too :+1:

osma:
Need to fix the conflicts caused by #1276.

osma:
I fixed the conflicts, cleaned up the PR (removed useless test data), edited the PR title, rewrote the PR description, and tested this. I think it should be good to go now. Will wait for QA jobs, do a final check, then merge if there are no problems.

PR #1272, opened 2022-02-15
Update Dockerfile to support PHP 8

kinow:

Support PHP 8 when using the test Docker image.

- Complements #1266 

Used the same Dockerfile changes I used to test the linked pull request.

There are existing images that support PHP 8 and produce smaller containers, but I think it would be better to move that work for another issue.

## Checklist


osma:
Thanks for the PR!

Now that I see the changes, I'm not sure this is the best way to support PHP 8. Basically this is taking an Ubuntu 20.04 base image (which by default has PHP 7.4 packages), then activating the `ondrej/php` repository and installing PHP 8.0 packages from that repo. I think it would be better to use a base image that directly provides PHP 8.0 packages. You mentioned this above:

> There are existing images that support PHP 8 and produce smaller containers, but I think it would be better to move that work for another issue.

Do you have any specifics on this - what base image alternatives are there?

We could also wait for Ubuntu 22.04, which is due in a month from now. But it will have PHP 8.1, which is not yet supported in Skosmos although we have a related open issues #1274 and #1291.

kinow:
>>There are existing images that support PHP 8 and produce smaller containers, but I think it would be better to move that work for another issue.
>>
>Do you have any specifics on this - what base image alternatives are there?

We could probably use some small image like `php:8.1.4-fpm-alpine3.15`. Although I had a look at the compressed size, and both this PHP 8 and Ubuntu 20.04 have similar compressed size. Interesting.

Nonetheless, we could use the PHP official images, with the extra things required for Skosmos, instead of using Ubuntu as base image. The Docker file would be smaller, and I think it would be easier to upgrade to newer versions of PHP, as long as there was a tag in the official PHP image.

From what I recall, I did not use the PHP images because I used our installation instructions as reference, and wanted to stay as close as possible to that documentation (although my memory often fails me, but I think that's the reason for using Ubuntu.)

>We could also wait for Ubuntu 22.04, which is due in a month from now. But it will have PHP 8.1, which is not yet supported in Skosmos although we have a related open issues https://github.com/NatLibFi/Skosmos/issues/1274 and https://github.com/NatLibFi/Skosmos/issues/1291.

This sounds like an easy fix, unless we have the need to upgrade the Docker files immediately. I think I only created this PR because I had it from some tests for Skosmos + PHP 8 after you fixed the build for this newer version of PHP.

Thanks @osma!

Bruno

kinow:
@osma, now using the official php Docker image that ships with Apache httpd. The `composer` is being fetched from the official Composer image too.

If you start `docker-compose` the UI should come up, but there is a warning in the cache image. That is fixed in #1294 :+1: 

-Bruno

PR #1276, opened 2022-02-17
Do not treat 0 as boolean value in the REST controller

### LINKED ISSUE #1275: Searching via API using the search term '0' without asterisk (0*) fails
https://api.dev.finto.fi/rest/v1/ykl/search?query=0

1. Go https://api.finto.fi/rest/v1/ykl/search?query=0 

It turned out that the fixes in the PR #1261 do not work completely (mostly it works :+1: ). There is still a problem with the searches with zeros ('0') if you do not use a trailing asterisk with the number zero. 

[https://api.finto.fi/rest/v1/ykl/search?query=0](https://api.finto.fi/rest/v1/ykl/search?query=0) should return a result containing the concept [https://dev.finto.fi/ykl/en/page/0](https://dev.finto.fi/ykl/en/page/0) (http://urn.fi/URN:NBN:fi:au:ykl:0) but it returns an empty result. 

FF

kinow:

Prevent `0` being treated as a truthy value in the REST controller. Complements #1261 

- Closes #1275

Same fix as #1261 

None.

## Checklist

### DIFF
--- a/tests/ConceptSearchParametersTest.php
+++ b/tests/ConceptSearchParametersTest.php
@@ -87,6 +87,26 @@ class ConceptSearchParametersTest extends PHPUnit\Framework\TestCase
     }
 
     /**
+     * For https://github.com/NatLibFi/Skosmos/issues/1275, to verify
+     * that querying for `0` (zero) does not evaluate it as a boolean
+     * value, causing issues in the SKOSMOS search functionality.
+     *
+     * @covers ConceptSearchParameters::getSearchTerm
+     */
+    public function testGetSearchTermZeroes() {
+        $params = new ConceptSearchParameters($this->request, new GlobalConfig('/../tests/testconfig.ttl'), true);
+        $this->assertEquals('', $params->getSearchTerm());
+        $this->request->method('getQueryParamRaw')->will(
+            $this->returnValueMap([
+                ['q', '0'],
+                ['query', '0'],
+                ['label', '10']
+            ])
+        );
+        $this->assertEquals('0', $params->getSearchTerm());
+    }
+
+    /**
      * @covers ConceptSearchParameters::getTypeLimit
      * @covers ConceptSearchParameters::getDefaultTypeLimit
      */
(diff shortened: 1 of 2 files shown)


osma:
Can you please rebase @kinow, so we can see if #1277 worked?

osma:
This is now overlapping with #1267 by @miguelvaara  - there are changes to the same lines in ConceptSearchParameters.getSearchTerm. These need to be consolidated somehow. Maybe simply by merging one first, then fixing the conflicts in the other.

kinow:
Rebased, GH Actions working :toolbox: thanks @osma !

osma:
The code looks good - I think the easiest way forward is to merge this first and then sort out the conflicts in PR #1267, as that one is still a bit WIP (at least the PR description is outdated).

The only thing I wonder about is the title of this PR (and the commit message) which says "Do not treat 0 as truthy value". Isn't the point actually the opposite - "Do not treat 0 as a **non-truthy** (i.e. false) value"?

kinow:
You're right! s/truthy/boolean , WDYT?

osma:
Sounds good!

tests/ConceptSearchParametersTest.php
```diff
@@ -86,6 +86,26 @@ public function testGetSearchTerm() {
         $this->assertEquals('test', $params->getSearchTerm());
     }
 
+    /**
+     * For https://github.com/NatLibFi/Skosmos/issues/1275, to verify
+     * that querying for `0` (zero) does not evaluate it as a truthy
```
osma, line 91: s/truthy/boolean/ here as well?

kinow: Oh, missed that one. Just a sec.

kinow: Done. That should be the last one ™

osma: Famous last words!

osma:
Will do one final test round and then merge!

osma:
Works like a charm! Thanks a lot @kinow !

PR #1279, opened 2022-03-02
Revert 1262 v2.13 maintenance

joelit:

PR #1262 was merged in the maintenance branch needlessly, should be reverted.


osma:
Wait, why are you merging these changes to master? The commits should be reverted on the *maintenance branch*!

osma:
I mean, this would just add clutter to the history master branch (first make changes, then revert them) but otherwise leave everything unchanged... The diff is 0 lines.

PR #1285, opened 2022-03-02
Issue1148 order of plugin rendering

### LINKED ISSUE #1148: Order of vocabulary-specific plugins is not configurable

Vainonen:
Fixes reopened issue #1148. Orders also callback names of plugins in the same order as sorted in VocabularyConfig.php and configured in skosmos:vocabularyPlugins.

- Closes #1148

Plugin order is configured with skosmos:vocabularyPlugins. PR #1201 made it possible to define order of plugins in configuration. PluginRegister class reorders the plugins in alphabetical order of plugin file names, but this fix rearranges the callback names to the same order as in skosmos:vocabularyPlugins. PR also adds the possibility to define parameter plugins outside of the ordered plugin list

Plugin rendering order is reverse than the order configured in skosmos:vocabularyPlugins. To solve this, Skosmos HTML template could have additional empty div element (e.g. `<div id="concept-info-after"></div>`), which serves as a location for widgets. Each widget could then reserve own empty div element during callback with JQuery `$('#concept-info-after').append(...)`. Then the rendering would be done in same order as in skosmos:vocabularyPlugins.

In same way a similar div element (e.g. `<div id="concept-info-before">`) could be added for widgets with message (e.g. https://github.com/NatLibFi/Skosmos-widget-message) above the concept info.

### DIFF
--- a/model/PluginRegister.php
+++ b/model/PluginRegister.php
@@ -5,6 +5,10 @@ class PluginRegister {
 
     public function __construct($requestedPlugins=array()) {
         $this->requestedPlugins = $requestedPlugins;
+        $this->pluginOrder = array();
+        foreach ($this->requestedPlugins as $index => $value) {
+            $this->pluginOrder[$value] = $index;
+        }
     }
 
     /**
@@ -25,6 +29,16 @@ class PluginRegister {
     }
 
     /**
+     * Sort plugins by order defined in class constructor
+     * @return array
+     */
+    private function sortPlugins($plugins) {
+        $sortedPlugins = array();
+        $sortedPlugins = array_replace($this->pluginOrder, $plugins);
+        return $sortedPlugins;
+    }
+
+    /**
      * Returns the plugin configurations found from plugin folders
      * inside the plugins folder filtered by filetype.
      * @param string $type filetype e.g. 'css', 'js' or 'template'
@@ -33,6 +47,7 @@ class PluginRegister {
      */
     private function filterPlugins($type, $raw=false) {
         $plugins = $this->getPlugins();
+        $plugins = $this->sortPlugins($plugins);
         $ret = array();
         if (!empty($plugins)) {
             foreach ($plugins as $name => $files) {
@@ -150,10 +165,6 @@ class PluginRegister {
     public function getCallbacks() {
         $ret = array();
         $sortedCallbacks = array();
-        $order = array();
-        foreach ($this->requestedPlugins as $index => $value) {
-            $order[$value] = $index;
-        }
         $plugins = $this->getPluginCallbacks($this->requestedPlugins);
         foreach ($plugins as $callbacks) {
             foreach ($callbacks as $callback) {
@@ -161,7 +172,7 @@ class PluginRegister {
                 $sortedCallbacks[$split[1]] = $split[2];
             }
         }
-        $sortedCallbacks = array_replace($order, $sortedCallbacks);
+        $sortedCallbacks = array_replace($this->pluginOrder, $sortedCallbacks);
         foreach ($sortedCallbacks as $callback) {
             $ret[] = $callback;
         }
--- a/model/PluginRegister.php
+++ b/model/PluginRegister.php
@@ -134,7 +134,7 @@ class PluginRegister {
      * @param array $names the plugin name strings (foldernames) in an array
      * @return array
      */
-    public function getPluginsCallbacks($names=null) {
+    public function getPluginCallbacks($names=null) {
         if ($names) {
             $names = array_merge($this->requestedPlugins, $names);
             return $this->filterPluginsByName('callback', $names);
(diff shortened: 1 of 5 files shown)


osma (changes requested):
Can you please use the [PR template](https://github.com/NatLibFi/Skosmos/blob/master/PULL_REQUEST_TEMPLATE.md) to describe the changes in this PR?

Also, it would be helpful to provide a minimal example of how to test this. For example a test configuration with 2-3 plugins that were in the wrong order before but are correctly ordered with the changes in this PR.

model/PluginRegister.php
```diff
@@ -130,21 +130,44 @@ public function getTemplates($names=null) {
     }
 
     /**
-     * Returns an array of javascript function names to call when loading pages
+     * Returns an array of plugin callback function names
+     * @param array $names the plugin name strings (foldernames) in an array
+     * @return array
+     */
+    public function getPluginCallbacks($names=null) {
+        if ($names) {
```
osma, line 138: The tests don't cover the case of $names being passed. I think it should be easy to modify the test for getPluginCallbacks. I will suggest below...

tests/PluginRegisterTest.php
```diff
@@ -71,9 +108,21 @@ public function testGetPluginsCSS()
...
+  /**
+   * @covers PluginRegister::getPluginCallbacks
+   * @covers PluginRegister::filterPlugins
+   * @covers PluginRegister::filterPluginsByName
+   */
+  public function testGetPluginCallbacks()
   {
     $plugins = new PluginRegister();
-    $this->assertEquals(array('test-plugin' => array('plugins/test-plugin/stylesheet.css')), $this->mockpr->getPluginsCSS(array('test-plugin')));
+    $this->assertEquals($this->mockpr->getPluginCallbacks()['test-plugin1'],
```
osma, line 124: Here passing a $names parameter to getPluginCallbacks would increase test coverage.

osma (changes requested):
The PR introduction is now good. Here are some suggestions for changes:

- if it's easy, make the tests cover the case of non-null $names parameter for getPluginCallbacks (see suggestions in comments)
- the `<script>` elements for plugins in HTML code should be in the configuration order of plugins
- there should be `concept-info-before` and `concept-info-after` elements in `concept-shared.twig`, like this:

Around line 15:
```
    <div id="concept-info-before"></div>
    <div class="concept-info{% if concept.deprecated %} deprecated-concept{% endif %}">
```

Around line 211:

```
      </div>
    </div>
    <div id="concept-info-after"></div>
    {% endfor %}
```

osma:
The Javascript execution order is now good.

The new `<div>` elements in the Twig template seem to be working as well. I was able to create a set of three plugins which insert themselves into the provided areas.

I added some annotations to increase test coverage. The case of non-empty `$names` parameter is still not covered, but I think we can live with that.

The documentation for plugins still needs to be updated. I think we need updated guidelines on how plugins should insert their elements into the DOM structure, namely:

1. The callback function of the plugin SHOULD append (insert as last) an element for the plugin into one of the provided DIVs (or elsewhere on the page). It can be an empty placeholder DIV with an id that identifies it to the plugin. For example like this: `$('#concept-info-after').append('<div id="my-widget"></div>');`
2. If the plugin performs async operations, for example calls an external API, it can update its DIV once the API call has completed and all information is available. As a minimal example, they can call `$('#my-widget').text(results);` where `results` is something that was returned by the external API call.

This way the order of the inserted elements remains the same as the configured order of plugins (which is the order for the function calls to the plugin callbacks). It doesn't matter how long async operations take, because the order has been established by the placeholder elements and won't change even if they are updated.

I will merge this PR as it does its job; updating the [documentation for plugins](https://github.com/NatLibFi/Skosmos/wiki/Plugins) still remains as a task.

PR #1287, opened 2022-03-10
Replace #concept-info-before div element with more generic #content-top

osma:

In PR #1285, new empty DIV elements `#concept-info-before` and `#concept-info-after` were added as placeholders so that plugins can add widgets into these predefined slots. But these are specific to the concept page and not available on other page types such as the vocabulary home page. This PR replaces those DIVs with more generic ones `#content-top` and `#content-bottom` which can be used on all (?) page types.

- Follow-up fix to #1285

* remove `#concept-info-before` and `#concept-info-after` DIV elements
* add `#content-top` and `#content-bottom` DIV elements
* some reindentation

n/a

## Checklist

Tests are not relevant since only Twig templates were changed.

### DIFF
--- a/view/concept-shared.twig
+++ b/view/concept-shared.twig
@@ -208,7 +208,6 @@
           >
       </div>
     </div>
-    <div id="concept-info-after"></div>
     {% endfor %}
 
   {% else %}
--- a/view/concept-shared.twig
+++ b/view/concept-shared.twig
@@ -12,7 +12,6 @@
       <p class="language-alert">{% trans %}There is no term for this concept in this language{% endtrans %}</p>
     </div>
     {% endif %}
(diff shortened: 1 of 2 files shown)


osma:
Tests on PHP 7.x are failing for some reason that is clearly unrelated to the changes in this PR:

```
PHPUnit\Framework\Exception: Fatal error: Uncaught TypeError: Argument 3 passed to PHP_Token::__construct() must be an instance of PHP_Token_Stream, instance of PHPUnit\PHP_Token_Stream given, called in phar:///usr/local/bin/phpunit/php-token-stream/Token/Stream.php on line 131 and defined in /app/vendor/phpunit/php-token-stream/src/Token.php:37
```

joelit (approved):

PR #1293, opened 2022-03-25
drop useless SKOSMOS_VERSION build arg from Dockerfile

osma:

I noticed that the Dockerfile for Skosmos declares a build argument SKOSMOS_VERSION which isn't actually used and the default value `v.2.9` refers to an old version of Skosmos. This PR simply removes it. 

n/a

Drop the line with the SKOSMOS_VERSION declaration in Dockerfile

n/a

## Checklist


kinow (approved):
I think it was used at some stage, to fetch a version of Skosmos to build, but shouldn't be used any longer.

CI errors appear to be unrelated. +1 Thanks!

PR #1303, opened 2022-04-25
Fix typeahead and handlebars template.

### LINKED ISSUE #1301: Autocomplete problems after Bootstrap 5 upgrade

https://dev.finto.fi/yso/en/

1. Search for e.g. "cat" in YSO and wait for the autocomplete popup

This is how it works in finto.fi, running Skosmos 2.14:

kuva

Here is how it looks with current master branch (dev.finto.fi):

kuva

There are at least three problems:

1. The entered string "cat" is shown with the wrong color. Minor CSS issue, probably in the Finto CSS file.
2. Matches on altLabels (e.g. "Catalan people") don't show the prefLabel of the matched concept ("Catalans").
3. The concept type information is missing.

This most likely broke with the Bootstrap 5 upgrade (PR #1182), which included an upgrade of the Typeahead.js widget.
Ping @kinow 

Firefox

kinow:

Fix styles and UI behavior, broken in #1182 

- Closes #1301 

I noticed that the data was present for Typeahead to display it, but for some reason we only saw part of the data when values were replaced (e.g. Catalans & Catalan People). I couldn't find anything in the Typeahead docs, but then it occurred to me that it could be now handling templates the same way Vue.js does, by forcing to have a single root-node.

I moved things under a `<div>` in the template rendered for each autocomplete entry, and then it finally display all the information received and used in the template.

1. I left a question in #1301 asking about the color of the text. This is a 1 minute change, just pending a reply over there :+1: 

2. @osma I couldn't understand why my types are not being replaced by shorter text. I wonder if it's because I have been using old vocabularies I exported from finto and/or finto.dev?

I set a few breakpoints, and after clearing my local storage, I can confirm the UI is sending a request to `http://localhost:9090/rest/v1/types?lang=en` to retrieve types, and is also using the `@context` returned from the autocomplete request.

But the right-most information is still using the URI instead of a value like "General Concept" as in your screenshot attached to #1301. See my screenshot below for an example:

Screenshot from 2022-04-25 17-18-55

It shows skos:Concept, and other URI's instead of a simpler text :disappointed: . Thanks!

## Checklist

### DIFF
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -159,6 +159,7 @@ a, a.versal, .nav-link, .nav-link.active, .jstree-node > .jstree-anchor {
 
 .replaced {
   font-style: italic;
+  color: var(--gray-900) !important;
 }
 
 .nav-link:link,.nav-link:visited,.nav-link:active,.nav-link:hover,.nav-link:focus {
@@ -2073,9 +2074,8 @@ span.date-info {
   overflow: hidden;
 }
 
-p.autocomplete-label {
+.autocomplete-label > span {
   color: var(--medium-color);
-  display: block;
   margin: 0;
 }
 
--- a/resource/js/docready.js
+++ b/resource/js/docready.js
@@ -749,11 +749,13 @@ $(function() { // DOCUMENT READY
   concepts.initialize();
 
   var autocompleteTemplate =[
-    '{{# if matched }}<p>{{matched}}{{# if lang}} ({{lang}}){{/if}} = </p>{{/if}}',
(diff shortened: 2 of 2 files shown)


osma:
Thanks @kinow , I will take a closer look!

No worries about the types showing up as URIs, that's a recurring problem. I fixed one instance of this recently (see #1254 ) but this is a common occurrence. Apparently the mechanism we have for looking up labels is a bit fragile.

osma:
Using the code from this PR, the type labels are working for me:

kuva

I think you have an unrelated problem with those.

I tested this and everything seems to be working, so I will merge this PR. It seems to me there are still some minor font style issues (color was already mentioned, but font sizes are also a bit different from before) but those can be handled in separate PRs.

Issue #1289, opened 2022-03-11
Possible open redirect


Amixtika:
Go to: 

https://finto.fi/https://example.com

Select: "In English" on the top right, or any other language.

After that you will be redirected to www.example.com.

osma:
Thanks for the issue.

I don't see this as an open redirect, as it requires the user to click on a link to https://example.com - but maybe I misunderstood something?

Amixtika:
Hi Osma, 

The potential issue is that a user can be instructed to click on this link,  redirecting the user to a external malicous page.

PR #1307, opened 2022-04-26
Sanitize language switching URLs

### LINKED ISSUE #289: YSO concept pages broken
http://dev.finto.fi/yso/fi/page/p20945

1. Open YSO concept page, e.g. above URL

Getting blank page. Error in log:

[Tue Sep 15 15:45:32 2015] [error] [client 128.214.223.217] PHP Fatal error:  Uncaught exception 'Exception' with message 'Vocabulary id 'Allfo' not found in configuration.' in /var/www/dev.finto.fi/model/Model.php:416\nStack trace:\n#0 /var/www/dev.finto.fi/controller/WebController.php(73): Model->getVocabulary('Allfo')\n#1 [internal function]: {closure}('http://www.yso....', 'Allfo', 'sv', 'page', 'sv')\n#2 /tmp/onki-light-template-cache/84/84/20176dff2d539bfe46efe0ae2ca3.php(898): call_user_func_array(Object(Closure), Array)\n#3 /var/www/dev.finto.fi/vendor/twig/twig/lib/Twig/Template.php(264): __TwigTemplate_848420176dff2d539bfe46efe0ae2ca3->doDisplay(Array, Array)\n#4 /var/www/dev.finto.fi/vendor/twig/twig/lib/Twig/Template.php(238): Twig_Template->displayWithErrorHandling(Array, Array)\n#5 /tmp/onki-light-template-cache/78/be/e3bde6a7268c7c323114145a0c0e.php(59): Twig_Template->display(Array)\n#6 [internal function]: __TwigTemplate_78bee3bde6a7268c7c323114145a0c0e->block_content(Array, Array)\n#7 /var/www/dev.finto.fi/vendor/twig/twig/lib/Twig/Template.php(133): call_user_func(Array, Arra in /var/www/dev.finto.fi/vendor/twig/twig/lib/Twig/Template.php on line 279

osma:

See issue #1289 - it was possible to trick Skosmos to generate links to an arbitrary URL. This could be a potential security issue together with some social engineering (instruct users to click on the link).

- Closes #289

1. add some more unit tests for Request->getLangUrl method (to make sure all relevant cases are covered)
2. sanitize the URL string returned by Request->getLangUrl so it won't contain special characters 

The sanitizing will cause the "smuggled" URLs to stop working, for example `http://example.com` becomes `http//examplecom` and thus the links won't work anymore.

The language switching links on the 404 page generated for a page such as `/Skosmos/http://example.com` will be broken. It's no worse than before though, since they were already broken, now they are just less harmful.

## Checklist

### DIFF
--- a/model/Request.php
+++ b/model/Request.php
@@ -185,7 +185,7 @@ class Request
             $langurl = preg_replace("#^(.*/)?{$this->lang}/#", "$1{$newlang}/", $langurl);
         }
         // make sure that the resulting URL doesn't contain suspicious characters
-        $langurl = preg_replace("#[^a-zA-Z0-9-/]#", "", $langurl);
+        $langurl = preg_replace("#[^a-zA-Z0-9/-]#", "", $langurl);
         return $langurl;
     }
 
--- a/model/Request.php
+++ b/model/Request.php
@@ -179,10 +179,13 @@ class Request
      */
     public function getLangUrl($newlang=null)
     {
-        $langurl = substr(str_replace(str_replace('/index.php', '', $this->getServerConstant('SCRIPT_NAME')), '', $this->getServerConstant('REQUEST_URI')), 1);
+        $script_name = str_replace('/index.php', '', $this->getServerConstant('SCRIPT_NAME'));
+        $langurl = substr(str_replace($script_name, '', $this->getServerConstant('REQUEST_URI')), 1);
         if ($newlang !== null) {
             $langurl = preg_replace("#^(.*/)?{$this->lang}/#", "$1{$newlang}/", $langurl);
         }
+        // make sure that the resulting URL doesn't contain suspicious characters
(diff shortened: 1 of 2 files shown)


osma:
Note that this may restrict the range of available characters in vocabulary IDs. Currently only `[a-zA-Z0-9/-]` are allowed.

osma:
The last commit changed the sanitizing. Now it's more minimal and only strips colons from the URL, preventing the smuggling of  absolute URLs. `http://example.com` becomes `http//example.com`. This has a minimal effect on vocabulary IDs since even before, it was in practice impossible to use colons in vocabulary IDs, they wouldn't work anyway.

joelit (approved):
LGTM - better to just break absolute URLs this way (instead of limiting the possibility of using non-alphanumeric characters in vocabulary IDs).

PR #1308, opened 2022-04-26
Store current Skosmos version in composer.json

### LINKED ISSUE #1202: Skosmos version not showing in Docker install
1. Run a Skosmos instance with the latest Docker install
2. Go to "About"

## What do you see?
Capture d’écran 2021-09-10 à 11 55 46

In the Dockerfile, it is written `ARG SKOSMOS_VERSION=v.2.9`

Is it the version used for the latest Docker install ?

osma:

Skosmos has been using git tags to infer the current version, which is shown on the About page and in the Generator field in the embedded HTML metadata. This has never worked on Docker containers (which don't have access to the full git repository) and recently it has been broken also in some regular installs.

This PR implements another mechanism, suggested by @kinow in #1202, where the version number is kept in `composer.json` and read via the Composer runtime API. This ensures that the version information is always available. On the downside, it requires some manual housekeeping to make sure that the version in `composer.json` is up to date. For now, I think we will do this manually. There are tools that can automate this (see https://github.com/NatLibFi/Skosmos/issues/1202#issuecomment-1105079666) but they didn't look very promising. I will add the necessary steps to the [Release Process](https://github.com/NatLibFi/Skosmos/wiki/Release-Process) documentation. I also added a unit test that verifies that the version from composer.json matches the version determined from git tags.

- Closes #1202

* add version number (here, `2.15-dev` - I don't think the `v` prefix is necessary) to `composer.json`
* use Composer runtime API to access the version number, instead of git operations
* add a unit test that verifies that the version from composer.json matches version determined using git tags
* adjust GitHub Actions configuration so that git version history (including tags) is checked out from the repo, so that the above test has a chance to succeed

I hope the new processes around release will work out well, but we'll find out soon.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,9 @@ jobs:
 
     steps:
     - name: Check out repository code
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
 
     - name: Cache Fuseki installation
       uses: actions/cache@v2
--- a/composer.json
+++ b/composer.json
@@ -3,6 +3,7 @@
   "description": "Thesaurus and controlled vocabulary browser using SKOS and SPARQL",
   "type": "project",
   "license": "MIT",
+  "version": "2.15-dev",
   "repositories": { 
     "lscache": {
       "type": "package",
--- a/model/Model.php
+++ b/model/Model.php
@@ -76,11 +76,7 @@ class Model
      */
     public function getVersion() : string
     {
-        $ver = null;
-        if (file_exists('.git')) {
-            $ver = rtrim(shell_exec('git describe --tags --always'));
-        }
-
+        $ver = \Composer\InstalledVersions::getRootPackage()['pretty_version'];
         if ($ver === null) {
             return "unknown";
         }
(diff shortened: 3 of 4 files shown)


osma:
I updated the Release Process documentation, see [diff](https://github.com/NatLibFi/Skosmos/wiki/Release-Process/_compare/e0d32df34e867bf1f0c6b5633463b758806cd8e4...5c3f9f549996a0c516d00191092147f8f804aeac)

kinow (approved):
Updated release docs look good, change looks good, tested with Docker Compose locally, all good! :tada: 

image

I think great part of the release process can be automated in GH Actions, another dev did an amazing work at Cylc (previous $work) using plain GH Actions [^1] and some custom actions (Node.js) [^2]. It can even generate PR's to be reviewed and include things like a Release Message [^3]. For some day.

[^1]: https://github.com/cylc/cylc-flow/blob/master/.github/workflows/1_create_release_pr.yml & https://github.com/cylc/cylc-flow/blob/master/.github/workflows/2_auto_publish_release.yml
[^2]: https://github.com/cylc/release-actions
[^3]: https://github.com/cylc/cylc-flow/pull/4753

osma:
Thanks for the review @kinow ! I will merge this.

Thanks also for the tips on release automation, this is very helpful and interesting. I'll open another issue to track those ideas. We've had similar discussions about automating parts of the release process, but for now I think it's best to do that manually.

PR #1312, opened 2022-04-27
Upgrade to PHPUnit 9, drop PHP 7.2 support

### LINKED ISSUE #1291: Drop support for PHP 7.2, upgrade to PHPUnit 9
Support for PHP 8.0 was added in PR #1266, released as part of Skosmos 2.14. Now Skosmos supports four versions of PHP: 7.2, 7.3, 7.4 and 8.0. (There are still problems with PHP 8.1, see #1274)

The next logical step would be to drop support for PHP 7.2, which has [been in EOL status by upstream PHP](https://www.php.net/supported-versions.php) since late 2020, although it is still supported by distributions such as Ubuntu 18.04. Dropping PHP 7.2 would allow upgrading PHPUnit to [version 9](https://phpunit.de/announcements/phpunit-9.html), which [requires at least PHP 7.3](https://phpunit.de/supported-versions.html). Probably also other PHP libraries are dropping support for PHP 7.2 and we need to do the same in order to be able to upgrade to newer versions.

osma:

We want to upgrade to PHPUnit 9 in order to stay up to date. It requires at minimum PHP 7.3, so this PR will also drop support for PHP 7.2.

- Closes #1291

* upgrade to PHPUnit 9.5.* (from 8.5.*) in `composer.json` and in GitHub Actions CI workflow
* migrate `phpunit.xml` configuration file using PHPUnit built in migration tool
* avoid using the deprecated expectException with PHP Errors, use expectError instead
* fix an invalid coverage annotation that wasn't previously noticed
* disable running GitHub Actions CI tests under PHP 7.2
* remove the rules that prevented collecting code coverage on PHP 8.0 (it didn't work under PHPUnit 8 but now works)

none so far :crossed_fingers: 

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php_version: ["7.2", "7.3", "7.4", "8.0"]
+        php_version: ["7.3", "7.4", "8.0"]
         experimental: [false]
         include:
           - php_version: "8.1"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,10 +67,6 @@ jobs:
         CC_TEST_REPORTER_ID: fb98170a5c7ea9cc2bbab19ff26268335e6a11a4f8267ca935e5e8ff4624886c
       with:
         prefix: /app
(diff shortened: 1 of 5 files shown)


joelit (approved):
Tested locally, works fine! :)

PR #1314, opened 2022-04-28
Fix sidebar search, post bootstrap

### LINKED ISSUE #1311: Search options/filters broken after Bootstrap 5 upgrade
https://dev.finto.fi/yso/en/search?clang=en&q=cat

1. Perform a search (e.g. "cat" in YSO) and press Enter to get to the search result page
2. Look at the search options sidebar on the left

Expected to see filters that can be used to narrow the search, like this (finto.fi / skosmos 2.14):

image

(Here, I've clicked on "By group" to open the list of possible groups)

But instead, the filters are not displaying normally and the lists of values cannot be opened:

image

I think this is a side effect of the Bootstrap 5 upgrade (#1182), ping @kinow

Firefox

kinow:

Bootstrap upgrade PR broke the sidebar search (completely forgot to test this part during the Bootstrap upgrade! Sorry :fearful: )

- Closes #1311

Fix sidebar search, by matching previous CSS styles. Had to modify the structure of HTML, and copy the `btn-default` style from Bootstrap 3.

## Checklist

### DIFF
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -230,6 +230,12 @@ ul {
   margin: 0 !important;
 }
 
+.btn-default {
+  color: #333;
+  background-color: #fff;
+  border-color: #ccc;
+}
+
 @media print { /* Quick style sheet for printing. */
   body * {
     visibility: hidden;
@@ -577,6 +583,11 @@ ul.dropdown-menu > li:last-child > input {
 
 .multiselect-all > .form-check {
   padding: 3px 20px;
+}
+
+.multiselect-selected-text::after {
+  content: " ";
+  white-space: pre;
 }
 
 .form-check-input:checked {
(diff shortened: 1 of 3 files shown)


kinow:
This will need some testing locally or on finto dev. Here's what it looks like for me, running with Docker.

image

osma:
Thanks a lot @kinow ! Will test...

osma:
I tested this briefly. Here you can see the old Skosmos 2.14 (left) and new (right):

kuva

It's working well now!

The font color is different for the options - it used to be blue (link color) but now it's black. I think the black color is better actually. These are not links so there is no reason to use blue color. Also the highlighting is different, but here too I think the new look is better.

For the group selection, I suggest changing the square checkboxes to round radio buttons, as they used to be. This is because only one selection is allowed. For the others, the square checkboxes are good.

In general there is a little bit more vertical spacing in the new layout, for example between the labels and the selection boxes (highlighted on picture). It's the same with the search results seen on the right - they take up more space than they used to. For example the headline "catacombs" is visible in the old layout on the left side but not in the new layout on the right side. I wonder if there is some general CSS setting (line-height maybe?) that is different in Bootstrap 5 than it used to be in Bootstrap 3? This isn't super critical, just something I wonder about because it seems to affect all page types, not just the search result page shown here.

kinow:
>For the group selection, I suggest changing the square checkboxes to round radio buttons, as they used to be. This is because only one selection is allowed. For the others, the square checkboxes are good.

:+1: 

>In general there is a little bit more vertical spacing in the new layout, for example between the labels and the selection boxes (highlighted on picture). It's the same with the search results seen on the right - they take up more space than they used to. For example the headline "catacombs" is visible in the old layout on the left side but not in the new layout on the right side. I wonder if there is some general CSS setting (line-height maybe?) that is different in Bootstrap 5 than it used to be in Bootstrap 3? This isn't super critical, just something I wonder about because it seems to affect all page types, not just the search result page shown here.

I will spend some time touching up after fixing the checkbox/radiobutton, trying to make it look more similar to finto. Can probably finish it tomorrow morning (not working, and my soccer team is playing in Sao Paulo, so eyes on the screen, ears to the radio :soccer: )

Thanks for testing it so quickly @osma!

kinow:
>For the group selection, I suggest changing the square checkboxes to round radio buttons, as they used to be. This is because only one selection is allowed. For the others, the square checkboxes are good.

Huh, interesting. Looking at the HTML DOM, the "By group" fields are actually using `type="radio"`.

image

One CSS rule that I added to have sharp corners in the checkboxes ended up causing the radio inputs to have sharp corners too. Essentially becoming check boxes :laughing: Fixed. Now looking at the remaining style issues.

kinow:
>In general there is a little bit more vertical spacing in the new layout, for example between the labels and the selection boxes (highlighted on picture). It's the same with the search results seen on the right - they take up more space than they used to. For example the headline "catacombs" is visible in the old layout on the left side but not in the new layout on the right side. I wonder if there is some general CSS setting (line-height maybe?) that is different in Bootstrap 5 than it used to be in Bootstrap 3? This isn't super critical, just something I wonder about because it seems to affect all page types, not just the search result page shown here.

Bootstrap 5 had a class `.form-label` with `margin-bottom: .5em`. But Bootstrap 3 had a fixed margin of `5px`. I've copied the Bootstrap 3 setting over. The downside is when zooming in the margin won't be adjusted, which goes against accessibility practices, like WCAG. But now the look and feel should be more similar.

Some screenshots.

image

image

Had a look at the Feedback form and at the search language switcher to see if these changes broke those `<select>` inputs, but everything looks the same to me.

Ready for review again :tada: 
Thanks
Bruno

osma:
Excellent work @kinow! All good now, I'm merging this.

PR #1315, opened 2022-04-28
fix 'Any language' choice in language selection menu

osma:

The "Any language" option in the content language selector was not working properly, probably because of changes caused by the Bootstrap 5 upgrade (#1182). When the user selected "Any language", the text changed to "undefined". This PR restores the original functionality.

- cleanup after #1182

* use `id="lang-button-all"` instead of `class="lang-button-all"` because Bootstrap 5 seems to overwrite the class attribute but the `id` is preserved

n/a

## Checklist

### DIFF
--- a/view/headerbar.twig
+++ b/view/headerbar.twig
@@ -68,7 +68,7 @@
               {%- if term %}&q={{ term }}{% endif -%}
               {%- if vocabs %}&vocabs={{ vocabs }}{% endif -%}
               "
-              class="lang-button lang-button-all">
+              class="lang-button" id="lang-button-all">
               {%- trans "Search language: any" -%}
             </a>
             <input name="anylang" type="checkbox"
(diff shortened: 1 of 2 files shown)


view/headerbar.twig
```diff
@@ -68,7 +68,7 @@
               {%- if term %}&q={{ term }}{% endif -%}
               {%- if vocabs %}&vocabs={{ vocabs }}{% endif -%}
               "
-              class="lang-button lang-button-all">
+              class="lang-button" id="lang-button-all">
```
kinow, line 71: Thanks! I remember seeing other cases where Bootstrap (or plug-ins) re-wrote the DOM structure or attributes :disappointed: 

Sorry for yet another post-bootstrap upgrade. Hopefully we are getting close to a release now, with no more issues like this :crossed_fingers:

kinow (approved):

view/headerbar.twig
osma: I think a few more patches will be needed...there is still a small backlog of reports from users currently testing dev.finto.fi - mostly minor layout issues such as spacing and font sizes. I will open issues for bigger functional problems, but for minor or purely layout fixes it's simpler to just make a fix and open a PR.

PR #1318, opened 2022-05-03
Create SECURITY.md

### LINKED ISSUE #1292: Create a SECURITY.md file
Hi,

Today I saw a security issue reported in the mailing list for Skosmos. I think an alternative workflow would be to have an email listed in a file where users could redirect this kind of request. Then devs could confirm the issue and work on a new release with the security fix.

It's common to have a `SECURITY.md` file detailing which channels to use to report security issues, what the process is like, etc. Maybe we should adopt one too?

Bruno

joelit:

Skosmos repository was missing a security policy file from its community guidelines.

- Closes #1292 

We need to discuss which versions we are supporting with security patches. Release documentation needs to include updating the security policy file.

## Checklist

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,14 +2,13 @@
 
 ## Supported Versions
 
-Following Skosmos versions are currently being supported with security updates.
+Following Skosmos versions are currently being supported with security updates. The "current development branch" means the master branch of the repository, whereas the "maintenance branch" corresponds with the latest release of Skosmos.
 
-| Version | Supported          |
-| ------- | ------------------ |
-| 2.15    |  ✔️                 |
-| 2.14.x  |  ✔️                 |
-| 2.13.x  | :white_check_mark: |
-| < 2.13  | :x:                |
+| Version                     | Supported          |
+| --------------------------- | ------------------ |
+| Current development branch  |  ✔️                 |
+| Current maintenance branch  |  ✔️                 |
+| Older versions              | :x:                |
 
 ## Reporting a Vulnerability
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,20 @@
+# Security Policy
+
+## Supported Versions
+
+Following Skosmos versions are currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 2.15    |  ✔️                 |
+| 2.14.x  |  ✔️                 |
+| 2.13.x  | :white_check_mark: |
+| < 2.13  | :x:                |
+
+## Reporting a Vulnerability
+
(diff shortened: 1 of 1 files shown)


kinow (approved):
🎉

Probably good to update the release checklist with a note to check the versions in this doc.

Thanks!!!

osma:
I wonder if we could make use of the newly published RFC 9116 which defines a machine- and human-readable `security.txt` file format: https://www.rfc-editor.org/rfc/rfc9116 Or is that perhaps overkill?

Ping @kinow , what do you think as the original proposer of this?

kinow:
> I wonder if we could make use of the newly published RFC 9116 which defines a machine- and human-readable `security.txt` file format: https://www.rfc-editor.org/rfc/rfc9116 Or is that perhaps overkill?
> 
> Ping @kinow , what do you think as the original proposer of this?

I didn't know about this RFC @osma, thanks!

Had a look at the specification, and looks like it'd something to have deployed to Finto, for instance, similar to the `robots.txt` file. 

I think `SECURITY.md` is intended for developers and end-users to learn how to report. And while `security.txt` can also be used for that, it was created to contain a format with specific fields so that it can be parsed by machines too. From the spec:

>This document defines a machine-parsable format ("security.txt") to help organizations describe their vulnerability disclosure practices to make it easier for researchers to report vulnerabilities.

SECURITY.md
```diff
@@ -0,0 +1,20 @@
+# Security Policy
+
+## Supported Versions
+
+Following Skosmos versions are currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 2.15    |  ✔️                 |
```
osma, line 9: I think having to maintain a table like this is a recipe for outdated information. Would it be better just to state that the latest vX.X-maintenance branch is supported and nothing else?

kinow (approved):
Easier! No need to add one more item to the release process, good idea!

joelit:
So, this is how the latest iteration would look - less upkeep with no version numbers.

Screenshot from 2022-05-04 11-04-31

SECURITY.md
```diff
@@ -0,0 +1,19 @@
+# Security Policy
+
+## Supported Versions
+
+Following Skosmos versions are currently being supported with security updates. The "current development branch" means the master branch of the repository, whereas the "maintenance branch" corresponds with the latest release of Skosmos.
```
osma, line 5: "currently being supported with security updates" - this doesn't say who is responsible. Could we say:

> currently being supported with security updates by the Skosmos development team at the National Library of Finland.

Minor grammar correction: "corresponds with" -> "corresponds to"

To be a bit more explicit about branches, could we express it like this:

> whereas the "maintenance branch" is a branch called `vX.X-maintenance`, where the version number X.X corresponds to the latest release of Skosmos.

osma (approved):
LGTM!

Issue #1151, opened 2021-04-07
Upgrading jQuery 2.x dependent libraries to support jQuery 3.x


Vainonen:
In order to properly migrate to jQuery 3.x at least the following libraries need to be upgraded:

- [x]     malihu/malihu-custom-scrollbar-plugin (visible difference between jQuery 2.x and 3.x)
- [x]     grimmlink/qtip2 (jQuery migration plugin warnings)
- [ ]     twitter/typeahead.js (jQuery migration plugin warnings)
- [x]     twitter/bootstrap (jQuery migration plugin warnings)
- [ ]     etdsolutions/waypoints (jQuery migration plugin warnings)
- [ ]     vakata/jstree (jQuery migration plugin warnings)
- [x]     components/jqueryui (jQuery migration plugin warnings)

Some of these "only" use deprecated features that are not per se dangerous and jQuery Migrate plugin (which restores old API) can be used in production for a while. See https://github.com/jquery/jquery-migrate/blob/main/warnings.md for more information.

Additional info: https://jquery.com/upgrade-guide/3.0/ and https://jquery.com/upgrade-guide/3.5/

osma:
Bootstrap just got upgraded in PR #1182 so now it's one blocker less. :tada:

kinow:
Had a look at some of the pending items of the list (excluding the first item):

- `grimmlink/qtip2`

Archived by maintainer. Easiest solution, IMO, is adopting Bootstrap's Popover (requires Popperjs only, no JQuery).

Update templates and docready.js to use the popover component instead. I think we should be able to customize the CSS, but not sure if it will look 100% like the current qtip popup (if it's a lot to customize, that could become an issue when upgrading to a new version of bootstrap too).

- `twitter/typeahead.js`

It hasn't been updated in a long time (last commit from 2015). We can review the warnings and keep using it, or search for a replacement component, or try the browser's Datalists. These are [supported in Bootstrap](https://getbootstrap.com/docs/5.1/forms/form-control/#datalists), here's the [caniuse](https://caniuse.com/datalist) for Datalists.

This solution with Datalists also is independent of JQuery. Not sure if we can customize how it looks like, I think it's more like the file & date pickers, that are handled by the browser.

- `etdsolutions/waypoints`

Hmm, last commit from 2015 too. Some users already [reported problems](https://github.com/imakewebthings/waypoints/issues/618) with JQuery.

[...]

PR #1324, opened 2022-05-29
Replace qtip JS library by a pure CSS tooltip

kinow:

Replace the deprecated qtip JS library by a pure CSS tooltip.

- Related to #1151 

In #1151, when I looked at the qtip library alternatives, the first that I thought about was using Bootstrap's popover component. That requires us including and loading the Popper.js library.

My idea was to add this new dependency, import before `docready.js`, and initialize the instances for each object. Similar to what we are doing with qtip. Finally, I would also have to spend some time customizing the CSS to match qtip's output.

However, I thought perhaps we could try a solution without JavaScript first. This way in the future we won't have to worry about tooltips, replacing libraries, updating popper, etc.

I used the example from this post: https://dev.to/r3zu3/pure-css-tooltip-1k3j. The author uses it in his library z.css, licensed under the Apache License.

We need to customize the time of the animation, and the location of the tooltip.

I also have implemented only the "Help" text tooltip. I am not sure if we are not using any other features of qtip in other parts of the UI.

## Checklist


composer.json
```diff
@@ -65,8 +65,7 @@
     "monolog/monolog": "1.23.*",
     "newerton/jquery-mousewheel": "dev-master",
     "pamelafox/lscache": "1.0.5",
-    "malihu/malihu-custom-scrollbar-plugin": "3.1.5",
-    "grimmlink/qtip2": "3.0.3"
```
kinow, line 69: Since this is a CSS only solution, we are only removing a JS dependency from `composer.json`.

resource/css/styles.css
```diff
@@ -2321,14 +2289,6 @@ body, .versal, h1, h2, h3, p, .versal-bold {
...
     margin-top: 20px;
   }
-
-  .qtip-skosmos {
-    min-width: auto;
-    left: 0 !important;
-  }
-  .qtip-skosmos canvas {
-    display: none !important;
-  }
```
kinow, line 2331: We can also remove the qtip CSS classes :point_up:

resource/js/docready.js
```diff
@@ -524,8 +524,6 @@ $(function() { // DOCUMENT READY
     style: { classes: 'qtip-tipsy qtip-skosmos' }
   };
 
-  $('#navi4').qtip(qtip_skosmos);
-
```
kinow, line 528: `#navi4` is the Help menu item. I did only this element before asking others for feedback here.

view/topbar.twig
```diff
@@ -49,7 +49,7 @@
   <a href="{% if request.vocabid and vocab.title%}{{ request.vocabid }}/{% endif %}{{ request.lang }}/feedback{% if request.contentLang and request.contentLang != request.lang %}?clang={{ request.contentLang }}{% endif %}{% if request.queryParam('anylang') == 'on' %}{% if request.contentLang == request.lang %}?{% else %}&{% endif %}anylang=on{% endif %}" id="navi3" class="navigation-font">
   {% trans "Feedback" %}
   </a>
-  <span id="navi4" tabindex="0" title="{% trans "helper_help" %}<br /><br />{% trans "search_example_text" %}">
+  <span class="skosmos-tooltip-wrapper skosmos-tooltip t-bottom" id="navi4" tabindex="0" data-title="{% trans "helper_help" %} &#xa; &#xa; {% trans "search_example_text" %}">
```
kinow, line 52: The `title` attribute has the advantage that in non-JS sites it appears as a tooltip, handled by the browser. But qtip is disabling it after we initialize it.

In order to be JS-free, the only alternative is to use another attribute. We can use aria-* attributes, I think, in case we realize this option causes issues for users using screen readers, for example.

kinow:
Preview of the CSS tooltip:

GIFrecord_2022-05-30_104018

What others think about this solution?

resource/js/docready.js
osma: Right, but there are also other UI elements where qtip is currently used, for example the lines just below this one.

osma (changes requested):
I agree 100% that a pure CSS approach is preferable to a JS-based solution such as Popper. This PR looks like a big improvement. Some comments:

1. I think the animation (appearing from the top) is unnecessary and just adds to visual noise. As a metaphor it doesn't make sense to me - why would the popup window appear from the top? If anything, I think it should "emerge" from the spot where the cursor is pointed, like smoke that comes out from a chimney :) But just appearing out of nowhere, like with qtip2, would be best.

2. There's a little problem with the cross that is used to clear the contents of the search box - it will stay on top of the popup (probably need to adjust z-index?). I can't see the cross at all in your animation but here's how it looks to me (similar in both Firefox and Chrome):

kuva

3. There are other places in the UI where qtip2 is used, you need to address those as well obviously (as pointed out in a comment)

4. I think you also need to remove the following lines:

```
light.twig:<link href="resource/css/jquery.qtip.min.css" rel="stylesheet" type="text/css" />
scripts.twig:<script src="vendor/grimmlink/qtip2/dist/jquery.qtip.min.js"></script>
```

osma:
This PR is a bit old, I think it would be best to merge the changes from `master` and possibly rebase as well

resource/js/docready.js
kinow: Yup. I did just one UI element first to see what others thought, as replacing all the other elements would take a bit longer and wouldn't be worth if the idea was discarded. I will check the rest of your feedback and once I've addressed all of it, I will start replacing in the other UI elements so that we can have a final preview :+1:

composer.json
```diff
@@ -65,8 +65,7 @@
     "monolog/monolog": "1.23.*",
     "newerton/jquery-mousewheel": "dev-master",
     "pamelafox/lscache": "1.0.5",
-    "malihu/malihu-custom-scrollbar-plugin": "3.1.5",
```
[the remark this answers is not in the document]
kinow: Ops! I think I was testing the other components and that change slipped here, sorry. Reverting. Thanks!

composer.json
```diff
@@ -65,8 +65,7 @@
...
-    "grimmlink/qtip2": "3.0.3"
+    "malihu/malihu-custom-scrollbar-plugin": "3.1.5"
```
kinow, line 68: @osma ah! I think it was the GitHub UI's diff, showing the entry as removed because I removed the comma? Still good to check, thanks!

kinow:
Hi @osma, replies inline

> I agree 100% that a pure CSS approach is preferable to a JS-based solution such as Popper. This PR looks like a big improvement. Some comments:

Excellent!

>     1. I think the animation (appearing from the top) is unnecessary and just adds to visual noise. As a metaphor it doesn't make sense to me - why would the popup window appear from the top? If anything, I think it should "emerge" from the spot where the cursor is pointed, like smoke that comes out from a chimney :) But just appearing out of nowhere, like with qtip2, would be best.

Agreed. I actually prefer no animation if possible. Will change it.

>     2. There's a little problem with the cross that is used to clear the contents of the search box - it will stay on top of the popup (probably need to adjust z-index?). I can't see the cross at all in your animation but here's how it looks to me (similar in both Firefox and Chrome):
> 
> 
> kuva

Oh, doesn't look good. Let me see what I can do to improve it. Good catch!

>     3. There are other places in the UI where qtip2 is used, you need to address those as well obviously (as pointed out in a comment)

Roger that :+1: I will update the other ones once we have fixed the feedback items above.

>     4. I think you also need to remove the following lines:
> 
> 
> ```
> light.twig:<link href="resource/css/jquery.qtip.min.css" rel="stylesheet" type="text/css" />
> scripts.twig:<script src="vendor/grimmlink/qtip2/dist/jquery.qtip.min.js"></script>
> ```

Ah! Yup. I left that because I didn't disable qtip in other places, I think. From what I remember, I left it enabled somewhere else and then opened two separate browser windows, to compare the generated CSS and how the UI behaved.

This will be removed when I have finished the item above, of updating all the other UI elements that use qtip :+1: 

Thanks!!

kinow:
Note to self:

1. Remove call to qtip, in `docready.js`;

`$('#navi4').qtip(qtip_skosmos);`

2. Find the original span for that qtip removed, in the views folder;

3. Add the CSS classes for new tooltip, `class="skosmos-tooltip-wrapper skosmos-tooltip t-bottom"`;

4. Rename the `title="…"` to `data-title="…"`;

5. Replace any `<br/>`'s by `&#xa;`;

Example:

- Before:

`<span id="navi4" tabindex="0" title="{% trans "helper_help" %}<br /><br />{% trans "search_example_text" %}">`

- After:

`<span class="skosmos-tooltip-wrapper skosmos-tooltip t-bottom" id="navi4" tabindex="0" data-title="{% trans "helper_help" %} &#xa; &#xa; {% trans "search_example_text" %}">`

TODO: add note about what to do when the JS call specifies the location (just need to change the CSS classes…)

kinow:
I changed and tested `property-click`, `hier-trigger`. For `redirected-vocab-id` I changed, but didn't know how to test it. Also, `redirected-vocab-id` had three occurrences, but only 1 with a `title` attribute (used by `qtip` and by this pure-css approach (so I am assuming the other two can be ignored).

I couldn't load AGROVOC to test reified properties. @osma would you have another smaller example dataset that I can load to test it, please?

The reified icon is trickier, I just realized, as it displays not a text element, but instead a child div with other HTML elements. Gotta have to think more how to replace this one with pure css (still doable I think).

Thanks
-Bruno

composer.json
osma: Ah, you are right. I misread the diff. Sorry for the noise!

osma:
> I couldn't load AGROVOC to test reified properties. @osma would you have another smaller example dataset that I can load to test it, please?

Can you use the snippet from https://github.com/NatLibFi/Skosmos/pull/1356#issuecomment-1239492145 ? See the screenshots in the same comment for how it should look (but note that the display of that kind of partial SKOS XL data is a bit broken unless you also apply the changes in PR #1356)

kinow:
Rebased, and then modified my existing YSO data to have the prefix for skosxl and to have the property in the `p6095` concept as in the example you provided @osma.

But now if I open `p6095` it's giving me a blank page and the Docker logs contain:

```bash
skosmos-web     | [Tue Sep 13 06:44:34.769517 2022] [php7:error] [pid 11] [client 172.19.0.1:41434] PHP Fatal error:  Uncaught EasyRdf\\Http\\Exception: HTTP request for SPARQL query failed in /var/www/html/vendor/easyrdf/easyrdf/lib/Sparql/Client.php:247\nStack trace:\n#0 /var/www/html/vendor/easyrdf/easyrdf/lib/Sparql/Client.php(128): EasyRdf\\Sparql\\Client->request()\n#1 /var/www/html/model/sparql/GenericSparql.php(93): EasyRdf\\Sparql\\Client->query()\n#2 /var/www/html/model/sparql/GenericSparql.php(1475): GenericSparql->query()\n#3 /var/www/html/model/Concept.php(536): GenericSparql->queryLabel()\n#4 /var/www/html/vendor/twig/twig/src/Extension/CoreExtension.php(1566): Concept->getProperties()\n#5 /tmp/skosmos-template-cache/ed/edd3a5fbdba4b947d8c145cd98bb65cab172eaea5513f6ffe050cafc75630ff2.php(400): twig_get_attribute()\n#6 /var/www/html/vendor/twig/twig/src/Template.php(405): __TwigTemplate_34d06a17ee23cf00128016f7cd55bc42f28d9e449367abf9e5f1bad707bb4a00->doDisplay()\n#7 /var/www/html/vendor/twig/twig/src/Template.php(378): Twig\\Template->displayWithErrorHandling()\n#8 /tmp/skosmos-template-cache/66/66e752b362c577ad7cde0210772fed34 in /var/www/html/view/concept-shared.twig on line 95
```

Do you know if I need to modify something else in my Skosmos installation to have skos-xl now? I can access other concepts just fine, it's just this one that's giving me the error.

Thanks

osma:
That looks like a SPARQL query failed. Can you do either/both of these:

1. enable SPARQL query logging in config.ttl to a file so you can see the query that failed
2. check the Fuseki log

osma:
Just to clarify this question:

> Do you know if I need to modify something else in my Skosmos installation to have skos-xl now? I can access other concepts just fine, it's just this one that's giving me the error.

You shouldn't need to modify anything. SKOS-XL should work right out of the box without any configuration. It's a mystery why you are getting an error for this concept, but hopefully the answer can be found in the SPARQL query logs.

kinow:
I'm on mobile right now. Will get the queries in a few minutes. I enabled the error logging and js console logging.

There were a few SPARQL queries logged and all were executed successfully but the last one returned no results (I think it was for property values in a specific language).

kinow:
Maybe that default setting, for union graph in Jena? I think I had to enable that for some other issue.

osma:
Are you still getting the PHP error from https://github.com/NatLibFi/Skosmos/pull/1324#issuecomment-1244984841 ?
It seemed to indicate a failed SPARQL query - perhaps a syntax error?
The results from the three queries are all valid responses (even the last one that's empty), but I'd expect there would be something like a 400 error from Fuseki hiding somewhere.

kinow:
Ah! Mystery partially solved. I know why it wasn't working, but not why the logs do not contain the error. If you would like to reproduce it, @osma, just intentionally enter a wrong global SPARQL URL :grimacing: 

I am using Docker, and used the values from the example config, which use a dataset called `ds`.

```
:config a skosmos:Configuration ;
    # SPARQL endpoint
    # a local Fuseki server is usually on localhost:3030
    skosmos:sparqlEndpoint <http://fuseki:3030/ds/sparql> ;
```

In order to find it I just changed the PHP code to write to a file in the container :rofl: never fails!

```
$var_str = var_export($this->client, true);
file_put_contents('/tmp/test.txt', $var_str);
```

osma:
> If you would like to reproduce it, @osma, just intentionally enter a wrong global SPARQL URL :grimacing:

Ah, of course! Been there, done that. But it always causes a lot of confusion when you can't find the Fuseki error that "should be there"...

> The icon for the SKOS-XL label is not there, but I think it might be due to the language.

In that example, the SKOS XL label was defined for just one language. You will need to switch to that as content language to see it (the icon with the popup) in the header, or to another content language to see it in the "in other languages" section.

kinow:
>In that example, the SKOS XL label was defined for just one language. You will need to switch to that as content language to see it (the icon with the popup) in the header, or to another content language to see it in the "in other languages" section.

I tried defining the XL label in other languages, and also accessing the /en/, /fi/, /se/, /sv/ URL's for the concept, varying the ?clang in the URL, but nothing made that icon appear in the UI.

At least the UI didn't break this time. I'll pause for now, and will try it again another day. Maybe starting afresh will solve it :) Thanks @osma

osma:
Are you sure you included all the triples from https://github.com/NatLibFi/Skosmos/pull/1356#issuecomment-1239492145 when you incorporated them into your copy of YSO?

In particular, you the skosxl:prefLabel triple that links the concept to the Label instance:

    yso:p6095 skosxl:prefLabel <http://www.yso.fi/onto/yso-meta/p6095#se_1> .

as well as the skos:label in Northern Sámi language:

    yso:p6095 skos:prefLabel "soahki"@se .

Now that I think of it, this example is a bit special because the label is in Northern Sámi language, a minority language that we have just began to include in YSO (and that's why we need better SKOS XL support, as we want to include some provenance information for the labels, for example the source of translations). You probably also need to include Northern Sámi as a content language for YSO in `config.ttl`:

    skosmos:language "fi", "sv", "en", "se";

kinow:
> Are you sure you included all the triples from [#1356 (comment)](https://github.com/NatLibFi/Skosmos/pull/1356#issuecomment-1239492145) when you incorporated them into your copy of YSO?
> 
> In particular, you the skosxl:prefLabel triple that links the concept to the Label instance:
> 
> ```
> yso:p6095 skosxl:prefLabel <http://www.yso.fi/onto/yso-meta/p6095#se_1> .
> ```

Just checked, it's in that whole block I copied from that comment :+1: 

> as well as the skos:label in Northern Sámi language:
> 
> ```
> yso:p6095 skos:prefLabel "soahki"@se .
> ```
> 
> Now that I think of it, this example is a bit special because the label is in Northern Sámi language, a minority language that we have just began to include in YSO (and that's why we need better SKOS XL support, as we want to include some provenance information for the labels, for example the source of translations).

I believe it's done too. I copied the whole code snippet from https://github.com/NatLibFi/Skosmos/pull/1356#issuecomment-1239492145. I downloaded the Turtle from YSO, and replaced the existing concept by that one.

>You probably also need to include Northern Sámi as a content language for YSO in `config.ttl`:
> 
> ```
> skosmos:language "fi", "sv", "en", "se";
> ```

I didn't know what was the se, but I had enabled it in my `config.ttl`. Right now I have:

```
    skosmos:languages (
        [ rdfs:label "fi" ; rdf:value "fi_FI.utf8" ]
        [ rdfs:label "sv" ; rdf:value "sv_SE.utf8" ]
        [ rdfs:label "se" ; rdf:value "se_SE.utf8" ]
        [ rdfs:label "en" ; rdf:value "en_GB.utf8" ]
    ) ;
```

And these languages appear in the comboboxes in the Skosmos UI. FWIW, here's my config for `yso` in `config.ttl` too:

```turtle
:yso a skosmos:Vocabulary, void:Dataset ;
    dc:title "YSO - Yleinen suomalainen ontologia"@fi,
        "ALLFO - Allmän finländsk ontologi"@sv,
        "YSO - General Finnish ontology"@en ;
    dc:subject :cat_general ;
    dc:type mdrtype:ONTOLOGY ;
    void:uriSpace "http://www.yso.fi/onto/yso/";
    skosmos:language "fi", "sv", "en";
    skosmos:defaultLanguage "fi";
    skosmos:showTopConcepts "true";
    skosmos:showStatistics "false";
    skosmos:loadExternalResources "false";
    skosmos:shortName "YSO",
                    "ALLFO"@sv;
    skosmos:groupClass isothes:ConceptGroup ;
    skosmos:arrayClass isothes:ThesaurusArray ;
    void:dataDump <http://api.finto.fi/download/yso/yso-skos.ttl> ;
    void:sparqlEndpoint <http://fuseki-cache:80/skosmos/sparql> ;
    skosmos:sparqlGraph <http://www.yso.fi/onto/yso/> ;
    skosmos:mainConceptScheme <http://www.yso.fi/onto/yso/>
```

kinow:
Ah, just realized I added the global language, not the lang in the vocab. Let me quickly restart the containers to try that…

kinow:
That solved it @osma! Thanks!!! :tada: 

image

kinow:
@osma, if CI passes, I believe this PR should now be ready for review :slightly_smiling_face: 

There are two ways to create tooltips, the one that uses the `data-title="tooltip message"` attribute, and another one that uses a child HTML element to be displayed in the pop up.

I tested the latter yesterday with the SKOS-XL snippet, with your help. Then, I changed the final places that had the reified tooltips… I think these might work, but would require someone with a vocabulary data to test it.

Things that reviewers must be aware while testing:

- This uses the HTML element hover and focus attributes. So play with the mouse passing over, and compare with the behavior of Finto+qtip, to see if the new component is behaving OK.
- For the reified and the SKOS-XL tooltips, I had to write the CSS classes copying the other tooltip. Both should look the same, but still good to have someone else to verify it.
- I tested it on Chrome and Firefox on Linux. It should work with any modern browser, but always good to test with other OS's and browsers.

Thanks! :bow: 
-Bruno

osma:
Thanks @kinow ! I've had to concentrate on other things but I will do a full review tomorrow!

osma:
I did some testing, comparing current `master` with qtip2 ("before") against this PR ("after") and looking at all the tooltips. This was with the Finto CSS applied on top of the basic Skosmos CSS; I hope that's not causing extra problems (please say so if you're not seeing the same things). Using Firefox on Ubuntu.

# Help tooltip

Before:
image

After:
image

Looks similar enough, and the font is actually now better, the old version was off. The only semi-important regression is that in the second paragraph, there is some extra space before the first word (the same happens with text in other languages). Pretty minor, but if there's an easy fix, it could be removed.

However, if I actually click on "Help", this happens:
image

The language selection pops up in front of the tooltip, just like the cross icon did earlier. Some z-index adjustment needed?

# Reified definitions

To make this visible, I inserted this somewhat bogus definition into the data for the example concept:

    skos:definition [ rdf:value "kirjallisuus joka on kirjoitettu suomen kielellä"@fi; dc:source "internet"@fi ] ;


Before:
image

After:
image

This is an improvement! For some reason the color was off in the old version. I also like the new layout better. But the text size is a bit larger than before.

Here is also a bug, the label for the property `dc11:source` is not being shown properly (should be "Source"). But it's not caused by this PR. I know the cause (I already fixed a similar case with SKOS XL labels), I fill fix it in a separate PR.

# SKOS XL information for the prefLabel as a heading

Before:
image

After:
image

Again, the font size is a bit larger than before. The color of the value is wrong.

# SKOS XL information for the prefLabel in another language

Before:
image

After:
image

Same situation as above. The font size is larger than before and the color of the value is wrong.

# Other

There's an unintended change to the property names.

Before:
image

After:
image

Previously, only "TYPE" was set in bold. Now the other property labels are in bold as well, except for "IN OTHER LANGUAGES" which is a bit special as it's not a regular SKOS/RDF property.

kinow:
Thanks a lot for the through testing @osma (as always). Everything sounds doable to fix, and not too hard. I may push a new commit over this weekend or in the beginning of the next week. I think we are almost there with this PR! :tada: :partying_face: 

Thanks
Bruno

osma:
Great @kinow , I also think this is very close to being ready. Today is the last day of our current dev sprint so we will try to finish off other work if possible. Even if there are still some layout issues in this PR, it should be good enough for testing the jQuery upgrade without qtip interfering (it seems to cause a lot of deprecation messages).

kinow:
Rebased.

kinow:
>Looks similar enough, and the font is actually now better, the old version was off. The only semi-important regression is that in the second paragraph, there is some extra space before the first word (the same happens with text in other languages). Pretty minor, but if there's an easy fix, it could be removed.

Fixed! I used `pre-wrap` which breaks at newline symbols, but keeping white spaces. Now I've replaced it by `pre-line`, which breaks at newlines but collapses white spaces.

image

kinow:
>However, if I actually click on "Help", this happens:
>
>The language selection pops up in front of the tooltip, just like the cross icon did earlier. Some z-index adjustment needed?

Huh, that's very interesting! There was a `:focus` rule, that sets any element with focus to `z-index: 1` in `styles.css`. Fixed as well.

image

kinow:
> Reified definitions
>This is an improvement! For some reason the color was off in the old version. I also like the new layout better. But the text size is a bit larger than before.

I checked that, and it's using the root css variable, set for `14px`, like the other tooltips, @osma. I **think** in the past the reified was also using 14px, but maybe the other elements were using a larger font? Not sure.

https://github.com/NatLibFi/Skosmos/blob/454f1d0c0e1903a9fd926cb8eb104bb7f1fb8e7e/resource/css/styles.css#L535-L539

Would you like me to make just the reified tooltips a little smaller? Maybe `12px`?

Here's a preview where I didn't touch the CSS of the normal tooltip, used in the Help link, and where I reduced the reified tooltip text to 12px:

image

kinow:
>Other
>There's an unintended change to the property names.
>Previously, only "TYPE" was set in bold. Now the other property labels are in bold as well, except for "IN OTHER LANGUAGES" which is a bit special as it's not a regular SKOS/RDF property.

Ah! Good spot, I accidentally removed spaces between css classes, so it was using `versal-boldpropertyclick` or something similar to that, instead of `versal-bold propertyclick`. It should be fixed now.

image

kinow:
>SKOS XL information for the prefLabel as a heading
>SKOS XL information for the prefLabel in another language

I think this is the last pending item in your review, @osma. I will need a little more time to re-read it and understand how/what I should reproduce and test it. I think it's with the birches example I copied into my copy of YSO vocab, but it appears to be working correctly for me, but also not rendering like in your screenshots. Will leave this one to be fixed for over the next days :wave: 

Thanks!

osma:
Thanks again @kinow!

I made a new round of testing, this time without applying the Finto CSS layout in case it might interfer. It looks great now, pretty much all of my worries have been addressed. I could only find two quite minor and trivial things to change:

# Property name tooltips

This is something I forgot to look at in detail in my previous review.

Before:
kuva

After:
kuva

1. Is it possible to show a little arrowhead pointing at the property name, to make it clear what the tooltip is referring to?
2. The text is all capitalized even though it shouldn't be.

# Display differences between reified definitions and SKOS XL labels

Here is a tooltip for a reified definition resource:
kuva

Here is a similar tooltip for a SKOS XL label:
kuva

They are very similar but not identical. The property name "Source:" is set in bold in the SKOS XL tooltip, but not the one for the definition. I like the bold version more, so could the tooltip for definitions (and similar resources) also use bold face for the property name?

Other than these, I can't think of anything more that would need to be fixed before merging, so I think this is really, really close now! And already this is IMHO better than the current status quo because qtip2 seems to be a bit buggy (sometimes tooltip contents just disappear) and also there are some font and color issues with the old tooltips that this PR fixes as well :+1:

kinow:
>Other than these, I can't think of anything more that would need to be fixed before merging, so I think this is really, really close now! And already this is IMHO better than the current status quo because qtip2 seems to be a bit buggy (sometimes tooltip contents just disappear) and also there are some font and color issues with the old tooltips that this PR fixes as well +1

Thanks a lot for testing it again and for the great feedback and screenshots @osma. I think we will have it ready for a final review and merge in the next iteration (i.e. next commits pushed soon). All your points sound doable, and not too hard :slightly_smiling_face:

kinow:
Replies inline

# Property name tooltips

1. [x] Is it possible to show a little arrowhead pointing at the property name, to make it clear what the tooltip is referring to?
2. [x] The text is all capitalized even though it shouldn't be.

Done! I also changed the alignment of the tooltip for `t-top` class. They should now appear left-aligned, more similar to what it was before. In the screenshot below I force two tooltips to be displayed, which is not possible normally, but you can force with the browser development tools. That's just to show that the two tooltips are now left-aligned, which I think makes it easier to keep your eyes in place as you move them down following the mouse to read the tooltips (easily reverted if needed). Also added the arrow sticking out of the tooltip balloon. It was covered by the tooltip.

image

# Display differences between reified definitions and SKOS XL labels

[x] They are very similar but not identical. The property name "Source:" is set in bold in the SKOS XL tooltip, but not the one for the definition. I like the bold version more, so could the tooltip for definitions (and similar resources) also use bold face for the property name?

These are controlled in the template, not in the CSS style. I'm looking at the birches concept page. The tooltip of the Definition icon doesn't have bold anywhere. The tooltip of the Text In Other Languages does.

image

That's controlled by having the `versal-pref` at the top element (with `text-weight: 500`), and the `versal` class at the tooltip text after the “Source” text (with `text-weight: 400`). I changed the template to have the Definition icon tooltip to use a similar structure, but note that that activates the `a.versal` class, which makes the text blue (I think you mentioned you liked more the all white text, easy to change it if needed, but not sure if that `a.versal` is not used elsewhere). Preview:

image

After this screenshot, I re-read your comment, @osma, and after reading this part “also use bold face for the property name?”, so I moved the `versal-pref` up one level. Could you take a look if that's what you meant? Otherwise we just need to move the CSS class one level down and we will have the result of the previous screenshot. Preview:

image

I also noticed that to reach the link sometimes the tooltip would disappear if the mouse cursor accidentally hovered outside the tooltip. I tweaked a bit the CSS classes and the location of the tooltip arrow, to see if that reduces the chances of the tooltip disappearing while the user moves it to click on the Source link. qtip could use JS to disappear with the tooltip after a few milliseconds, which gave some buffer for the user to move the mouse outside the element. We don't have that feature with CSS-only (a good trade-off I think, given that it reduces the JS dependencies and simplifies the code a bit IMO).

The good news is that if there are other parts where the tooltip must adjust the text weight we should be able to do so without changing any or much of the tooltip CSS, and controlling that via the existing CSS classes and HTML structure.

Let me know if this is looking better or if we need to further adjust it (feel free to commit directly if you'd like, I can sync the branch and post a preview screenshot here too). So ready again for review whenever you have time @osma! Thanks :+1: 

Cheers,
Bruno

osma:
Thanks @kinow ! Everything was good except I wasn't 100% happy with the way properties and values were displayed - which was inconsistent from the beginning so no fault in this PR really). I made a few more tweaks in a commit that I just pushed to your PR branch. I will merge this now :tada:

kinow:
Thanks @osma!

MikkoAleksanteri:
A rather strange artifact that looks like an empty tooltip box has appeared to dev.finto.fi where we are testing the new skosmos version. This appears when mouse is hovered over "Hierarcy" tab while any of the other tabs are active. See for example: https://dev.finto.fi/yso/en/ 

image


According to @osma this might be related to this PR, any ideas about the cause and possible fix Osma and @kinow ?

kinow:
Hi @MikkoAleksanteri 

>A rather strange artifact that looks like an empty tooltip box has appeared to dev.finto.fi where we are testing the new skosmos version. This appears when mouse is hovered over "Hierarchy" tab while any of the other tabs are active. See for example: https://dev.finto.fi/yso/en/

Ah, I loved the description :smile: That's the tooltip for the Hierarchy tab option I think. The rotated (45 degree?) square is the "triangle" or "arrow" that sticks out of the tooltip. The black/darker part is the actual tooltip, but it appears to be completely empty. Causing it to look strange that way.

>According to @osma this might be related to this PR, any ideas about the cause and possible fix Osma and @kinow ?

That's really odd. I couldn't see anything wrong from a quick inspection with the browser. The top menu “Help” option has a tooltip too, and that one appears to be working fine. Let me check the templates to see if there's anything that could break it, and if I cannot find it I will try to reproduce it locally.

Thanks @MikkoAleksanteri 
-Bruno

kinow:
Reproduced with YSO locally using `master`:

image

Issue #678, opened 2017-12-22
The REST API types query is broken


henriyli:
## At which URL did you encounter the problem?
http://api.dev.finto.fi/rest/v1/types?lang=fi

## What steps will reproduce the problem?
1. Query the rest api for concept types.  

## What is the expected output? What do you see instead?
Expected to see all the different types of concepts and their labels, getting an empty list instead.

osma:
Fuseki union default graph setting was disabled on the dev server. This started working again when I enabled it.

However, I don't think we should rely on union default graph, if it's possible to avoid.

PR #1326, opened 2022-06-01
Fix typeahead templates in the search auto-complete

### LINKED ISSUE #1323: Vocabulary information is missing from the drop-down results list
At the vocabularies list, front page, e.g. dev.finto.fi

1. At the search bar, make sure that "from all" is checked in the source vocabulary list
2. Type a search term and wait for autocomplete to show the search results
3. The source vocabulary information is missing from the result drop-down list

There should be information of the source vocabulary for each search result.

A list of results with no source information means it is not possible to differentiate between the results with same preferred labels.
Screenshot from 2022-05-17 10-49-44

Firefox & Chrome

kinow:
The template appears to be more or less like the previous one. One thing that I noticed is that we have two `div`'s (in both 2.14, and current version). So my first guess appears to be correct, that Twitter Typeahead drops the second div, instead of displaying it as before.

- [x] Use a single div to keep the Vocabulary information
- [x] Fix the CSS style (colors, font)
- [ ] Fix the concept mapping (for some reason we are showing the URI only)

After #1182 , apparently we got a regression in the search auto-complete.

- Closes #1323 

Trying to match the previous design/appearance as much as possible.

## Checklist

### DIFF
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -2126,7 +2126,7 @@ span.date-info {
   float: right;
 }
 
-.tt-selectable > .concept-type {
+.tt-selectable > .concept-type, .autocomplete-label > div:first-child > span.concept-type {
   float: right;
   height: auto;
 }
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -2135,7 +2135,7 @@ span.date-info {
   overflow: hidden;
 }
 
-.autocomplete-label > span {
+.autocomplete-label > div:first-child > span:first-child, .autocomplete-label > div:first-child > span.replaced + span {
   color: var(--medium-color);
   margin: 0;
 }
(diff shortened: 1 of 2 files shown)


kinow:
After the previous commit, here's the auto-complete, now showing the vocabulary information.

image

The URI displayed is the one for General Concept, I reckon. Just need to understand why it is not showing the mapped text instead, and adjust CSS now.

kinow:
I think the highlighted text had the link color? Hard to inspect the HTML element in the browser as finto.fi appears to remove the element (whereas in 2.15-dev, the element is hidden, so I can force it to be displayed again).

image

kinow:
>Not getting labels for concept types seems to be a recurring problem... In this case, I think it could be caused by https://github.com/NatLibFi/Skosmos/issues/678 - try enabling the union default graph setting for your Fuseki.

That fixed it. I simply uncommented that setting in our [config file](https://github.com/NatLibFi/Skosmos/blob/5d193c2365dad60da204656a260f7c9275969eb2/dockerfiles/config/skosmos.ttl#L49) and re-created my environment.

Here's a preview of the final version in this PR:

image

osma (approved):
LGTM, works great! Thanks a lot!

kinow:
> LGTM, works great! Thanks a lot!

Wohoo! Thanks for reviewing it so quickly @osma!

PR #1327, opened 2022-06-01
Combine default graph and optional graphs via UNION to return all the types available in the dataset

### LINKED ISSUE #678: The REST API types query is broken

kinow:

Only the types of the default graph are returned in the REST API, unless the user enables the option in Jena that includes all graphs.

- Closes #678

In the parts of the query where it references types (like rdfs label, or a subclass, etc) I have added a sibling `UNION` that includes the `GRAPH ?g { same_statement }`, to simulate what Jena does with that setting.

With the changes in this branch, after clearing the Local Storage in my browser, I can see all the types in both the REST response, and also in the UI (e.g. search auto-complete, see #1323 ), without needing to enable that option to merge the default graph in Jena Fuseki.

Does it need a test? Does it sound like a valid solution? I thought about doing a simpler `{ { query_as_is } UNION { GRAPH ?g { query _as is } } }` but since the query wasn't very long I opted for this current approach.

## Checklist

### DIFF
--- a/model/sparql/GenericSparql.php
+++ b/model/sparql/GenericSparql.php
@@ -109,10 +109,10 @@ class GenericSparql {
      * @return string
      */
     protected function generateFromClause($vocabs=null) {
-        $clause = '';
         if (!$vocabs) {
             return $this->graph !== '?graph' && $this->graph !== NULL ? "FROM <$this->graph>" : '';
         }
+        $clause = '';
         $graphs = $this->getVocabGraphs($vocabs);
         foreach ($graphs as $graph) {
             $clause .= "FROM NAMED <$graph> ";
@@ -540,36 +540,40 @@ EOQ;
     /**
      * Generates the sparql query for queryTypes
      * @param string $lang
+     * @param string $vocid
      * @return string sparql query
      */
-    private function generateQueryTypesQuery($lang) {
-        $fcl = $this->generateFromClause();
+    private function generateQueryTypesQuery($lang, string $vocid = null): string {
+        $vocabs = isset($vocid) ? [$vocid] : $this->model->getVocabularies();
+        $fcl = $this->generateFromClause($vocabs);
         $query = <<<EOQ
 SELECT DISTINCT ?type ?label ?superclass $fcl
 WHERE {
-  {
-    { BIND( skos:Concept as ?type ) }
-    UNION
-    { BIND( skos:Collection as ?type ) }
-    UNION
-    { BIND( isothes:ConceptGroup as ?type ) }
-    UNION
-    { BIND( isothes:ThesaurusArray as ?type ) }
-    UNION
-    { ?type rdfs:subClassOf/rdfs:subClassOf* skos:Concept . }
-    UNION
-    { ?type rdfs:subClassOf/rdfs:subClassOf* skos:Collection . }
-  }
-  OPTIONAL {
-    ?type rdfs:label ?label .
-    FILTER(langMatches(lang(?label), '$lang'))
-  }
-  OPTIONAL {
-    ?type rdfs:subClassOf ?superclass .
-  }
-  FILTER EXISTS {
-    ?s a ?type .
-    ?s skos:prefLabel ?prefLabel .
+  GRAPH ?g {
+    {
+      { BIND( skos:Concept as ?type ) }
+      UNION
+      { BIND( skos:Collection as ?type ) }
+      UNION
+      { BIND( isothes:ConceptGroup as ?type ) }
+      UNION
+      { BIND( isothes:ThesaurusArray as ?type ) }
+      UNION
+      { ?type rdfs:subClassOf/rdfs:subClassOf* skos:Concept . }
+      UNION
+      { ?type rdfs:subClassOf/rdfs:subClassOf* skos:Collection . }
+    }
+    OPTIONAL {
+      ?type rdfs:label ?label .
+      FILTER(langMatches(lang(?label), '$lang'))
+    }
+    OPTIONAL {
+      ?type rdfs:subClassOf ?superclass .
+    }
+    FILTER EXISTS {
+      ?s a ?type .
+      ?s skos:prefLabel ?prefLabel .
+    }
   }
 }
 EOQ;
@@ -581,7 +585,7 @@ EOQ;
      * @param EasyRdf\Sparql\Result $result
      * @return array Array with URIs (string) as key and array of (label, superclassURI) as value
      */
-    private function transformQueryTypesResults($result) {
+    private function transformQueryTypesResults(\EasyRdf\Sparql\Result $result): array {
         $ret = array();
         foreach ($result as $row) {
             $type = array();
@@ -601,10 +605,11 @@ EOQ;
     /**
      * Retrieve information about types from the endpoint
      * @param string $lang
+     * @param string $vocid
      * @return array Array with URIs (string) as key and array of (label, superclassURI) as value
      */
-    public function queryTypes($lang) {
-        $query = $this->generateQueryTypesQuery($lang);
+    public function queryTypes($lang, string $vocid = null): array {
+        $query = $this->generateQueryTypesQuery($lang, $vocid);
         $result = $this->query($query);
         return $this->transformQueryTypesResults($result);
     }
--- a/tests/GenericSparqlTest.php
+++ b/tests/GenericSparqlTest.php
@@ -491,7 +491,10 @@ class GenericSparqlTest extends PHPUnit\Framework\TestCase
       'http://www.skosmos.skos/test-meta/TestClass' => array(
         'superclass' => 'http://www.w3.org/2004/02/skos/core#Concept',
         'label' => 'Test class'
-      )
+      ),
+      'http://www.w3.org/2004/02/skos/core#Collection' => array(),
+      'http://purl.org/iso25964/skos-thes#ThesaurusArray' => array(),
+      'http://purl.org/iso25964/skos-thes#ConceptGroup' => array()
     );
     $this->assertEquals($expected, $actual);
   }
--- a/tests/ModelTest.php
+++ b/tests/ModelTest.php
@@ -183,7 +183,14 @@ class ModelTest extends PHPUnit\Framework\TestCase
    */
   public function testGetTypesWithoutParams() {
     $result = $this->model->getTypes();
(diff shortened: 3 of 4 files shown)


osma:
Thanks for giving this a shot @kinow , this has been a longstanding issue that has bitten many users (including yourself)!

However, I'm not sure this is the right approach, adding lots of UNION clauses to the query which seems overcomplicated to me. The intent of the `types` query is simply to list all the concept types across all the vocabularies. The result should be similar, if not identical, to the result of performing a `types` query separately for each vocabulary and merging them into one list.

There are at least a few ways of doing that that I can think of:

1. Relying on the union default graph (current status quo). This is really a hack and should be avoided. For one, it relies on the Fuseki setting. Second, this isn't really accurate as the dataset may include other graphs that are not configured as Skosmos vocabularies but may still contain concept type definitions that will end up in the results.
2. Adding a single `GRAPH ?g {` block around the whole query. I *think* that this should be equivalent to the above, at least in sensible scenarios (each graph contains both the type definition and its label). It's still a bit problematic since types from graphs that are not configured as vocabularies may be included.
3. Same as 2, but selecting only those graphs that are configured as vocabularies and using `VALUES ?g` to enumerate those, thus excluding irrelevant graphs.
4. Same idea as 3, but using `FROM NAMED` clauses for selecting individual graphs instead of `GRAPH ?g` with `VALUES ?g`.

Would you like to try some of the approaches 2-4? I think any of them would be an improvement over the current situation, as long as the query still performs well.

kinow:
>Would you like to try some of the approaches 2-4? I think any of them would be an improvement over the current situation, as long as the query still performs well.

Sounds like fun! I will give it a try later after or in between the other JQuery/Bootstrap work. Will be a good way to brush up  SPARQL and Jena Fuseki :)

Do you have any suggestion on which item to try first, between options 2, 3, and 4?

Thanks!

osma:
I think 3 or 4 would be most preferable as they don't suffer from the problem of irrelevant data being picked up from other possible named graphs in the dataset.

model/sparql/GenericSparql.php
```diff
@@ -542,34 +542,38 @@ public function queryConceptInfo($uris, $arrayClass = null, $vocabs = array(), $
      * @param string $lang
      * @return string sparql query
      */
-    private function generateQueryTypesQuery($lang) {
+    private function generateQueryTypesQuery(string $lang): string {
         $fcl = $this->generateFromClause();
+        $fromClause = $this->generateFromClause();
```
kinow, line 547: Went with `FROM NAMED` since we already had a function for creating that clause :tada:

model/sparql/GenericSparql.php
```diff
@@ -542,34 +542,37 @@ public function queryConceptInfo($uris, $arrayClass = null, $vocabs = array(), $
      * @param string $lang
      * @return string sparql query
      */
-    private function generateQueryTypesQuery($lang) {
-        $fcl = $this->generateFromClause();
+    private function generateQueryTypesQuery(string $lang): string {
+        $vocabs = $this->model->getVocabularies();
```
kinow, line 546: If there are no vocabs, then the query will return all the types from all the graphs. Would it be better to return an empty result int that case?

kinow:
I decided to try `FROM NAMED` clauses since we already had the code for that. We actually called the function for that creating the `$fcl` variable, but without passing any vocabularies, so it wasn't returning anything.

Here's an example of what the query may look like. It is from my development environment with a few vocabularies configured.

```sparql
PREFIX  skos: <http://www.w3.org/2004/02/skos/core#>
PREFIX  isothes: <http://purl.org/iso25964/skos-thes#>
PREFIX  rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX  finaf: <http://urn.fi/URN:NBN:fi:au:finaf:>

SELECT DISTINCT  ?type ?label ?superclass
FROM NAMED <http://skos.um.es/unescothes/>
FROM NAMED <http://zbw.eu/stw/>
FROM NAMED <http://vocabs.rossio.fcsh.unl.pt/tesauro/>
FROM NAMED <http://www.yso.fi/onto/yso/>
FROM NAMED finaf:
WHERE
  { GRAPH ?g
      { {   { BIND(skos:Concept AS ?type) }
          UNION
            { BIND(skos:Collection AS ?type) }
          UNION
            { BIND(isothes:ConceptGroup AS ?type) }
          UNION
            { BIND(isothes:ThesaurusArray AS ?type) }
          UNION
            { ?type rdfs:subClassOf/(rdfs:subClassOf)* skos:Concept }
          UNION
            { ?type rdfs:subClassOf/(rdfs:subClassOf)* skos:Collection }
        }
        OPTIONAL
          { ?type  rdfs:label  ?label
            FILTER langMatches(lang(?label), "en")
          }
        OPTIONAL
          { ?type  rdfs:subClassOf  ?superclass }
        FILTER EXISTS { ?s  a               ?type ;
                            skos:prefLabel  ?prefLabel
                      }
      }
  }
```

image

And the REST API response:

```json
{
  "@context": {
    "skos": "http:\\/\\/www.w3.org\\/2004\\/02\\/skos\\/core#",
    "uri": "@id",
    "type": "@type",
    "rdfs": "http:\\/\\/www.w3.org\\/2000\\/01\\/rdf-schema#",
    "onki": "http:\\/\\/schema.onki.fi\\/onki#",
    "label": "rdfs:label",
    "superclass": {
      "@id": "rdfs:subClassOf",
      "@type": "@id"
    },
    "types": "onki:hasType",
    "@language": "en",
    "@base": "http:\\/\\/localhost:9090\\/rest\\/v1\\/"
  },
  "uri": "",
  "types": [
    {
      "uri": "http:\\/\\/www.w3.org\\/2004\\/02\\/skos\\/core#Concept",
      "label": "Concept"
    },
    {
      "uri": "http:\\/\\/zbw.eu\\/namespaces\\/zbw-extensions\\/Descriptor",
      "label": "Descriptor",
      "superclass": "http:\\/\\/www.w3.org\\/2004\\/02\\/skos\\/core#Concept"
    },
    {
      "uri": "http:\\/\\/zbw.eu\\/namespaces\\/zbw-extensions\\/Thsys",
      "label": "Thsys",
      "superclass": "http:\\/\\/www.w3.org\\/2004\\/02\\/skos\\/core#Concept"
    },
    {
      "uri": "http:\\/\\/www.w3.org\\/2004\\/02\\/skos\\/core#Collection",
      "label": "Collection"
    },
    {
      "uri": "http:\\/\\/purl.org\\/iso25964\\/skos-thes#ConceptGroup",
      "label": "Concept group"
    },
    {
      "uri": "http:\\/\\/purl.org\\/iso25964\\/skos-thes#ThesaurusArray",
      "label": "Array of sibling concepts"
    },
    {
      "uri": "http:\\/\\/www.yso.fi\\/onto\\/yso-meta\\/Concept",
      "label": "General concept",
      "superclass": "http:\\/\\/www.w3.org\\/2004\\/02\\/skos\\/core#Concept"
    },
    {
      "uri": "http:\\/\\/www.yso.fi\\/onto\\/yso-meta\\/Individual",
      "label": "Individual concept",
      "superclass": "http:\\/\\/www.w3.org\\/2004\\/02\\/skos\\/core#Concept"
    },
    {
      "uri": "http:\\/\\/www.yso.fi\\/onto\\/yso-meta\\/Hierarchy",
      "label": "Hierarchical concept",
      "superclass": "http:\\/\\/www.w3.org\\/2004\\/02\\/skos\\/core#Concept"
    }
  ]
}
```

tests/GenericSparqlTest.php
```diff
@@ -491,7 +491,10 @@ public function testQueryTypes()
       'http://www.skosmos.skos/test-meta/TestClass' => array(
         'superclass' => 'http://www.w3.org/2004/02/skos/core#Concept',
         'label' => 'Test class'
-      )
+      ),
+      'http://www.w3.org/2004/02/skos/core#Collection' => array(),
+      'http://purl.org/iso25964/skos-thes#ThesaurusArray' => array(),
+      'http://purl.org/iso25964/skos-thes#ConceptGroup' => array()
```
kinow, line 497: Had a look at [`testconfig.ttl`](https://github.com/NatLibFi/Skosmos/blob/master/tests/testconfig.ttl), and I think the function shuold return these new types now as they are types used in vocabularies there. These extra values are added now since the query is searching types in all graphs for all configured vocabs.

tests/ModelTest.php
```diff
@@ -183,7 +183,14 @@ public function testSearchWithNoParams() {
    */
   public function testGetTypesWithoutParams() {
     $result = $this->model->getTypes();
-    $this->assertEquals(array('http://www.w3.org/2004/02/skos/core#Concept', 'http://www.w3.org/2004/02/skos/core#Collection', 'http://purl.org/iso25964/skos-thes#ConceptGroup', 'http://purl.org/iso25964/skos-thes#ThesaurusArray', 'http://www.skosmos.skos/test-meta/TestClass'), array_keys($result));
+    $this->assertEquals(
+        array(
+            'http://www.w3.org/2004/02/skos/core#Concept',
+            'http://www.w3.org/2004/02/skos/core#Collection',
+            'http://purl.org/iso25964/skos-thes#ThesaurusArray',
+            'http://purl.org/iso25964/skos-thes#ConceptGroup',
```
kinow, line 191: The only change in this array is the order of `ThesaurusArray` and `ConceptGroup`. Not sure why these two swapped places in this array.

kinow:
The SonarCloud security issue reported is for a prefix URL, so that shouldn't be a blocker, I think.

model/sparql/GenericSparql.php
```diff
@@ -540,36 +540,40 @@ public function queryConceptInfo($uris, $arrayClass = null, $vocabs = array(), $
...
-    ?type rdfs:label ?label .
-    FILTER(langMatches(lang(?label), '$lang'))
-  }
-  OPTIONAL {
-    ?type rdfs:subClassOf ?superclass .
-  }
-  FILTER EXISTS {
-    ?s a ?type .
-    ?s skos:prefLabel ?prefLabel .
+  GRAPH ?g {
```
osma, line 552: This looks wrong to me. You are already using `$fcl` i.e. a generated block of FROM clauses. There is no need for the `GRAPH ?g {` block.

osma:
I can try to finish this off now that @kinow you are busy with the qtip-to-CSS PR #1324 ...

kinow:
> I can try to finish this off now that @kinow you are busy with the qtip-to-CSS PR #1324 ...

That'd be great, please :D Thank you @osma!

PR #1332, opened 2022-06-04
API docs: Add 'uri' parameter to groupMembers

danmichaelo:
The `uri` parameter was listed in https://github.com/NatLibFi/Skosmos/wiki/REST-API#vocidgroupmembers , but seems like it didn't make it over to the swagger format.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -1237,6 +1237,13 @@
             "type": "string"
           },
           {
+            "name": "uri",
+            "in": "query",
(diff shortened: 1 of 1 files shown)


kinow (approved):

osma:
The issue that this PR fixes was reported in #1333, already closed.

LGTM, I will merge this now and also add it to the `v2.15-maintenance` branch.

Thanks a lot @danmichaelo !

PR #1337, opened 2022-06-08
(External) Authentication layer

YOUR1:
We are in need of a Skosmos instance that is not available for the public, but only available through an authentication layer (SimpleSamlPHP in our case). 

Added an authentication layer, with minimal changes in the existing model and controller classes.


osma:
Thanks for the PR @YOUR1 and sorry for the somewhat late response.

Can you explain why you decided to create an authentication layer in PHP? The alternative would be to configure Apache with [authentication and authorization](https://httpd.apache.org/docs/2.4/howto/auth.html) using one of the many `mod_auth*` modules. Of course this has its own limitations but at least it should be bullet-proof once configured since it protects a whole directory / URL space.

YOUR1:
> Thanks for the PR @YOUR1 and sorry for the somewhat late response.
> 
> Can you explain why you decided to create an authentication layer in PHP? The alternative would be to configure Apache with [authentication and authorization](https://httpd.apache.org/docs/2.4/howto/auth.html) using one of the many `mod_auth*` modules. Of course this has its own limitations but at least it should be bullet-proof once configured since it protects a whole directory / URL space.

No problem. Using mod_auth implies that you are using apache2 as a webserver. That's obviously not always the case. Also; as you mentioned - mod_auth has its own limitations. We are using SimpleSamlPHP as a authentication backend for some of our clients; and mod_auth couldn't support that the way we wanted. Also; this adds more flexibility to add more/different authentication methods.

osma:
> Authentication as part of Skosmos would make more sense if only some vocabularies are non-public.

Skosmos (and its REST API) still has global search and other ways to access combinations of vocabularies. Having to hide some vocabularies entirely from unauthorized would take a lot of work. Right now the assumption is that everything is public information.

nichtich:
> Having to hide some vocabularies entirely from unauthorized would take a lot of work. 

So then there is no need to add authentication functionality into Skosmos: the question of access is better solved on a different layer (e.g. webserver/proxy) than the PHP code of Skosmos. Just my 2 cents.

PR #1354, opened 2022-09-06
Fix header bar search form

### LINKED ISSUE #1345: Search function does not limit the search to a single vocabulary
In the search function on the Finto front page, after selecting a vocabulary and while typing a search term, the search result is not limited to the chosen vocabulary, but instead the search shows occurrences from all the vocabularies.

http://finto.fi/en/?clang=en

1. Select the vocabulary you want from the list in the search function
2. Start to type a search term
3. View the list that appears

Only results related to the selected vocabulary

Firefox

joelit:

Cookies are not set correctly for things like multiple vocabulary selection in the global search field. This would break the global search UI element, restricting the use of global search to only skosmos API or writing the search parameters directly to the URL.

- Closes #1345 

- scripts.js sets the cookie correctly, using the newly required SameSite attribute (http and https are considered to be different websites)
  - More info: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite
- docready.js reads the selected vocabulary values correctly
- headerbar.twig passes parameter separators correctly

The search vocabulary parameter for "all vocabularies" is not set correctly in the cookie.

## Checklist

### DIFF
--- a/resource/js/docready.js
+++ b/resource/js/docready.js
@@ -942,7 +942,7 @@ $(function() { // DOCUMENT READY
   // activating the custom autocomplete
   function updateVocabParam() {
     vocabSelectionString = '';
-    var $vocabs = $('li > a.active input');
+    var $vocabs = $('button.active input');
     $.each($vocabs,
       function(index, ob) {
         if (ob.value === 'multiselect-all') {
--- a/resource/js/docready.js
+++ b/resource/js/docready.js
@@ -580,7 +580,7 @@ $(function() { // DOCUMENT READY
     createCookie('SKOSMOS_SEARCH_LANG', qlang, 365);
   }
 
-  $('.lang-button').on('click', function() {
+  $('a.dropdown-item').on('click', function() {
     qlang = $(this)[0].attributes.hreflang ? $(this)[0].attributes.hreflang.value : 'anything';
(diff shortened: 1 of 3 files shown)


joelit:
When fixing the language setting for the same menu, it turned out the component which writes the search query url parameters hasn't worked properly beforehand, as the paremeters aren't separeted and the parameter values are not encoded.
Screenshot from 2022-09-07 11-29-29
Skosmos tries to pre-fill the search query in the text field and fails parsing it.

osma (changes requested):
I tested this and couldn't make it not work :) Looks good, apart from the issue that "select all" doesn't set the cookie, which you already noted. I suggested a fix in the above comment. It's also possible to consider that out of scope and file a separate issue about it.

joelit:
The fix was a pretty simple one; removing the selected vocabularies and calling the updateVocabParam() function seemed to do the trick, as the empty selectedVocabs JS variable and no SKOSMOS_SELECTED value in the cookie is interpreted as having all the vocabularies chosen. :)

osma (approved):
Looks good and works perfectly! :tada: 

Before merging, I suggest changing the PR title to something more descriptive. This fixes many problems related to global search, not just setting cookies.

PR #1356, opened 2022-09-07
Graceful handling of partial SKOS XL data for concept labels

### LINKED ISSUE #1346: Partial SKOS XL label data breaks headline on concept page

https://dev.finto.fi/yso/fi/page/p16810

(not visible right now as the offending SKOS XL data was removed)

1. Add SKOS XL data for one of the prefLabels of a concept (e.g. Northern Sami prefLabel in this case)
2. Look at the concept page in a different content language (e.g. Finnish)

Expected to see a normal headline showing the preferred label in the chosen content language. But instead the headline is missing:

image

(originally reported by @nikopartanen)

Firefox

osma:

This PR changes the way SKOS XL data for labels is handled. The old code expected that if a concept has _some_ SKOS XL labels, it will have SKOS XL data for **all** labels. This isn't the case in current YSO where SKOS XL information is added only for a small number of labels. This PR makes the checks for SKOS XL data more careful. Result now looks like this (compare with issue #1346):

image

- Closes #1346

* Concept.hasXlLabel() and ConceptPropertyValueLiteral.hasXlLabel() methods changed to be more specific so that they only return `true` when there is non-trivial SKOS XL data available specifically for that concept (prefLabel) / literal value
* some logic (filtering of trivial properties `rdf:type` and `skosxl:literalForm`) moved from Twig templates to PHP code
* use `dc:` namespace for gettext lookups for properties in the `dc11:` namespace also for properties of SKOS XL Labels (this is already done for e.g. regular properties on concepts and it fixes issues displaying the correct labels for Dublin Core metadata)

* when displaying properties of SKOS XL labels with literal values (e.g. `dc:source "Kansalliskirjasto"@fi` in above screenshot), there is no mechanism for trying to find and select the value with the correct language tag; so if there were many similar values in different languages, only one of them would be shown
* I noticed that if I hover over the "info" symbol many times, eventually the tooltip will become empty; this could be a bug in qtip2 but it should go a way if/when we replace it with CSS (see PR #1324)

## Checklist

### DIFF
--- a/model/Concept.php
+++ b/model/Concept.php
@@ -163,12 +163,10 @@ class Concept extends VocabularyDataObject implements Modifiable
         return "";
     }
 
-    public function hasXlLabel($prop = 'prefLabel')
+    public function hasXlLabel()
     {
-        if ($this->resource->hasProperty('skosxl:' . $prop)) {
-            return true;
-        }
-        return false;
+        $xlLabel = $this->getXlLabel();
+        return ($xlLabel !== null && !empty($xlLabel->getProperties()));
     }
 
     public function getXlLabel()
--- a/model/ConceptPropertyValueLiteral.php
+++ b/model/ConceptPropertyValueLiteral.php
@@ -87,9 +87,8 @@ class ConceptPropertyValueLiteral extends VocabularyDataObject
 
     public function hasXlProperties()
     {
-        $graph = $this->resource->getGraph();
-        $resources = $graph->resourcesMatching('skosxl:literalForm', $this->literal);
-        return !empty($resources);
+        $xlLabel = $this->getXlLabel();
+        return ($xlLabel !== null && !empty($xlLabel->getProperties()));
     }
 
     public function getXlLabel()
--- a/model/LabelSkosXL.php
+++ b/model/LabelSkosXL.php
@@ -24,8 +24,11 @@ class LabelSkosXL extends DataObject
         $ret = array();
         $props = $this->resource->properties();
         foreach($props as $prop) {
-            if ($prop !== 'skosxl:prefLabel') {
-                $ret[$prop] = $this->resource->get($prop);
+            if ($prop !== 'rdf:type' && $prop !== 'skosxl:literalForm') {
+                // make sure to use the correct gettext keys for DC namespace
+                $propkey = str_starts_with($prop, 'dc11:') ?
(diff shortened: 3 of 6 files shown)


joelit (changes requested):
The handling of concept.xlLabel.properties tooltip for the title row seems to be broken. Examples in the comments. OTOH, this could be merged as a fix to the original bug of concept preflabels not being shown at all.

joelit:
I tested his with SKOS-XL data and it is indeed an improvement - the labels work now but the tooltips don't.

With concept data looking like this:

```
yso:p6095 a skos:Concept,
        yso-meta:Concept ;
    dct:created "1985-01-01"^^xsd:date ;
    skos:altLabel "Betula"@en,
        "birch"@en,
        "koivu"@fi,
        "björk"@sv ;
    skos:broader yso:p14339,
        yso:p3425 ;
    skos:closeMatch <http://id.loc.gov/authorities/subjects/sh85014281>,
        allars:Y16158,
        ysa:Y96273 ;
    skos:inScheme yso: ;
    skos:narrower yso:p16291,
        yso:p26048,
        yso:p6094 ;
    skos:prefLabel "birches"@en,
        "koivut"@fi,
        "soahki"@se,
        "björkar"@sv ;
    skos:related yso:p21624 ;
    skosxl:prefLabel <http://www.yso.fi/onto/yso-meta/p6095#se_1> .

<http://www.yso.fi/onto/yso-meta/p6095#se_1> a skosxl:Label ;
    dc:source "Lähde Giellagáldu"@fi,
        "Gáldu Giellagáldu"@se ;
    skos:inScheme yso: ;
    skosxl:literalForm "soahki"@se .
```
    
The result in Skosmos UI looks like this. The labels work and Skosmos picks the right source information from the skosxl:Label object. The tooltips work when they are displayed in the concept information (foreign label info in this case), but the tooltips won't work as part of the title information.
    
Screenshot from 2022-09-07 17-20-12
Screenshot from 2022-09-07 17-20-27

osma:
Thanks for the review and testing @joelit !

I think what you are seeing is a different bug, probably having more to do with qtip2 than the changes in this PR. I hope that replacing qtip2 with CSS in PR #1324 will help solve that issue.

It did work for me, at least briefly (on Firefox):

image

PR #1357, opened 2022-09-07
Use the --font-size variable for versal, and set width: auto for the restore-breadcrumbs link

### LINKED ISSUE #1355: "show all X paths" message misplaced when concept has many paths

https://finto.fi/koko/en/page/p32086

1. Go to a concept with many (>4) paths in the hierarchy (e.g. above URL)

Expected to see four possible breadcrumb paths and the `[show all X paths]` link below them. Instead, the message appears in between the first and second paths:

image

This was likely caused by the Bootstrap 5 upgrade (ping @kinow)

(originally found by @kouralex )

Firefox

kinow:

Wrong formatting for the link to show more breadcrumbs.

- Closes #1355 

## Checklist

### DIFF
--- a/resource/js/scripts.js
+++ b/resource/js/scripts.js
@@ -250,6 +250,9 @@ function hideCrumbs() {
       $($crumbs[i]).addClass('hidden-path');
     }
     if ($('.restore-breadcrumbs').length === 0) {
+      // We attach the link to display all breadcrumbs after the zero-th element
+      // to force the browser to re-arrange the elements. If we moved that to the
+      // bottom, then we wouldn't have an easy way to re-arrange it at the top.
       $($crumbs[0]).after('<a class="versal restore-breadcrumbs" href="#">[' + expand_paths.replace('#',($crumbs.length)) + ']</a>');
     }
   }
--- a/resource/css/styles.css
+++ b/resource/css/styles.css
@@ -1627,11 +1627,13 @@ li.sub-group {
 
 .restore-breadcrumbs {
   background-color: var(--white-color);
-  float: right;
+  width: auto;
+  position: absolute;
+  right: 0;
+  display: inline;
 }
 
 .crumb-path {
-  float:left;
   width: 100%;
 }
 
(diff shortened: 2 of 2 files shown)


kinow:
Needs testing to confirm the changes look OK (re-creating my Docker environment).

osma:
This is how it looks to me now (both Firefox and Chromium, screenshot from Chromium), and where I think the link should go:

image

Also, the text size still doesn't match the breadcrumbs, it's a little bit too large.

kinow:
Finished setting up my Docker environment, and then loading some sample data. Got the same result as you did @osma, although the font at least is looking better for me (could it be the browser cache?).

Screenshot from 2022-09-07 19-44-43

kinow:
@osma, how's it looking now?

Screenshot from 2022-09-07 22-46-19

Going to hit the sack now, but I can still work on this tomorrow :+1: 

Cheers
Bruno

osma:
Looks great on your screenshot now! I will test it locally as well.

resource/js/scripts.js
```diff
@@ -250,6 +250,9 @@ function hideCrumbs() {
       $($crumbs[i]).addClass('hidden-path');
     }
     if ($('.restore-breadcrumbs').length === 0) {
+      // We attach the link to display all breadcrumbs after the zero-th element
+      // to force the browser to re-arrange the elements. If we moved that to the
+      // bottom, then we wouldn't have an easy way to re-arrange it at the top.
```
kinow, line 255: @osma turns out `$crumbs[0]` was correct. But it's not the best way to have the link at the right top corner.

What we are doing, is injecting the element after the first item, to let the browser resize it. Previously it was doing it with `float: left` + `float: right`, and some width setting here and there.

But with a parent CSS flex element it gets trickier. For the future I think it would be good to restructure it so that the design was better enforced without relying on how the browser re-arranges elements. But not needed for now if the fix in this PR fixes it :+1:

kinow: (Leaving the note to myself or other devs that work on this code)

osma:
Works perfectly, I will merge this. Thanks a lot @kinow !

PR #1360, opened 2022-09-13
Issue1347 malihu custom scrollbar fix

### LINKED ISSUE #1347: Upgrade/replace malihu-custom-scrollbar ahead of jQuery upgrade
## The current state of the package
https://github.com/malihu/malihu-custom-scrollbar-plugin
(Last update in Nov 2016)

## Suggested alternatives
https://www.jqueryscript.net/blog/Best-Custom-Scroll-Bar-jQuery-Plugins.html
https://openbase.com/js/malihu-custom-scrollbar-plugin/alternatives
Implementing the scrollbar in pure CSS might be worth exploring.

Vainonen:
Unmaintained Malihu scrollbar plugin outdated when jQuery gets updated to version 3.6

- Closes #1347 

Malihu scrollbar plugin removed and scrolling functionality done without any external libraries. This PR is partly based on @henriyli's previously closed PR of this issue: https://github.com/NatLibFi/Skosmos/pull/1175.

- When list of new and deprecated concepts is scrolled and at least one API request made, clicking the tab "new and deprecated" results to an empty list
- Scroll bar color is changed as mentioned in the previous PR of this issue: https://github.com/NatLibFi/Skosmos/pull/1175.
- tab header anchor should be made inactive if tab listing is shown
- addresses in browser address bar do not always change when changing between sidebar tabs (e. g. if in "new and deprecated" tab, clicking "hierarchy" tab leaves address in browser unchanged)
- refactoring of code and renaming of variables more conviniently needed

## Checklist


resource/css/styles.css
```diff
@@ -2227,6 +2227,29 @@ body, .versal, h1, h2, h3, p, .versal-bold {
   to   { opacity: 1; }
 }
 
+/* CSS custom scrollbars
+ ************************************/
+ :root {
+  --scrollbarWidth: 14px;
+  --scrollbarBg: #b9c1c6;
+  --scrollbarThumb: #474b4f;
+}
```
kinow, line 2236: This file already has a :root block elsewhere with colors, font sizes, etc. Maybe worth keeping the root variables all in a single place.

kinow (commented):
Haven't tested it but looked at the code from GH UI and it's looking really good!!! One minor comment.

Vainonen:
A new question/issue https://github.com/NatLibFi/Skosmos/issues/1361 made after commit [Sidebar active nav tab made unclickable](https://github.com/NatLibFi/Skosmos/pull/1360/commits/5099078fbc192016fd761884d28183bf8fe16c57). Making current tab inactive produces a feature that the previous tab is inactive as long as next tab content is loading. Is this feature bearable or a bug to be fixed?

resource/css/styles.css
joelit: Yep. And the variable names should be brought in line with the naming and conventions of the other css variables. To nudge this in the right direction, the beginning of the :root element in styles.css could look like this, with the corresponding changes for the variable names in styles.css row 2237 onwards:

```
:root {
  /* Color definitions */
  --dark-color: #394554; /* topbar nav link text, dropdown text, buttons */
  --secondary-dark-color: #3C535B; /* disabled buttons, bg of "content language" dropdown label in header */
  --medium-color: #00748F; /* used for skosmos service name (top left logo), link texts (a), header bar, hovered dropdown menu items, placeholders in form input fields */
  --light-color: #D4EDEB; /* used for concept property dividers */
  --link-hover-color: #23527C; /* used for hovered links, clipboard icon */
  --tooltip-bg-color: #1E1E1E;
  --tooltip-text-color: #00ACD3;
  --tooltip-border-color: #75889F; /*also used as typeahead focus border */
  --main-bg-color: #D5DBDE;
  --alert-color-pale: #D95F8A;
  --alert-color-bright: #ED0D6C;
  --uri-id-text-color: #006621; /* only used in search results */

  --scroll-box-color: #474b4f; /* used for the sidebar scroll */
  --scroll-bar-color: #b9c1c6; /* used for the sidebar scroll */
  --scroll-bar-width: 14px; /* used for the sidebar scroll */
```

resource/js/docready.js
```diff
@@ -96,6 +96,12 @@ $(function() { // DOCUMENT READY
     });
 
     var active_tab = $('a.active').parent().attr('id');
+
+    active_tab == 'alpha' ? $('#alpha').css('pointer-events','none') : $('#alpha').css('pointer-events','auto');
+    active_tab == 'changes' ? $('#changes').css('pointer-events','none') : $('#changes').css('pointer-events','auto');
+    active_tab == 'groups' ? $('#groups').css('pointer-events','none') : $('#groups').css('pointer-events','auto');
+    active_tab == 'hierarchy' ? $('#hierarchy').css('pointer-events','none') : $('#hierarchy').css('pointer-events','auto');
+
```
joelit, line 104: Would this achieve the same result?

```
    var active_tab = $('a.active').parent().attr('id');
    $('#'+active_tab).css('pointer-events','none') ;
```

joelit (changes requested):
* I made a couple of suggestions for tweaking the CSS jand JS
* The bigger issue seems to be that the nav tabs of the sidebar are now triggering full page loads instead of partial ones
  * Also, the anchor of the hierarchy tab points to the Skosmos front page, omitting the vocabulary ID. It's not a big thing to change - but first and foremost the page load triggered by the nav tab change should be a partial one.
  * This might also cause the problem with choosing another tab while the sidebar content is loading resulting in an empty concept info like this:
Screenshot from 2022-09-16 11-37-14

resource/js/docready.js
miguelvaara: The code you propose is more elegant and we will go with it

joelit (approved):
Apparently the clunkier JS was working better so better leave it like it was. Also, the asynchronous page loading was broken due to mismatching jQuery version.

Vainonen:
Thank you @joelit & @kinow for reviewing and comments!

PR #1365, opened 2022-09-16
Remove jQuery UI dependency

osma:

The jQuery Migrate plugins prints deprecation warnings about jQuery UI. We are currently using jQuery UI 1.12.1.

It would be possible to upgrade to jQuery UI 1.13.x, but it is not available via Packagist / Composer.

But it turns out we're not actually using any(?) functionality of jQuery UI. It was probably used in the past, but since then, UI widgets have been rewritten using other libraries. So this PR simply removes jQuery UI.

- related to #1151

* drop jQuery UI dependency in composer.json
* drop references to jQuery UI JS and CSS files from HTML templates

Not 100% sure there isn't some functionality that is still based on jQuery UI that I just didn't find.

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,6 @@
   },
   "require": {
     "components/jquery": "3.6.*",
-    "components/jqueryui": "1.12.*",
     "components/handlebars.js": "v4.7.7",
     "davidstutz/bootstrap-multiselect": "v1.1.1",
     "easyrdf/easyrdf": "1.1.*",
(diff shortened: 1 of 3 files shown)


osma:
I've tested most UI functionality without jQuery UI and couldn't find anything broken.
I'll merge this into the `upgrade-jquery-3.6` branch and hope that nothing breaks. Or if it does, we will notice it soon enough.

PR #1369, opened 2022-09-28
Modify counting for search limit in typeahead (patching it locally)

### LINKED ISSUE #1367: Search box autocomplete does not give a full list of results
https://finto.fi/yso/fi/

1. Go to the frontpage of YSO vocabulary at finto.fi
2. Use the search box to type "sosiaalinen", wait for the autocomplete list to open
3.

Expected to see the full list of results (according to our vocabulary experts), instead I see only 29 first results 

image

Maybe something we broke during Bootstrap update?

Firefox (102.3.0esr (64-bit))

kinow:

- Closes #1367

Monkey patching typeahead to simulate as if https://github.com/twitter/typeahead.js/pull/1774 had been merged.

It's hard to write a test for this change, sorry. Only way to review and verify is really deploying somewhere with YSO to compare the search results.

## Checklist

### DIFF
--- a/resource/js/docready.js
+++ b/resource/js/docready.js
@@ -742,6 +742,55 @@ $(function() { // DOCUMENT READY
     var typeaheadInstance = $typeahead.data("ttTypeahead");
     typeaheadInstance.menu.off("mouseenter.tt", ".tt-suggestion");
     typeaheadInstance.menu.off("mouseleave.tt", ".tt-suggestion");
+    // Monkey-patching for:
+    // - https://github.com/twitter/typeahead.js/pull/1774
+    // - https://github.com/twitter/typeahead.js/blob/0700b186e98401127b051302365e34b09def2285/src/typeahead/dataset.js#L265-L278
+    var update = function update(query) {
+      var that = this, canceled = false, syncCalled = false, rendered = 0;
+
+      // cancel possible pending update
+      this.cancel();
+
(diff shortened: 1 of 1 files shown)


osma:
Whoa, this is awesome @kinow ! You sure have earned a :medal_sports: for this feat! 

I tested this and it seems to work, at least for the known example cases. There's no way for me to tell whether this will cause any new troubles, but having to choose between known-buggy and possibly-buggy-but-seems-to-work, I think the latter is a much better option.

It's clear we need to replace Typeahead.js, and I think this will be done rather soon as we are going to rebuild the UI for 3.0 anyway. But this solves the problem for now so I'll merge it! :+1:

kinow:
>Whoa, this is awesome @kinow ! You sure have earned a medal_sports for this feat!

:tada: hooray! 

>I tested this and it seems to work, at least for the known example cases. There's no way for me to tell whether this will cause any new troubles, but having to choose between known-buggy and possibly-buggy-but-seems-to-work, I think the latter is a much better option.

:+1: agreed.

> It's clear we need to replace Typeahead.js, and I think this will be done rather soon as we are going to rebuild the UI for 3.0 anyway. But this solves the problem for now so I'll merge it! +1

Writing from scratch is a little tricky, but if there's some reactive framework (react/vue/svelte/or even those smaller micro frameworks) it could save a few lines.

Thanks @osma!

PR #1371, opened 2022-10-04
Tooltips are now conditional on the hierarchy

kinow:

- Addresses https://github.com/NatLibFi/Skosmos/pull/1324#issuecomment-1266708112

Hierarchy is not always available. When it is not, we want to display a tooltip with a warning/note about it. However, the tooltips were always enabled, causing an empty tooltip sometimes.

Hard to test. I looked at the previous modified files in the linked PR, then repeated the same fix. Tested on Firefox with the :hover modified on the hierarchy tab. In one case the tooltip is displayed as hierarchy is disabled, and in the other nothing is shown.

image
image

Hopefully that fixes the issue. Sorry!

## Checklist

### DIFF
--- a/view/changes.twig
+++ b/view/changes.twig
@@ -12,7 +12,10 @@
     {% endif %}
     <h3 class="sr-only">{% trans "List vocabulary concepts hierarchically" %}</h3>
     {% if active_tab == 'hierarchy' %}{% set css_class = css_class|merge(['active']) %}{% endif %}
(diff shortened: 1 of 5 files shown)


osma:
Wow, thanks a lot for the quick fix @kinow ! I will test this and if it works OK, I will merge it.

kinow:
> Wow, thanks a lot for the quick fix @kinow ! I will test this and if it works OK, I will merge it.

Not a problem @osma. Good to fix this while it's still fresh on my (normally short-term) memory. Let me know if that works :+1:

osma:
Works perfectly! Thanks a lot @kinow !

PR #1372, opened 2022-10-04
fix auto-scrolling and highlighting current concept in hierarchy

osma:

The custom scrollbar implementation was removed in PR #1360. But the PR also removed some functionality from the hierarchy tree: highlighting the current concept and automatically scrolling so that the current concept (or at least one of the places where it appears) is visible. This PR restores those functionalities. It also cleans up some leftover JS code.

- Follow-up fix to #1360

* select the active node(s) in the hierachy after the page has loaded
* automatically scroll the hierarchy display so that the active node is visible and centered (this is now done using the standard [Element.scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) method instead of using mCustomScrollbar methods)
* remove the getLeafOffset method that was used by the previous scrolling method but is now unnecessary

The auto-scrolling now works a bit differently than in 2.15: it may not scroll exactly the same amount, and the scrolling is smooth instead of instant. I like the smooth scrolling better but this can be easily disabled by adjusting the `behavior` option of `scrollIntoView`.

## Checklist

### DIFF
--- a/resource/js/hierarchy.js
+++ b/resource/js/hierarchy.js
@@ -36,19 +36,13 @@ function invokeParentTree(tree) {
     // TODO: a good fix would mean fixing the underlying DOM structure
     $('.jstree-container-ul').parent().addClass('jstree-default');
 
+    var $leafProper = $('.jstree-leaf-proper');
+    if ($leafProper.length > 0) {
+      var $sidebarGrey = $(".sidebar-grey");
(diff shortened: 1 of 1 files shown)


joelit (approved):
Works fine with Chrome, Firefox, and Edge!

PR #1383, opened 2022-10-26
Avoid slow HTML parsing in hierarchy sort function

### LINKED ISSUE #1377: Opening the hierarchy for a concept takes a long time

https://finto.fi/yso/en/page/p28785

1. Load the page with the network inspector open
2. The page content loads in a reasonable time, but the hierarchy takes a long time to open
3. This is shown in the network inspector as the time to finish, for example:
- Chrome: Finish: **5.75 s** DOMContentLoaded: 551 ms Load: 5.67 s
-  Firefox: Finish: **7.06 s** DOMContentLoaded: 583 ms load: 1.01 s
- Edge: Finish: **9.97 s** DOMContentLoaded: 691 ms Load: 9.78 s
(Firefox seems to interpret load time differently to Chrome and Edge)

This lag time seems to be consistent to certain concept pages. My initial hunch was that this would be proportional to the depth of the concept in the hierarchy tree, but that doesn't seem to be the case. Something like https://finto.fi/yso/en/page/p7873 has more than double the hierarchy depth while keeping the time to finish much shorter.

Firefox, Chrome & Edge

osma:

The rendering of the concept hierarchy was very slow in some cases, see #1377.

- Closes #1377

* attach the labels to jsTree nodes representing concepts as a separate attribute `label` (similar to e.g. `uri` and `notation`)
* when sorting nodes, look at the label attribute instead of parsing the HTML code to find it
* remove extra trailing whitespace from a few lines that are directly adjacent to the above changes

This PR changes the functions `createCpnceptObject`, `createObjectsFromChildren`, `vocabRoot` and `createObjectsFromNarrowers` which are also [overridden](https://github.com/NatLibFi/Finto-data/blob/cd411bdf80c2e74af63be9dd73f92c871cba287c/conf/dev.finto.fi/footer.inc#L48-L141) in the `footer.inc` file which is part of the Finto customizations. If you have installed the customizations as they are currently written, this code will not work properly because the `label` attribute will be missing from nodes. The changes need to be applied to `footer.inc` as well. I will do that shortly.

## Checklist

### DIFF
--- a/resource/js/hierarchy.js
+++ b/resource/js/hierarchy.js
@@ -60,7 +60,8 @@ function createObjectsFromChildren(conceptData, conceptUri) {
   var childArray = [];
   for (var i = 0; i < conceptData.narrower.length; i++) {
     var childObject = {
-      text: getLabel(conceptData.narrower[i]), 
+      text: getLabel(conceptData.narrower[i]),
+      label: pickLabel(conceptData.narrower[i]),
       a_attr: getHrefForUri(conceptData.narrower[i].uri),
       uri: conceptData.narrower[i].uri,
       notation: conceptData.narrower[i].notation,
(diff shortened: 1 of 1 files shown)


kinow (approved):

osma:
Adjusted the `footer.inc` overrides for dev.finto.fi in [this commit](https://github.com/NatLibFi/Finto-data/commit/12b42af3d686cd47dabc94fca4b75214604c6c92). I will merge this now so it will be available in dev.finto.fi.

Issue #1381, opened 2022-10-21
Raising a TypeError on method_exists() in PHP 8.x


KlausIllmayer:
## At which URL did you encounter the problem?
Internal test instance of Skosmos, running with PHP 8.0.8 and Skosmos release version 2.16

## What steps will reproduce the problem?
1. Have Skosmos 2.16 running with PHP 8.x
2. Have a vocabulary with a concept that has a skos:notation with an integer value, e.g. 35
3. View the concept

## What is the expected output? What do you see instead?
Seeing the detail view of the concept, instead there is a blank view. In the apache log files I get this error:

<pre>
PHP Fatal error:  Uncaught TypeError: method_exists(): Argument #1 ($object_or_class) must be of type object|string, int given in /var/www/html/model/Concept.php:663
</pre>

## What browser did you use? (eg. Firefox, Chrome, Safari, Internet explorer)
Firefox but this is not relevant here.

## How to solve the issue?
Since PHP 8.x the behaviour of method_exists() changed, now it raises a TypeError if it is not an object/string, which is not the case in PHP 7.x (see e.g. https://stackoverflow.com/questions/70584290/php-8-0-method-exists-on-non-object-causes-fatal-typeerror) - I've also compared the behaviour with our [production instance](https://vocabs.acdh.oeaw.ac.at/dha_taxonomy/en/page/?uri=https%3A%2F%2Fvocabs.acdh.oeaw.ac.at%2Fdhataxonomy%2FConcept35) running an older version of Skosmos and a PHP 7.x, where this error does not occur.

In https://github.com/NatLibFi/Skosmos/blob/v2.16-maintenance/model/Concept.php#L663 `method_exists` is called without checking if it is an object. In case of an integer the TypeError is raised, as it is the case with a `skos:notation 35`. Adding such a check with `is_object()` in the line 663 solves the issue in my understanding:
<pre>
                $propertyValues[(is_object($label) && method_exists($label, 'getValue')) ? $label->getValue() : $label][] = $value->getType();
</pre>

Hope, that this is the correct solution - if you prefer, I can also create a pull request.

PR #1385, opened 2022-11-03
Add support for PHP 8.1

### LINKED ISSUE #1274: Tests fail on PHP 8.1

Any recent CI jobs, e.g. https://github.com/NatLibFi/Skosmos/runs/5185995020?check_suite_focus=true

1. Try to run unit tests on PHP 8.1

Expected: Tests should pass.
Reality: "There were 474 errors"

The main culprit seems to be EasyRdf, which isn't yet compatible with PHP 8.1, but there's an [open PR](https://github.com/easyrdf/easyrdf/pull/384) which should hopefully fix this before long.

kouralex:
Skosmos currently supports PHP versions 7.3 - 8.0 even though 7.3 EOL has already been reached and 7.4 is nearing its own end, see https://www.php.net/supported-versions.php. Therefore, there is a growing need of new version support.
This PR straighforwardly fixes incompatibility issues (and deprecation notices) encountered whilst developing and testing Skosmos with PHP 8.1. I used the proposed solution described in https://github.com/NatLibFi/Skosmos/issues/1274#issuecomment-1288956384 to get past EasyRdf-related issues. Closes #1274 
* I was a bit uncertain about how to replace `FILTER_SANITIZE_STRING` (see https://www.php.net/manual/en/filter.filters.sanitize.php for more information, also see https://stackoverflow.com/questions/69207368/constant-filter-sanitize-string-is-deprecated and especially its accepted answer). It could be more appropriate to use the 3rd option of the accepted answer via a FILTER_CALLBACK, for example.
* I have *not* used a systematic approach (e.g., PHPStan) for checking out lingering issues in code base. This means that there may still reside some issues in the code base as the tests currently do not cover for everything. For instance, I discovered some not-so-apparent issues just by testing the autocomplete search, issues (deprecation) that did not appear whilst running the test kit.
* This PR does not address #1381
## Checklist


joelit:
Hooray for sweetrdf! Core functionalities need to be checked to be sure there are no regressions - I don't 100% trust the coverage of tests in this case.

joelit (changes requested):
Excellent work! :)

I gave this a thorough run-through, and here are my findings:

 - EasyRDF seems to work fine
 - Of the things not covered by PHPUnit, I tested that the update does not break the way Skosmos interacts with JS components like the JS tree or the search bar
 - Finna-plugin seems to work too
 - The new sanitation seems to work, seeing as the following special chars work with SPARQL
   - http://urn.fi/URN:NBN:fi:au:finaf:000083892
   - http://www.yso.fi/onto/yso/p8723
   - http://www.yso.fi/onto/yso/p13591
 
 But there is something wrong with library dependencies to AssetCollection for robloach/component-installer:
> Deprecation Notice: Return type of Assetic\Asset\AssetCollection::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/Skosmos/vendor/kriswallsmith/assetic/src/Assetic/Asset/AssetCollection.php:215
> Deprecation Notice: Return type of Assetic\Filter\FilterCollection::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/Skosmos/vendor/kriswallsmith/assetic/src/Assetic/Filter/FilterCollection.php:73
> Deprecation Notice: Return type of Assetic\Filter\FilterCollection::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/Skosmos/vendor/kriswallsmith/assetic/src/Assetic/Filter/FilterCollection.php:78
> Deprecation Notice: Return type of Assetic\Asset\Iterator\AssetCollectionFilterIterator::getChildren() should either be compatible with RecursiveFilterIterator::getChildren(): ?RecursiveFilterIterator, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/Skosmos/vendor/kriswallsmith/assetic/src/Assetic/Asset/Iterator/AssetCollectionFilterIterator.php:80
> Deprecation Notice: Return type of Assetic\Asset\Iterator\AssetCollectionFilterIterator::accept() should either be compatible with FilterIterator::accept(): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/Skosmos/vendor/kriswallsmith/assetic/src/Assetic/Asset/Iterator/AssetCollectionFilterIterator.php:50
> Deprecation Notice: Return type of Assetic\Asset\Iterator\AssetCollectionIterator::hasChildren() should either be compatible with RecursiveIterator::hasChildren(): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/Skosmos/vendor/kriswallsmith/assetic/src/Assetic/Asset/Iterator/AssetCollectionIterator.php:104
> Deprecation Notice: Return type of Assetic\Asset\Iterator\AssetCollectionIterator::getChildren() should either be compatible with RecursiveIterator::getChildren(): ?RecursiveIterator, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/Skosmos/vendor/kriswallsmith/assetic/src/Assetic/Asset/Iterator/AssetCollectionIterator.php:112
> Deprecation Notice: Return type of Assetic\Asset\Iterator\AssetCollectionIterator::current($raw = false) should either be compatible with Iterator::current(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/Skosmos/vendor/kriswallsmith/assetic/src/Assetic/Asset/Iterator/AssetCollectionIterator.php:54
> Deprecation Notice: Return type of Assetic\Asset\Iterator\AssetCollectionIterator::next() should either be compatible with Iterator::next(): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/Skosmos/vendor/kriswallsmith/assetic/src/Assetic/Asset/Iterator/AssetCollectionIterator.php:89
> Deprecation Notice: Return type of Assetic\Asset\Iterator\AssetCollectionIterator::key() should either be compatible with Iterator::key(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/Skosmos/vendor/kriswallsmith/assetic/src/Assetic/Asset/Iterator/AssetCollectionIterator.php:84
> Deprecation Notice: Return type of Assetic\Asset\Iterator\AssetCollectionIterator::valid() should either be compatible with Iterator::valid(): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/Skosmos/vendor/kriswallsmith/assetic/src/Assetic/Asset/Iterator/AssetCollectionIterator.php:99
> Deprecation Notice: Return type of Assetic\Asset\Iterator\AssetCollectionIterator::rewind() should either be compatible with Iterator::rewind(): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/Skosmos/vendor/kriswallsmith/assetic/src/Assetic/Asset/Iterator/AssetCollectionIterator.php:94
> Deprecation Notice: strrpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in /var/www/html/Skosmos/vendor/kriswallsmith/assetic/src/Assetic/Asset/Iterator/AssetCollectionIterator.php:40


Maybe the [composer.json](https://github.com/NatLibFi/Skosmos/blob/master/composer.json#L67) should change to https://github.com/oomphinc/composer-installers-extender or https://github.com/oomphinc/composer-installers-extender

osma:
Excellent work! A couple of remarks on unit tests:

- I suggest merging the latest changes from `master` because the GitHub Actions CI tests were just fixed (PR #1397)
- PHP version matrix for unit tests should be adjusted in `.github/workflows/ci.yml` so that PHP 8.1 is no longer set as experimental (i.e. tests are no longer allowed to fail on 8.1)

osma:
I'll see if I can finish this off.

osma:
> But there is something wrong with library dependencies to AssetCollection for robloach/component-installer:

This is caused by the [components/handlebars.js](https://packagist.org/packages/components/handlebars.js) package we are using. We have the latest released version (4.7.7), but the packaging for Composer uses the outdated `robloach/component-installer` package which depends on the also outdated `kriswallsmith/assetic` package that causes deprecation warnings on PHP 8.1 and will likely break on 8.2. There's not much we can do here except to switch away from the Composer package components/handlebars.js into something better maintained - if nothing else works, we could just include the handlebars.js code in this repo. I'll open a separate issue about this shortly.

osma:
Reported the Handlebars.js dependency issue as #1401 .

PR #1391, opened 2022-11-24
Refactor sorting code for hierarchical and groups tabs

### LINKED ISSUE #1390: Sorting by notation does not work in groups tab if showNotation is false
At my local installation using a vocabulary with notated groups.
1. Set `skosmos:showNotation "false";
        skosmos:sortByNotation "true";` for the locally installed vocabulary in question (similarly to https://finto.fi/yso-aika/fi/ where these configuration settings are applied and working)
2. Observe the Groups tab

I expect to see the values ordered by notation codes. For YSO-aika, this is working in the Hierarchy tab. Looking at the code, it seems like sorting is done differently: https://github.com/NatLibFi/Skosmos/blob/0a51dd58bae105fa4c27ea0c1955a908489180e2/resource/js/groups.js#L55 https://github.com/NatLibFi/Skosmos/blob/0a51dd58bae105fa4c27ea0c1955a908489180e2/resource/js/hierarchy.js#L503-L532

I would assume that refactoring so that both codes use the same approach would yield identical behavior between the tabs in question.

kouralex:
Tackling issues described in #1390 
- Closes #1390 
This PR changes the code base via refactoring as some methods (namely: getLabel, pickLabel, nodeLabelSortKey, hierarchySort) and their respective code lines have been transferred from `hierarchy.js` to `scripts.js` so that they can be utilized in both `hierarchy.js`and `groups.js`.

Additional changes were made to how `$(document).on('click','div.group-hierarchy a', ...` (clicking groups) was handled as there were apparent issues that stood out whilst testing the now-refactored code. A careful research and comparison with `  $(document).on('click', '.concept-hierarchy a',` (clicking the hierarchy concepts) revealed that the code for clicking groups lacked some features the more-tested clicking hierarchy concepts had so I added them - now the jsTree hiearchy in groups tab seems to be working as expected.
I noticed that autocomplete shows notation codes even after applying `skosmos:showNotation "false"` setting and this is how it is working in https://finto.fi/yso-aika/fi/. It may be OK for YSO-aika vocabulary, but for the vocabulary I am currently testing it certainly is not. However, fixing the aforementioned is out of the scope of this PR.
## Checklist

### DIFF
--- a/resource/js/docready.js
+++ b/resource/js/docready.js
@@ -428,25 +428,29 @@ $(function() { // DOCUMENT READY
   $(document).on('click','div.group-hierarchy a',
       function(event) {
         $.ajaxQ.abortContentQueries();
+        var targetUrl = event.target.href || event.target.parentElement.href;
+        $('#hier-trigger').attr('href', targetUrl);
         var $content = $('.content').empty().append($delayedSpinner.hide());
         var loading = delaySpinner();
         // ajaxing the sidebar content
         $.ajax({
-            url : event.target.href,
+            url : targetUrl,
             req_kind: $.ajaxQ.requestKind.CONTENT,
+
             complete: function() { clearTimeout(loading); },
             success : function(data) {
+              $content.empty();
+              var response = $('.content', data).html();
+              if (window.history.pushState) { window.history.pushState({url: targetUrl}, '', targetUrl); }
               initHierarchyTooltip();
-              $('#hier-trigger').attr('href', event.target.href);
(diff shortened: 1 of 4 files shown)


joelit:
I asked for further clarifications in the comment section for the related issue #1390

kouralex:
@joelit I have given further clarifications in the comment section for the related issue https://github.com/NatLibFi/Skosmos/issues/1390, any news on this?

PR #1407, opened 2023-02-17
Update BARTOC Skosmos URL

nichtich:
BARTOC Skosmos is currently being moved from Basel to VZG with new domain https://skosmos.bartoc.org/.


osma:
Excellent, thanks for the PR!

PR #1409, opened 2023-02-20
Don't crash on incomplete vocabulary configuration

### LINKED ISSUE #1408: Skosmos crashes silently on missing void:uriSpace

The start page of a vocabulary missing `void:uriSpace` in its configuration.

## Description of the problem

Configure Skosmos with a vocabulary but omit/comment out its `void:uriSpace`.

## Expected behaviour

An error message should be shown instead. Pull request coming...

nichtich:
If a vocabulary configuration has no `void:uriSpace`, an error message should be shown instead of crashing. This closes #1408.

### DIFF
--- a/model/Vocabulary.php
+++ b/model/Vocabulary.php
@@ -71,7 +71,10 @@ class Vocabulary extends DataObject implements Modifiable
     {
         if ($this->urispace === null) // initialize cache
         {
-            $this->urispace = $this->resource->getLiteral('void:uriSpace')->getValue();
+            $urispace = $this->resource->getLiteral('void:uriSpace');
+            if ($urispace) {
+                $this->urispace = $urispace->getValue();
+            }
         }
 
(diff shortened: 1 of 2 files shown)


osma:
Thanks for the PR!

I wonder what the implications are for omitting uriSpace. I mean, this PR obviously fixes the immediate crash, but the setting is used in several places in both PHP and JS code, so just returning a `null` value may have some consequences. Have you tried this in practice?

nichtich:
Without `uriSpace`, basic vocabulary information is shown but not concepts, so it's better than a white screen. One may further think about how do handle incompletely configured vocabularies but for our instance this fix is enough.

osma:
Thank you for the PR @nichtich ! This will go into the `master` branch and eventually the 2.17 release. But I will also cherry-pick the commits for the v2.16-maintenance branch.

PR #1412, opened 2023-02-23
Support inclusion of configuration

### LINKED ISSUE #1403: Feature request: get configuration via SPARQL endpoint

Support reading content of `config.ttl` from a SPARQL endpoint. This would require to add two new properties in `config.ttl`, such as `skosmos:configEndpoint` and `skosmos:configGraph` to `config.ttl` so additional triples are taken from referenced SPARQL endpoint

Configuration is loaded in `model/GlobalConfig.php` and cached for performance reasons. I think the end of `initializeConfig` may be a good place to add loading of additional configuration via SPARQL. To avoid doing a SPARQL request each time, the result should be cached as well. I'm not sure on cache invalidation. Maybe only reload on request with a special query parameter e.g. `?config=reload`.

User who prefer to manage their configuration in Fuseki database.

In particular this would allow to manage vocabularies in one place (Fuseki). 

Adition of vocabularies requires to modify both `config.ttl` and import into Fuseki. This feature woul allow to manage it in one place and make sure the configuration is valid RDF (no invalid Turtle syntax).

nichtich:

This adds repeatable configuration property `skosmos:includeConfig` to include configuration from file or URL. The configuration is cached as usual, based on modification time of `config.ttl` only.

- Closes #1403

Introduce non-recursive inclusion of configuration.

The code is not *fully* covered by tests, in particular caching of configuration (was not covered before neither) and error messages (would require to add more example configuration files but I did not manage to get this into `GlobalConfigTest.php`).

## Checklist


osma:
Thanks a lot, this looks promising! I like the way this sidesteps the thorny cache invalidation problem by relying only on the timestamp of `config.ttl`.

model/GlobalConfig.php
```diff
@@ -88,31 +90,89 @@ private function initializeConfig()
...
      * @throws \EasyRdf\Exception
      */
-    private function parseConfig($filename)
-    {
-        $this->graph = new EasyRdf\Graph();
+    private function parseConfig($location, $depth=0) {
+        if (str_starts_with($location, "http://") || str_starts_with($location, "https://")) {
+            $ch = curl_init($location);
+            curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+            curl_setopt($ch, CURLOPT_HTTPHEADER, array('Content-type: text/turtle'));
```
osma, line 120: Shouldn't this be `Accept: text/turtle` ? Content-type is a response header, not a request header.

osma (changes requested):
I started looking at this but unfortunately didn't finish testing everything. However, I found two issues:

1. The `Content-type` header (see comment)

2. The test are passing, but there is now some extra output `Error: $prefix should only contain alpha-numeric characters`, as can be seen [here](https://github.com/NatLibFi/Skosmos/actions/runs/4356356955/jobs/7614252839?pr=1412) in the CI output under "Run PHPUnit tests". I think this happens because EasyRdf [doesn't like namespace prefixes with hyphens](https://github.com/easyrdf/easyrdf/issues/265) (only letters, numbers and underscores are allowed). I did a little digging and found out that the problematic prefix is `multiple-schemes:` but I have no idea where that comes from. Could you take a look?

nichtich:
> 2. The test are passing, but there is now some extra output `Error: $prefix should only contain alpha-numeric characters`, as can be seen [here](https://github.com/NatLibFi/Skosmos/actions/runs/4356356955/jobs/7614252839?pr=1412) in the CI output under "Run PHPUnit tests". I think this happens because EasyRdf [doesn't like namespace prefixes with hyphens](https://github.com/easyrdf/easyrdf/issues/265) (only letters, numbers and underscores are allowed). I did a little digging and found out that the problematic prefix is `multiple-schemes:` but I have no idea where that comes from. Could you take a look?

I doubt this is related to this isse because the `multiple-schemes:` prefix was already in the code before and only in *other* tests (see `grep -r multiple-schemes tests`).

PR #1423, opened 2023-03-22
fix conversion of DateTime to string in getContainsHtml

osma:

There is a recently surfaced (probably PHP 8.x related) problem in the YSO-aika (YSO Time) vocabulary. Concept pages failed to display and instead we got this message in the error log:

    [Wed Mar 22 09:53:30.124150 2023] [php:error] [pid 709825] [client 127.0.0.1:38096] PHP Fatal error:  Uncaught TypeError: preg_match(): Argument #2 ($subject) must be of type string, DateTime given in /var/www/html/Skosmos/model/ConceptPropertyValueLiteral.php:60

The reason is pretty simple: if a concept has a property with a datetime value, it cannot be implicitly casted to a string value. This PR fixes the problem by relying on the getLabel() method in the same class, which already handles DateTime to string conversion.

(not reported as an issue, I decided to fix this directly)

* fix DateTime to string conversion in ConceptPropertyValueLiteral.getContainsHtml() by relying on .getLabel()
* add unit test for verifying above conversion

none

## Checklist

### DIFF
--- a/model/ConceptPropertyValueLiteral.php
+++ b/model/ConceptPropertyValueLiteral.php
@@ -57,7 +57,7 @@ class ConceptPropertyValueLiteral extends VocabularyDataObject
     }
 
     public function getContainsHtml() {
-        return preg_match("/\/[a-z]*>/i", $this->literal->getValue()) != 0;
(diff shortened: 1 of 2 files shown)


miguelvaara:
Great job @osma :+1:

Issue #1436, opened 2023-04-25
Create new Twig templates based on HTML on skosmos-3-html branch


joelit:
- [ ] Ttwig template architecture
- [x] JS variables
- [ ] css
- [ ] images
- [ ] libraries
- [ ] fonts

joelit:
Superseded by #1451. Closed as duplicate.

PR #1438, opened 2023-04-26
First JS variables under SKOSMOS namespace.

joelit:
 - This is part of the twig overhaul
- Part of #1436 

- Tidying up twig templates that produce JS variables
- Putting the Skosmos JS variables under SKOSMOS namespace
- This PR does not have CSP nonce implemented
- This PR omits the variables generated with translation components
- This PR hasn't been tested

## Checklist

### DIFF
--- a/src/view/scripts.inc
+++ b/src/view/scripts.inc
@@ -1,0 +1,48 @@
+<!-- Namespace for Skosmos variables -->
+<script>
+  let SKOSMOS = (function () {
+    var content_lang = "{{ request.contentLang }}";
+    var explicitLangCodes = {{ explicit_langcodes ? 'true' : 'false' }};
+    var lang = "{{ request.lang }}";
+    var vocab = "{{ request.vocabid }}";
+    var waypoint_results = {{ parameters ? parameters.searchLimit : NULL }};
+
+    {% if request.page == 'page' and search_results and search_results|length == 1 %}
+      var prefLabels = [{"lang": "{{ search_results|first.label.lang }}","label": "{{ search_results|first.label }}"}{% for lang in search_results|first.foreignLabels %}{% for literal in lang %}{% if literal.type == 'skos:prefLabel' %},{"lang": "{{literal.lang}}", "label": "{{literal.label}}"}{% endif %}{% endfor %}{% endfor %}];
+      var uri = "{{ search_results|first.uri }}";
+    {% endif %}
+
+    {% if request.vocab  %}
+      var languageOrder = [{% for lang in request.vocab.config.languageOrder(request.contentLang) %}"{{ lang }}"{% if not loop.last %},{% endif %}{% endfor %}];
+      var showNotation = {{ not request.vocab.config.showNotation ? false : true }}
+      var sortByNotation = {% if request.vocab.config.sortByNotation %}"{{ request.vocab.config.sortByNotation }}"{% else %}null{% endif %};
+      var uriSpace = {{ request.vocab.uriSpace ? "{{ request.vocab.uriSpace }}" : null }}
+      var vocShortName = "{{ request.vocab.config.shortname }}";
+    {% endif %}
+
+    {% if plugin_params %}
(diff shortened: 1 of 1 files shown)


src/view/scripts.inc
```diff
@@ -0,0 +1,48 @@
...
+<script>
+  let SKOSMOS = (function () {
+    var content_lang = "{{ request.contentLang }}";
+    var explicitLangCodes = {{ explicit_langcodes ? 'true' : 'false' }};
+    var lang = "{{ request.lang }}";
+    var vocab = "{{ request.vocabid }}";
+    var waypoint_results = {{ parameters ? parameters.searchLimit : NULL }};
+
+    {% if request.page == 'page' and search_results and search_results|length == 1 %}
+      var prefLabels = [{"lang": "{{ search_results|first.label.lang }}","label": "{{ search_results|first.label }}"}{% for lang in search_results|first.foreignLabels %}{% for literal in lang %}{% if literal.type == 'skos:prefLabel' %},{"lang": "{{literal.lang}}", "label": "{{literal.label}}"}{% endif %}{% endfor %}{% endfor %}];
```
osma, line 11: There's quite a bit of spaghetti code in this. It's not clear to me what the code is actually doing. I wonder if we could move the logic to the PHP side? (possibly out of scope for this PR though)

src/view/scripts.inc
```diff
@@ -0,0 +1,48 @@
+<!-- Namespace for Skosmos variables -->
```
osma, line 1: I would make this a little bit more descriptive, e.g.

src/view/scripts.inc
```diff
@@ -0,0 +1,48 @@
...
+<script>
+  let SKOSMOS = (function () {
+    var content_lang = "{{ request.contentLang }}";
+    var explicitLangCodes = {{ explicit_langcodes ? 'true' : 'false' }};
+    var lang = "{{ request.lang }}";
+    var vocab = "{{ request.vocabid }}";
+    var waypoint_results = {{ parameters ? parameters.searchLimit : NULL }};
+
+    {% if request.page == 'page' and search_results and search_results|length == 1 %}
```
osma, line 10: Style issue: do we really want to indent the variable declarations that are inside conditional `if` blocks? It means that the rendered HTML/JS will have inconsistent indentation.

src/view/scripts.inc
```diff
@@ -0,0 +1,48 @@
...
+      var prefLabels = [{"lang": "{{ search_results|first.label.lang }}","label": "{{ search_results|first.label }}"}{% for lang in search_results|first.foreignLabels %}{% for literal in lang %}{% if literal.type == 'skos:prefLabel' %},{"lang": "{{literal.lang}}", "label": "{{literal.label}}"}{% endif %}{% endfor %}{% endfor %}];
+      var uri = "{{ search_results|first.uri }}";
+    {% endif %}
+
+    {% if request.vocab  %}
+      var languageOrder = [{% for lang in request.vocab.config.languageOrder(request.contentLang) %}"{{ lang }}"{% if not loop.last %},{% endif %}{% endfor %}];
+      var showNotation = {{ not request.vocab.config.showNotation ? false : true }}
```
osma, line 17: Is this double negation really necessary? I find it quite confusing

src/view/scripts.inc
```diff
@@ -0,0 +1,48 @@
...
+      var sortByNotation = {% if request.vocab.config.sortByNotation %}"{{ request.vocab.config.sortByNotation }}"{% else %}null{% endif %};
```
osma, line 18: This could be written in a cleaner way, look at the next line (19)

osma (changes requested):
I think the SKOSMOS namespace declaration looks good. But I think that before this can be merged, it should at least be integrated with some of the other templates - there should be an `include` directive that inserts this into the rendered output.

I gave a few suggestions for improvements in the comments.

src/view/scripts.inc
osma: Leaving for later.

PR #1441, opened 2023-04-26
Standardjs GitHub action

miguelvaara:
To automate JavaScript code style checking

- Closes #

Modified the ci.yaml file as necessary and uploaded two test files, to ensure success (files can be deleted later)

- 

## Instructions for the reviewer
1. Disable JavaScript checking in the pre-commit hook.
2. Edit one file in the resource/js directory so that it contains an intentional style error.
3. Git add, commit, push ..
4. Check GitHub Actions to see if the check process succeeds.
5. Repeat steps 2-4, but instead use the correctly styled JS file.
6. Restore the pre-commit to the original state.

## Checklist
- [ x] I have added tests that show that the new code works, or tests are not relevant for this PR (e.g. only HTML/CSS changes)
- [ x] The PR doesn't reduce accessibility of the front-end code (e.g. tab focus, scaling to different resolutions, use of `.sr-only` class, color contrast)
- [ x] The PR doesn't introduce unintended code changes (e.g. empty lines or useless reindentation)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,12 +50,10 @@ jobs:
         php_version: ${{ matrix.php_version }}
         php_extensions: gettext intl xsl pcov
 
-    - name: Install Node
     - uses: actions/setup-node@v3
       with:
         node-version: 16.x
 
-    - name: Install JavaScript dependencies
       run: npm ci
 
     - name: Check PHP code style
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,18 +50,19 @@ jobs:
         php_version: ${{ matrix.php_version }}
         php_extensions: gettext intl xsl pcov
 
-    - uses: actions/setup-node@v3
+    - name: Install Node
+      uses: actions/setup-node@v3
       with:
         node-version: 16.x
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -50,9 +50,20 @@ jobs:
         php_version: ${{ matrix.php_version }}
         php_extensions: gettext intl xsl pcov
 
+    - name: Install Node
+      uses: actions/setup-node@v3
+      with:
+        node-version: 16.x
+
+    - name: Install JavaScript dependencies
```
osma, line 58: I think this step could be avoided by making use of [npm package support](https://github.com/actions/setup-node/blob/main/README.md#caching-global-packages-data) in the setup-node action above. That also includes caching for npm dependencies.

Basically, just adding `cache: npm` to the setup-node action should ensure that npm dependencies are installed.

osma (approved):
LGTM. I made a suggestion in the comment, but we can also leave it for later and merge this as it is.

.github/workflows/ci.yml
miguelvaara: Good point. We agreed to make this improvement later.

PR #1442, opened 2023-04-27
Add Cypress e2e tests

### LINKED ISSUE #1189: Add e2e tests

osma:

We could use end-to-end (e2e) testing to verify functionality on the UI level. This is not possible, or not practical, with the current PHPUnit tests.

- Closes #1189

- add Cypress as a npm dev dependency
- basic scaffolding for Cypress tests
- a simple test that checks the transition from the front (landing) page to a vocabulary home page

This is just a very basic Cypress configuration, with a single example test, otherwise relying on the defaults provided by the Cypress first run wizard. More substantial tests would be needed later.

This only works locally for now, assuming that Skosmos is running at `http://localhost/Skosmos` . Setting up Cypress to work under GitHub Actions CI is a separate task.

If we're going to use Cypress, we also need guidance and developer documentation on when and how to use it.

## Checklist


cypress/support/e2e.js
```diff
@@ -0,0 +1,20 @@
...
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import './commands'
+
+// Alternatively you can use CommonJS syntax:
+// require('./commands')
```
kinow, line 20: Missing newlines here and in the `command.js`. But not important.

cypress/e2e/spec.cy.js
```diff
@@ -0,0 +1,12 @@
+describe('template spec', () => {
+  it('passes', () => {
+    // go to the Skosmos front page
+    cy.visit('http://localhost/Skosmos')
+    // click on the first vocabulary in the list
+    cy.get('#vocabulary-list').find('a').first().click()
+    // check that we are still within the Skosmos URL space
+    cy.url().should('include', 'http://localhost/Skosmos/')
```
kinow, line 8: This could fail is others have already port `80` used and somehow tell Cypress to use another port number. But easy to fix later if needed :+1:

cypress/e2e/spec.cy.js
```diff
@@ -0,0 +1,12 @@
...
+    // check that we are on the vocab home page
+    cy.get('#vocab-info')
```
kinow, line 10: This should probably still do something when it gets the element. I can't recall if it will return an error if the element is not found, but I suspect it might instead just return `null` or `undefined`, without raising an error.

kinow (commented):
Great idea to start small. It's very easy to have nondeterministic, brittle tests in Cypress when adding it to a project (usually fixed by changing how a test is written, to wait on some element like a spinning wheel, or for an HTTP call to be made and returned).

Left a few comments, but it's already looking really great! Would be interesting to find some issue recently reported where the UI was broken (IIRC, we had popup tooltips not displayed, maybe the scroll in hierarchy, or the layout broken in mobile/responsive mode) and write a Cypress test for that.

:clap:

cypress/e2e/spec.cy.js
osma: Yes, this needs to be more general, also in order to run the same tests under GitHub Actions.

cypress/support/e2e.js
osma: Right, I saw that too - those were autogenerated by Cypress, I didn't touch them myself. Anyway easy to fix.

osma:
Thanks for your comments @kinow!

This is targeting the `skosmos-3` branch where we are currently setting up the scaffolding for Skosmos 3 development. The idea is to build e2e tests alongside the new functionality. But there isn't much of that yet in working condition, so not much to test yet.

cypress/e2e/spec.cy.js
osma: Left for later.

miguelvaara (approved):
Looks good! :+1:

PR #1444, opened 2023-04-27
Add vue components for vocab statistics

UnniKohonen:

To add vue components for vocabulary statistics

- Closes #

Added global vue script and vocab stats as two vue components in separate files

## Checklist

### DIFF
--- a/resource/js/term-counts.js
+++ b/resource/js/term-counts.js
@@ -1,0 +1,45 @@
+const termCountsApp = Vue.createApp({
+  data () {
+    return {
+      languages: []
+    }
+  },
+  mounted () {
+    fetch('https://api.finto.fi/rest/v1/yso/labelStatistics?lang=fi') /* What should this be? */
+      .then(data => {
+        return data.json()
+      })
+      .then(data => {
+        console.log('term count', data)
+        this.languages = data.languages
+      })
+  },
+  template: `
+    <h3 class="fw-bold py-3">Term counts by language</h3>
+    <table class="table" id="statistics">
+      <tbody>
+        <tr>
+          <th class="main-table-label fw-bold">Concept language</th>
+          <th class="main-table-label fw-bold">Preferred terms</th>
+          <th class="main-table-label fw-bold">Alternate terms</th>
+          <th class="main-table-label fw-bold">Hidden terms</th>
+        </tr>
+        <term-counts :languages="languages"></term-counts>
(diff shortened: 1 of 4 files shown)


resource/js/term-counts.js
```diff
@@ -0,0 +1,45 @@
+const termCountsApp = Vue.createApp({
+  data () {
+    return {
+      languages: []
+    }
+  },
+  mounted () {
+    fetch('https://api.finto.fi/rest/v1/yso/labelStatistics?lang=fi') /* What should this be? */
```
osma, line 8: This should be a relative URL. In Skosmos 2.x, the variable `rest_base_url` was defined in a script [config.js](https://github.com/NatLibFi/Skosmos/blob/31f430b65be209d55209d53729c6beb7c92df984/resource/js/config.js) and then all AJAX style calls used that as the base URL.

For now I think we could just hardwire this to `rest/v1/...` i.e. strip the `https://api.finto.fi/` URL prefix and it should work.

resource/js/vocab-counts.js
```diff
@@ -0,0 +1,54 @@
+const resourceCountsApp = Vue.createApp({
+  data () {
+    return {
+      concepts: {},
+      subTypes: {},
+      conceptGroups: {}
+    }
+  },
+  mounted () {
+    fetch('https://api.finto.fi/rest/v1/yso/vocabularyStatistics?lang=fi') /* What should this be? */
```
osma, line 10: Same as above, strip the `https://api.finto.fi/` URL prefix until we have a mechanism to define the REST base URL.

osma (changes requested):
Looks good, I gave a few suggestions for changes.

osma:
Also, StandardJS style checks under GitHub Actions are now failing with this error:

> Run cd resource/js; npx standard *.js
standard: Use JavaScript Standard Style (https://standardjs.com/)
  /home/runner/work/Skosmos/Skosmos/resource/js/term-counts.js:1:23: 'Vue' is not defined. (no-undef)
  /home/runner/work/Skosmos/Skosmos/resource/js/vocab-counts.js:1:27: 'Vue' is not defined. (no-undef)

This is a bit unfortunate and needs to be fixed but it's probably not a fault in the code in this PR, we just need to somehow tell ESLint/StandardJS that Vue is available.

osma (approved):
LGTM!

PR #1445, opened 2023-04-27
Upgrade some GitHub actions

osma:

We are using old versions of some GitHub actions and there are deprecation messages on the GitHub workflow summary page.

n/a

Upgrade the following actions:

- actions/cache@v3 (was v2)
- paambaati/codeclimate-action@v4.0.0 (was v2.7.5)
- codecov/codecov-action@v3 (was v1)
- php-actions/composer@v6 (was v5)

This gets rid of two kinds of deprecation warnings:

* [Node 12 deprecated](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/)
* [save-state and set-output deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

I wanted to also use setup-node action to install npm dependencies (instead of a separate step as currently done), but it doesn't work because we don't have `package.lock` committed to version control. Whether we want that is a separate discussion.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
           ${{ runner.os }}-php${{ matrix.php_version}}-skosmos3-
 
     - name: Install Composer dependencies
-      uses: php-actions/composer@v5
+      uses: php-actions/composer@v6
       with:
         php_version: ${{ matrix.php_version }}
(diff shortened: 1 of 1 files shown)


osma:
All seems well so I will just merge this.

PR #1455, opened 2023-05-23
UI language negotiation fix

### LINKED ISSUE #1453: Skosmos UI loads in Arabic
Some users are reporting that the Skosmos UI loads in Arabic by default, without having chosen that language.

Will try to investigate and gather more information.

namedgraph:
With these fixes applied we are not experiencing #1453 anymore.

Code fix and configuration fix

- Closes #1453 

* Pass language value to `in_array()` as string instead of object
* Replaced `_` with `-` in `skosmos:languages` language tags found in the `dockerfiles` configuration files

## Checklist


osma:
Thanks a lot @namedgraph , this looks like a very reasonable set of fixes.

The code that this PR is targeting is really begging for unit tests. I think this would be the right moment to add those tests, verifying that the fix actually worked. I will see if I can add those easily before merging. This needs to go both to the `master` branch and to the `skosmos-3` branch where we have started to develop version 3 of Skosmos.

osma:
I implemented unit tests for the guessLanguage method, then applied the commits from this PR on top, and put it into PR #1457 which supersedes this one.

osma:
Merged via #1457. Thanks again!

namedgraph:
You're welcome :)

PR #1463, opened 2023-05-25
Cyperess axe tests

miguelvaara:
To distinguish between web and cli accessibility tests

- Closes #

Everything is new compared to the past. The intention is to run accessibility-related Axe tests under Cypress, while also offering developers the option to run the tests either in the browser or on the command line (command line implementation supports later GH Actions implementations).

The development of Skosmos and defining the usability tests are moving at different paces, so not all tests cases can be completed yet.

## How to
This is how you run cypress-axe tests:

In the root folder of Skosmos:

1. `npm install`

GUI Tests:

2. `test-accessibility-all-gui`
(Once the test browser has been opened and under the accessibility-tests folder)
- Open Inspector -> console -> clear the previous listing
- Click: axe-landing.cy.js (result table in Inspector)
... you can continue to run all the tests one by one in the same way .
- Close the browser and the CY application or run the command `ctrl-C` on the command line

CLI Tests:

3. (run tests on all page templates at the same time) `npm run test-accessibility-all-cli`
You can see the results in the table that appears in the command line listing
OR
Run accessibility tests separately per page template:
4. `npm run test-accessibility-about-cli`
`npm run test-accessibility-vocab-cli`
`npm run test-accessibility-feedback-cli`
`npm run test-accessibility-landing-cli`

## Checklist


cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
osma, line 1: This placeholder file should not be added

cypress/screenshots/axe-cli-feedback.cy.js/This is used to avoid unnecessary redundancy of code -- Logs (CLI) (failed).png
osma, line 1: Please drop this image from the PR (it might be a good idea to adjust .gitignore to prevent similar files from being accidentally added to commits)

cypress/screenshots/axe-cli-landing.cy.js/This is used to avoid unnecessary redundancy of code -- Logs (CLI) (failed).png
osma, line 1: Same as above, please drop this image

osma (changes requested):
I took a quick look at this PR and tried running the accessibility tests. This looks promising, but I think there are things to improve. I wonder if it's really necessary to have so much differing code just to be able to support both CLI and GUI tests - ideally the same test code should work for both.

Runnin the Cypress tests now shows test failures because of accessibility problems in the current templates. This is to be expected, but I don't think we can merge a PR with failing tests because it would create chaos for development processes. So either the a11y issues should be fixed, or more likely, the failing tests should be marked as "skipped" or "expected to fail".

Minor housekeeping details: There were some extra, unnecessary files included in the PR. Many JS files also had unnecessary extra lines at the end.

cypress/fixtures/example.json
miguelvaara: > I took a quick look at this PR and tried running the accessibility tests. This looks promising, but I think there are things to improve. I wonder if it's really necessary to have so much differing code just to be able to support both CLI and GUI tests - ideally the same test code should work for both.
> 
> Runnin the Cypress tests now shows test failures because of accessibility problems in the current templates. This is to be expected, but I don't think we can merge a PR with failing tests because it would create chaos for development processes. So either the a11y issues should be fixed, or more likely, the failing tests should be marked as "skipped" or "expected to fail".
> 
> Minor housekeeping details: There were some extra, unnecessary files included in the PR. Many JS files also had unnecessary extra lines at the end.

Yes, all the "extra" has to be removed. I will do that before the next push. Also, JavaScript is not yet tested and before the merge, code checking also needs to be enabled as agreed.

Could we have a short design/review session together to go over what we actually want (reference: "ideally the same test code should work for both")?

package.json
```diff
@@ -1,10 +1,20 @@
 {
+  "scripts": {
+    "test-accessibility-all-gui": "npx cypress open",
```
osma, line 3: If you want to define these kinds of shortcuts (personally I'm not convinced they're all necessary), the could the order of the command parts be changed slightly? I would prefer e.g. `test-accessibility-gui-all` instead of the above. This way it's easy to delete `-all` from the end and change to e.g. `-about`. It takes more work (more keystrokes) if you have to edit the middle part of the command.

cypress/support/accessibility.js
```diff
@@ -0,0 +1,54 @@
+export const checkA11y = options => {
```
osma, line 1: I find this a bit confusing. What is this local `checkA11y` and how does it differ from `cy.checkA11y`? Perhaps a comment would clarify.

It seems that this is only used once, below on line 43. Perhaps it could be inlined instead of defining a single-use function?

Also, it is exported, but apparently not being used outside this file (in a meaningful way), so exporting seems unnecessary.

osma (changes requested):
The general direction this is going seems good, but I have some comments on the implementation details.

cypress/e2e/accessibility-tests/axe-feedback.cy.js
```diff
@@ -0,0 +1,10 @@
+import { accessibilityTestRunner } from '../../support/accessibility.js'
+
+describe('Check accessibility of the landing page', () => {
```
osma, line 3: This should be "feedback page", right?

cypress/e2e/feedback.cy.js
```diff
@@ -1,3 +1,3 @@
 describe('Feedback page', () => {
   it('no-op test', () => {})
-})
+})
```
osma, line 3: Line break missing

cypress/support/accessibility.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+}
+
+function getConfigurationForCLITests () {
+  return it('Logs (CLI)', () => {
+    cy.checkA11y(null, null, terminalLog)
+  })
+}
+function getConfigurationForGUITests () {
+  return it('Goes through all needed levels', () => {
```
osma, line 41: Could this be more descriptive? I don't quite understand what this means

cypress.config.js
```diff
@@ -1,10 +1,20 @@
-const { defineConfig } = require("cypress");
+const { defineConfig } = require("cypress")
 
 module.exports = defineConfig({
   e2e: {
-    baseUrl: 'http://localhost/Skosmos',
+    // You also can run like this: npx cypress run --config "baseUrl=http://localhost/skosmos"
+    baseUrl: 'http://localhost/skosmos',
```
osma, line 6: If you change this setting then running Cypress doesn't work for developers whose Skosmos installation is at /Skosmos instead of /skosmos. I hope we can eventually find a better way to run Cypress tests in a standardized setting, but meanwhile I would suggest not changing the baseUrl. Like the comment above says, it can always be overridden on the command line. Or you could rename your Skosmos installation to /Skosmos so it matches what others use.

cypress/e2e/landing.cy.js
```diff
@@ -1,3 +1,5 @@
+import { checkA11y } from '../support/accessibility.js'
```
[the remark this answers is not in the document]
osma: Still there, this line should be deleted I think

cypress.config.js
osma: I suggest not changing the baseUrl but you can decide

osma (approved):
LGTM, still a couple of comments you can address before merging, but I leave that up to you

osma (approved):
Code looks good now, the only concern I have is that the a11y tests are now failing (for example when running `npx cypress run`) and I don't think we want to merge failing tests. So the tests should be marked as skipped, and reenabled when the a11y issues are actually fixed.

Another option would be marking tests as "expected to fail", but apparently Cypress [doesn't support that](https://github.com/cypress-io/cypress/issues/7420) (issue marked as wontfix).

PR #1469, opened 2023-05-31
Issue1436 global search template

joelit:
## Reasons for creating this PR

Twig template for global search

## Link to relevant issue(s), if any

Fixes #1481 . This PR is based on #1470, and it should be reviewed first. 

## Description of the changes in this PR

* global-search.twig
* font awesome icons
* multi-purpose search-results.inc

## Known problems or uncertainties in this PR

* The header is lacking the search bar

## Checklist

- [x] phpUnit tests pass locally with my changes
- [x] I have added tests that show that the new code works, or tests are not relevant for this PR (e.g. only HTML/CSS changes)
- [x] The PR doesn't reduce accessibility of the front-end code (e.g. tab focus, scaling to different resolutions, use of `.sr-only` class, color contrast)
- [x] The PR doesn't introduce unintended code changes (e.g. empty lines or useless reindentation)

PR #1470, opened 2023-05-31
Issue1436 vocab search template

joelit:

Twig template for vocabulary search

Linked to #1436

This PR is built on the PR #1469 as it uses the same search-results.inc

* Layout for the vocabulary search page
* Uses the same search-results.inc as the global search

* The header is lacking the search bar
* The search result filter sidebar is still just a placeholder

## Checklist


joelit:
Search-results.inc produces more or less similar html compared to the static examples
* The whitespace control of foreign labes could be improved
* Gettext translations are needed to display all of the properties for a given search result

cypress/e2e/vocab-search.cy.js
osma, line 1: If PR #1501 is merged before this one, then this file needs to be moved under tests/cypress/template

src/view/global-search.twig
```diff
@@ -0,0 +1,14 @@
+{% extends "base-template.twig" %}
+{% block title %}: "Search results"{% endblock %}
```
osma, line 2: Needs to be translated

src/view/global-search.twig
```diff
@@ -0,0 +1,14 @@
...
+{% block content %}
+<div class="container">
+  <div class="mx-5 bg-light">
+    <div class="px-3 pb-3">
+      <div class="search-count {% if limit_type or limit_parent or limit_group or limit_scheme %}search-limited{% endif %}">
+        <p class="py-2"><span class="fw-bold">{{ search_count }} results for</span> '{{ term }}'</p>
```
osma, line 8: Needs to be translated

src/view/search-results-filter.inc
```diff
@@ -0,0 +1,7 @@
+{% block content %}
+  <div class="col-md-4 pe-3">
+    <div class="p-4" id="sidebar">
+      <h4 class="fw-bold mb-5">Rajaus</h4>
```
osma, line 4: Needs to be translated

src/view/search-results.inc
```diff
@@ -0,0 +1,128 @@
+{% block content %}
+  {%~ if search_results is not null and search_results|length == 0 ~%}
+  <p class="no-results">"The search provided no results."</p>
```
osma, line 3: Needs to be translated

src/view/search-results.inc
```diff
@@ -0,0 +1,128 @@
...
+      {%~ endfor ~%}
+      </span>
+    </li>
+  {%~ endif ~%}
+{% endfor -%}
+{%~ if concept.uri ~%}
+    <li class="list-group-item px-0 py-1">
+      <span class="URI-icon">URI</span>
+      <span class="search-result-uri">{{ concept.uri }}</span>
+      <i class="fa-solid fa-copy"></i> <!-- no copy to clipboard functionality yet -->
```
osma, line 121: Please open a new issue about the missing copy functionality

osma (changes requested):
Looks like a good start, I gave some quick comments.

Please make sure the PR description is up to date.

cypress/e2e/vocab-search.cy.js
osma: It was just merged, so indeed this needs to be moved.

src/view/base-template.twig
```diff
@@ -10,7 +10,10 @@
 <meta name="generator" content="Skosmos {{ request.version }}" />
 <link href="node_modules/bootstrap/dist/css/bootstrap.min.css" media="screen, print" rel="stylesheet" type="text/css">
 <link href="resource/css/fonts.css" media="screen, print" rel="stylesheet" type="text/css">
+<link href="resource/css/color-scheme-1.css" rel="stylesheet" media="screen, print" type="text/css">
```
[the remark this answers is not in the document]
joelit: For some reason I forgot to add the color scheme CSS file in this PR. I found it easier if the color variables are already referred. I added the missing CSS files (although only one is used ATM) and opened an issue #1505  about the configuration.

src/view/search-results.inc
joelit: Opened, #1504 .

osma:
There's something wrong with this PR, there are now over 100 commits. I think it should be rebased on current skosmos-3 branch.

src/view/global-search.twig
joelit: Translated in #1469

src/view/global-search.twig
joelit: Translated in #1469

osma:
I tested this locally. Some quick comments:

1. The `color-scheme-*.css` files are unnecessary in this PR and should be removed. (We will need them later, but let's keep that out of scope for this PR)
2. You should make sure that the Cypress test works now that PR #1509 was merged. Probably you will at least need to switch the test to use another test vocabulary since we don't currently have a `yso` vocabulary for the phpunit/cypress tests. (Alternatively, you could wait until we have a YSO subset vocabulary in place for the tests.)

# Layout issues

I found some layout issues in the templates:

When searching YSA for `kissa`, the first result looks like this:

image

Problems:

3. The header font looks smaller than the rest of the text. I didn't check what it should be but I would expect it to be at least as large.
4. This concept has no labels in other languages, yet the globe icon is displayed.
5. The mappings to Allärs and YSO are displayed, but shouldn't there be an icon next to them?

I did the same search for KOKO and got this:

image

Problems:

6. The types are not separated by commas.

I also searched YSO for `flunssa` and got this:

image


Problems:

7. I would expect that the matching label "flunssa" would be displayed in the same font as "nuhakuume" (EB Garamond).

joelit:
Good points @osma ! I should have tested the template with vocabularies other than YSO. I resolved all of the above, and made some minor tweaks.

osma (changes requested):
Getting better. A few brief comments:

1. There seems to be something wrong with the commit history. I see duplicated commits and I cannot merge this from the command line (although GitHub doesn't indicate a merge conflict). I suggest doing a rebase+squash, combining all current work into a single commit against the current skosmos-3 branch.
2. The title of the page doesn't reflect that a search has been performed. E.g. on Skosmos 2, the page title is constructed as `Finto: YSO 'kissa'` (service name, vocabulary and search term). A similar title should be used.
3. I gave you slightly wrong advice on the header font for matched terms (e.g. flunssa) in the previous review. The font of the matching label should be the one used for regular text, same size and styling.
4. The arrows for broader/narrower do not match the layout spec. They should be rotated. Also, the alignment of icons is currently problematic, I think that center-align could be better.
5. Long lines are wrapped in an ugly way. They should be truncated, with a possibility to click them to reveal the full line. (This could be turned into a separate issue.)
6. For terms in other languages, the prefLabel should be emphasized (bold style, different color).
7. The HTML contains many `span` elements and uses of `versal` class. These are remnants from Skosmos 2 which are unnecessary. IMHO we should try to use simpler HTML structure, as in the static example.
8. The copy to clipboard button should be white
9. The URI text/icon is too large

joelit:
The latest update of the PR contains:

- [x] Squashed the commit history
- [x] Fixed the page title to be similar to Skosmos 2.x (e.g. Finto: YSO 'kissa')
- [x] Fixed the prefLabel emphasis to link color
- [x] Fixed the altLabel emphasis to italics within the list items
- [x] Fixed the font of the altLabels in the tilte row for each search result
- [x] Fixed the broader and narrower arrows to reflect the examples of the layout secification documents instead of the static example page
- [x] Center aligned the FA icons
- [x] Made an issue #1528 about the long search result lines
- [x] Removed superfluous CSS classes that were there "just in case"
- [x] Tidied the intendation and spacing of the output HTML code
- [x] Fixed the copy to clipboard icon
- [x] Fixed the URI icon-text

osma (changes requested):
Again getting much better. But I do have more comments!

# PR title

Could you change the PR title to something a bit more informative?

# global-search.inc

Is this file actually needed in this PR? It seems like a preliminary version of the global search template (e.g. doesn't include translations).

# General page layout

There's a lot of empty space above the search result content:

image

Compare this with the Finto spec:

image

# Extra language tags & untranslated types

I searched YSO for "vaatteet" and got this:

image

For some reason, there are `(fi)` language tags (without a preceding space) on the third narrower concept and all later ones, but not the first and second. (This happens with Skosmos 2 as well). I don't think they should be there or at least it should be consistent. 

Also, the types isothes:ThesaurusArray and skos:Collection are not properly translated. (This is a regression compared to Skosmos 2.)

# Extra whitespace and padding

The result for "kisat" looks like this:

image

I've selected the altlabels to show that there is an extra space before "kilvat". The same happens with all other property values as well. It would be better to omit this space (removing unnecessary whitespace from the template or using twig whitespace control).

But I think there should be slightly more padding between the icon and the text (best added after removing the whitespace). Compare with the Finto spec:

image

# Possibly unnecessary functionality in the template

I've pointed out in the comments a couple of `if` clauses which I think are unlikely to be ever true (deprecated concepts & submembers). Unless there is a known use for these, I think these should be left out. (And if there is a use, a Cypress test to verify the functionality would be great!)

# Cypress tests

The current test is very minimal. I suggested in a comment some additional test cases that could be implemented. I realize that making the tests may require the "mini YSO" test vocabulary that is currently being developed (see #1516).

src/view/global-search.twig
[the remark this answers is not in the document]
joelit: That's right, they are separate PRs these days - removed it.

tests/cypress/template/vocab-search.cy.js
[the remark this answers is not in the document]
joelit: I added more texts for the search result properties. We should maye wait for test-yso before writing tests for the non-concept types.

src/view/search-results.inc
```diff
@@ -0,0 +1,123 @@
...
+  {%- elseif property.type == 'skos:related' -%}<i class="property-hover fa-solid fa-cloud"></i>
+  {%- elseif property.type == 'skosmos:memberOf' -%}<i class="property-hover fa-solid fa-layer-group"></i>
+  {%- elseif property.type == 'skos:note' -%}<i class="property-hover fa-solid fa-note-sticky"></i>
+  {%- else -%}<i class="property-hover fa-solid fa-globe"></i>
+  {%- endif -%}
+      </span>
+    {%~ set previous = false ~%}
+    {%- for propval in property.values -%} {# loop through ConceptPropertyValue objects #}
+      {%~ set outerLast = loop.last ~%}
+      {%- if propval.SubMembers ~%} {# if property is a group concept that has sub properties #}
```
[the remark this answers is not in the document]
joelit: Apparently not - I removed this and some other cases where there was legacy logic that is not probably needed.

osma (approved):
LGTM

PR #1495, opened 2023-08-24
Skosmos 3: Use Apache Jena Fuseki docker image

osma:
## Reasons for creating this PR

PR #1477 by @kinow switched to the official Apache Jena Fuseki image on the `master` branch. The same needs to be done on the `skosmos-3` branch. This PR makes the same changes for Skosmos 3.

## Link to relevant issue(s), if any

- Fixes #1415 for Skosmos 3 as well
- adapts PR #1477 for Skosmos 3 code

## Description of the changes in this PR

* cherry-picked all commits from PR #1477 i.e. changes to docker configuration and documentation

## Known problems or uncertainties in this PR

n/a

(I decided to remove the Ubuntu base image update from this PR as it wasn't really related to the fuseki image update)

## Checklist

- [x] phpUnit tests pass locally with my changes
- [x] I have added tests that show that the new code works, or tests are not relevant for this PR (e.g. only HTML/CSS changes)
- [x] The PR doesn't reduce accessibility of the front-end code (e.g. tab focus, scaling to different resolutions, use of `.sr-only` class, color contrast)
- [x] The PR doesn't introduce unintended code changes (e.g. empty lines or useless reindentation)

PR #1496, opened 2023-08-25
Use dockerized Fuseki for PHPUnit tests

### LINKED ISSUE #1488: Use dockerized Fuseki for PHPUnit tests
Currently we have set up PHPUnit tests so that Fuseki is installed by the [init_fuseki.sh](https://github.com/NatLibFi/Skosmos/blob/master/tests/init_fuseki.sh) script as a regular Java process. If necessary, the script downloads the Fuseki installation tarball from the Apache Jena download site, starts it up, and then loads the test vocabulary data from the `tests/test-vocab-data` directory using the s-put utility that comes with Fuseki.

For Skosmos 3, I would like to switch to a dockerized Fuseki instance instead of using a regular Java process. This would have the following benefits:

1. The same Fuseki container (together with dockerized Skosmos, both running via docker compose) can also be used for running Cypress tests
2. We can use the exact same container setup both under GitHub Actions CI and local developer machines
3. "Eat your own dogfood": it's good practice to make use of our own Dockerfiles now that we have them

# Prerequisites

We should first switch to the official Fuseki Dockerfile from Apache Jena; this is being worked on in PR #1477 by @kinow 

# How to implement

- init_fuseki.sh needs to be rewritten so it simply runs a `docker compose` command instead of the current code for downloading, installing and starting up Fuseki; also, test vocabularies should be loaded using e.g. `curl` since we don't have `s-put` available anymore
- some changes to the GitHub Actions CI configuration may be necessary
- CONTRIBUTING.md and the wiki documentation for [unit and integration tests](https://github.com/NatLibFi/Skosmos/wiki/Unit-and-integration-tests) should be modified accordingly (e.g. need to explain that Docker is necessary for running PHPUnit tests locally)

osma:

Modernizing the test suite. See #1488

- Closes #1488
- Includes commits from #1495 (those should be merged first)

See commit 817dd97476caeb144139b6e205d7c390d3ed1e65 for the changes that are specific to this PR (not just copied from #1495).

- rewritten init_fuseki.sh to use `docker compose` to start containers (instead of downloading, installing and starting up Fuseki from a tarball distribution) and regular `curl` commands for loading the vocabularies (instead of s-put that comes with the Fuseki tarball).
- changed the PHPUnit tests to use the dockerized Fuseki at `http://localhost:9030/skosmos/sparql` instead of `http://localhost:13030/skosmos-test/sparql` that was used in the previous configuration.
- Fuseki configuration (for the container) changed to enable `unionDefaultGraph`, because it is currently required by Skosmos for some functionality (see #678) and some tests will fail if it's not enabled.
- adjusted GitHub Actions CI configuration a little (no need to use `source` for running `init_fuseki.sh` anymore)

1. After starting up containers, the new `init_fuseki.sh` script sleeps for 5 seconds before trying to load vocabularies. I had to add this as a stopgap because Fuseki wasn't always immediately ready to handle the queries. But I think a more elegant mechanism would be needed here, for example waiting in a loop that checks if Fuseki is ready.
2. The `docker compose` command used here starts up three containers (skosmos, skosmos-cache, fuseki) but the PHPUnit tests only need the fuseki container. However, we will soon need the others as well for running Cypress tests (see e.g. #1479), so I think it's not too bad.
3. One of the tests is failing. This is `JenaTextSparql::testQueryConceptsAlphabeticalOrderBy` which checks for locale-aware collation in Jena SPARQL queries. It seems to me that collation is not currently working in the dockerized Fuseki - perhaps locale data is missing from the container?

I tested this also directly with the example query from the [Jena ARQ Collation](https://jena.apache.org/documentation/query/collation.html) documentation. Basically I put this in `collation.rq`:

```
PREFIX arq: <http://jena.apache.org/ARQ/function#>
SELECT ?label WHERE {
    VALUES ?label { "tsahurin kieli"@fi "tšekin kieli"@fi "tulun kieli"@fi "töyhtöhyyppä"@fi }
}
ORDER BY arq:collation("fi", ?label)
```

then ran it against the Fuseki Docker container using `rsparql` (from Jena) and the result was this:

```
$ rsparql --query collation.rq --service http://localhost:9030/skosmos/sparql
-----------------------
| label               |
=======================
| "töyhtöhyyppä"@fi   |
| "tsahurin kieli"@fi |
| "tšekin kieli"@fi   |
| "tulun kieli"@fi    |
-----------------------
```

This is not the correct, locale-aware order; `tsahurin kieli` should be first and `töyhtöhyyppä` last. So I think that the Fuseki container configuration should be modified to enable support for locale-aware collation. @kinow would you be able to help with this?

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
         key: fuseki-${{ hashFiles('tests/init_fuseki.sh') }}
 
     - name: Start up Fuseki
-      run: cd tests; sh ./init_fuseki.sh
+      run: cd tests; ./init_fuseki.sh
     
     - name: Cache Composer dependencies
       uses: actions/cache@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,12 +24,6 @@ jobs:
       with:
         fetch-depth: 0
 
-    - name: Cache Fuseki installation
-      uses: actions/cache@v3
-      with:
-        path: tests/apache-jena-fuseki-*
-        key: fuseki-${{ hashFiles('tests/init_fuseki.sh') }}
-
     - name: Start up Fuseki
       run: cd tests; ./init_fuseki.sh
     
--- a/dockerfiles/Dockerfile.ubuntu
+++ b/dockerfiles/Dockerfile.ubuntu
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:22.04
 
 LABEL maintainer="National Library of Finland"
 LABEL version="0.1"
@@ -7,19 +7,18 @@ LABEL description="A Docker image for Skosmos with Apache httpd."
 ARG DEBIAN_FRONTEND=noninteractive
 
 # git is necessary for some composer packages e.g. davidstutz/bootstrap-multiselect
-# gettext is necessary as php-gettext was available in 18.04, but not in 20.04
 RUN apt-get update && apt-get install -y \
     apache2 \
     curl \
     gettext \
     git \
-    libapache2-mod-php7.4 \
+    libapache2-mod-php8.1 \
     locales \
-    php7.4 \
-    php7.4-curl \
-    php7.4-xsl \
-    php7.4-intl \
-    php7.4-mbstring \
+    php8.1 \
+    php8.1-curl \
+    php8.1-xsl \
+    php8.1-intl \
+    php8.1-mbstring \
     php-apcu \
     php-zip \
     unzip \
@@ -53,7 +52,7 @@ ENV LC_ALL=en_US.UTF-8
 ENV LANG=en_US.UTF-8  
 
 # timezone
-RUN sed -i 's/;date.timezone =/date.timezone = "UTC"/g' /etc/php/7.4/apache2/php.ini
+RUN sed -i 's/;date.timezone =/date.timezone = "UTC"/g' /etc/php/8.1/apache2/php.ini
 
 COPY dockerfiles/config/000-default.conf /etc/apache2/sites-available/000-default.conf
 
--- a/dockerfiles/config/skosmos.ttl
+++ b/dockerfiles/config/skosmos.ttl
@@ -32,7 +32,7 @@
 
 :tdb_dataset_readwrite rdf:type tdb2:DatasetTDB2 ;
     tdb2:location               "/fuseki/databases/skosmos" ;
-    # tdb2:unionDefaultGraph    true ;
+    tdb2:unionDefaultGraph    true ;
 .
 
 <#indexLucene> rdf:type text:TextIndexLucene ;
--- a/tests/JenaTextSparqlTest.php
+++ b/tests/JenaTextSparqlTest.php
@@ -371,6 +371,7 @@ class JenaTextSparqlTest extends PHPUnit\Framework\TestCase
      */
     public function testQueryConceptsAlphabeticalOrderBy()
     {
+        $this->markTestSkipped("disabled because ARQ collation doesn't work in dockerized Fuseki (Jena issue #1998)");
         $vocab = $this->model->getVocabulary('collation');
         $graph = $vocab->getGraph();
         $sparql = new JenaTextSparql($this->endpoint, $graph, $this->model);
--- a/tests/ModelTest.php
+++ b/tests/ModelTest.php
@@ -117,7 +117,7 @@ class ModelTest extends PHPUnit\Framework\TestCase
     public function testGetVocabularyByInvalidGraphUri()
     {
         $this->expectException(ValueError::class);
-        $this->expectExceptionMessage("no vocabulary found for graph http://no/address and endpoint http://localhost:13030/skosmos-test/sparql");
+        $this->expectExceptionMessage("no vocabulary found for graph http://no/address and endpoint http://localhost:9030/skosmos/sparql");
         $vocab = $this->model->getVocabularyByGraph('http://no/address');
         $this->assertInstanceOf('Vocabulary', $vocab);
     }
--- a/tests/VocabularyTest.php
+++ b/tests/VocabularyTest.php
@@ -43,7 +43,7 @@ class VocabularyTest extends \PHPUnit\Framework\TestCase
     {
         $vocab = $this->model->getVocabulary('testdiff');
         $endpoint = $vocab->getEndpoint();
-        $this->assertEquals('http://localhost:13030/skosmos-test/sparql', $endpoint);
+        $this->assertEquals('http://localhost:9030/skosmos/sparql', $endpoint);
     }
 
     /**
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -5,8 +5,8 @@ error_reporting(E_ALL);
 // make sure that a SPARQL endpoint is set by an environment variable
 $endpoint = getenv('SKOSMOS_SPARQL_ENDPOINT');
 if (!$endpoint) {
-    // default to Fuseki running on localhost:13030 as provided by init_fuseki.sh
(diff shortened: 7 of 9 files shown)


kinow (commented):
>After starting up containers, the new init_fuseki.sh script sleeps for 5 seconds before trying to load vocabularies. I had to add this as a stopgap because Fuseki wasn't always immediately ready to handle the queries. But I think a more elegant mechanism would be needed here, for example waiting in a loop that checks if Fuseki is ready.

I will have time to test it in some days or in one or two weeks. But with a healthcheck we should be able to remove the sleep 5.

https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck

osma:
> I will have time to test it in some days or in one or two weeks. But with a healthcheck we should be able to remove the sleep 5.

Thanks @kinow . This is not a big problem, I can probably find a better solution myself and the `sleep 5` works for now.

The bigger issue from my perspective is the lack of support for locale-aware collation in the Fuseki container. I think the Jena Fuseki Dockerfile needs to be modified, perhaps installing more locale packages and/or generating locales. But it's based on Alpine Linux and I'm not very familiar with it. Do you have an idea how to approach this?

kinow:
For Alpine I normally start a container with alpine and then do an apk search $pkg, apk add $pkg. Then if it works I update the dockerfile.

dockerfiles/jena-fuseki2-docker/Dockerfile
kinow, line 1: Ah, now it clicked to me. We will have to modify this image to add the extra packages. Meaning it might be harder for us to upgrade it as Jena does so :disappointed_relieved: 

The best solution would really be if with every release of Jena, they also published the Docker image to Docker Hub…

osma: Yes, exactly.

It could also be argued that there is a bug, or at least a missing feature, in the Jena Fuseki Dockerfile. Jena itself [supports locale-aware collation](https://jena.apache.org/documentation/query/collation.html), but the container currently doesn't. Is this something that should be reported as a Jena issue? Or contributed to upstream Jena when we have figured out a fix?

I've looked a bit at Alpine locale support, as explained in [this post](https://grrr.tech/posts/2020/add-locales-to-alpine-linux-docker-image/). It seems to me that locale support in Alpine is a bit lacking, but it is possible to install a limited set of locales like in [this recent gist](https://gist.github.com/alextanhongpin/aa55c082a47b9a1b0060a12d85ae7923). I didn't yet have time to test this in practice with the Jena Dockerfile.

kinow: >It could also be argued that there is a bug, or at least a missing feature, in the Jena Fuseki Dockerfile. Jena itself [supports locale-aware collation](https://jena.apache.org/documentation/query/collation.html), but the container currently doesn't. Is this something that should be reported as a Jena issue? Or contributed to upstream Jena when we have figured out a fix?
>
>I've looked a bit at Alpine locale support, as explained in [this post](https://grrr.tech/posts/2020/add-locales-to-alpine-linux-docker-image/). It seems to me that locale support in Alpine is a bit lacking, but it is possible to install a limited set of locales like in [this recent gist](https://gist.github.com/alextanhongpin/aa55c082a47b9a1b0060a12d85ae7923). I didn't yet have time to test this in practice with the Jena Dockerfile.

Yes, and if Alpine limits the installation or configuration of these locales, we could switch to another slim image, like debian-slim.

osma: Thanks @kinow !

I'm currently playing around with this. I found a more up to date [recipe for installing Alpine locale support](https://www.nixcraft.com/t/how-to-change-date-command-output-language-locales-in-alpine-linux/4434?u=nixcraft), not specific to Docker, but seems applicable. It seems it would be nowadays possible to install locale support for Alpine without compiling anything.

Alas, I haven't been able to make ARQ collation work yet. I'm considering skipping the failing PHPUnit test for now, so that I can continue working on other aspects of the testing infrastructure and leave this for later (maybe even hoping it will get fixed on the Jena side :crossed_fingers:). Thanks for opening the [Jena issue](https://github.com/apache/jena/issues/1998)!

kinow: Thanks for replying there, and for the better explanation & examples. I'll try to have a look and see if I can help, but I am in Brazil right now so timezone & family events might stop me from helping much, I am afraid :bow:

osma: Thanks @kinow and no hurry, enjoy your time in Brazil!

I decided to skip the collation unit test for now, so that we can move on with this PR. I will open a separate Skosmos issue that tracks the collation problem on this side. I hope that it can be fixed directly in upstream Jena and then we can copy the solution to our Dockerfile as well.

osma:
I created a separate [issue](https://github.com/NatLibFi/Skosmos/issues/1499) about the failing collation so that we can move on with this PR.

miguelvaara (commented):
Looks good! 

When I ran `init_fuseki.sh`, I got the following error message:
_./init_fuseki.sh: line 5: docker: command not found
Waiting for Fuseki to get ready
...waiting...
...waiting..._

I installed Docker Ubuntu 22.04 packages, but the result was the same. Based on the tip you provided :+1: , I installed the official Docker packages, after which the tests passed. Although the tests had some skip parts, the intention was not to test the application itself, but to test its testing. And it works very well! :+1: 

However, here are some minor formatting-related suggestions:

The arrangement/style of code in the [Dockerfile.ubuntu](https://github.com/NatLibFi/Skosmos/pull/1496/files#diff-be158a1bef21d98a9dacfc4d3ff070aabfc7e3464a883120194f2c2edee6783d) file is inconsistent. It would be easier to read and interpret the file if it followed the same style from start to finish:

**There are two newlines between the following, although usually there are no:**
_RUN php composer.phar install --no-dev_
[newLine]
[newLine]
_\# Configure Skosmos_

**Sometimes related or similar commands are grouped together under one header, and sometimes they are not:**
_\# skosmos layer
COPY . /var/www/html
RUN php composer.phar install --no-dev_

vs.

_RUN a2enmod rewrite
RUN a2enmod expires_

**Sometimes a single and separate command is labeled, and sometimes it's not:**
_\# timezone
RUN sed -i 's/;date.timezone =/date.timezone = "UTC"/g' /etc/php/8.1/apache2/php.ini_

vs.

_COPY dockerfiles/config/000-default.conf /etc/apache2/sites-available/000-default.conf_

It certainly does not matter what formatting we follow, but it does matter that it is consistent and easy to read.

kinow:
>It certainly does not matter what formatting we follow, but it does matter that it is consistent and easy to read.

:+1: good feedback! Either on this PR or later we can do a cleanup of the files for docker and improve docs, formatting, and consistency -- when I updated these recently I probably didn't take all those aspects into account :bow:

osma:
Thanks for testing and feedback @miguelvaara !

I see now that due to the complex branching history, this PR shows commits that were already merged into the `skosmos-3` branch earlier, and also the diff view is similarly misleading. This PR actually does not change Dockerfile.ubuntu at all from [how it already looks](https://github.com/NatLibFi/Skosmos/blob/skosmos-3/dockerfiles/Dockerfile.ubuntu) on the skosmos-3 branch. The recent changes to Dockerfile.ubuntu were done in PRs #1498 and #1495, already merged.

Rebasing this PR on top of current skosmos-3 branch should fix the commits and diff view, but I don't want to do that, because I've already opened PR #1508 which includes the same commits, and rebasing would likely cause a conflict.

I tried changing the base branch to `master` and back to `skosmos-3`, and it helped a bit, as the number of commits in this PR went from 8 to 5, but still the diff view incorrectly shows changes to Dockerfile.ubuntu which are not actually changes w.r.t. to the current `skosmos-3` branch.

Would it be OK to merge this PR and deal with the formatting of Dockerfile.ubuntu (which is not related to this PR at all) separately?

PR #1506, opened 2023-08-31
Checka11y function parameters with nulls and cypress folder tuning

miguelvaara:
A table that presents accessibility errors more effectively is necessary, even though SonarCloud reports the error. Additionally, the 'screenshots' and 'videos' directories should be in the correct test directory.

- Closes #

--- a/cypress.config.js
+++ b/cypress.config.js
@@ -22,6 +22,8 @@ module.exports = defineConfig({
       'tests/cypress/accessibility/**/*.cy.js',
       'tests/cypress/template/**/*.cy.js',
       'tests/cypress/e2e/**/*.cy.js'
-    ]
+    ],
+    screenshotsFolder: 'tests/cypress/screenshots',
+    videosFolder: 'tests/cypress/videos'
   }
 })

AND

--- a/tests/cypress/support/accessibility.js
+++ b/tests/cypress/support/accessibility.js
@@ -39,7 +39,7 @@ function getConfigurationForCLITests () {
 }
 function getConfigurationForGUITests () {
   return it('Check for possible accessibility errors at all logging levels set below in accordance with WCAG AA requirements', () => {
-    checkA11y({
+    checkA11y(null, null, {
       includedImpacts: ['minor', 'moderate', 'serious', 'critical'],
       runOnly: {
         type: 'tag',

## Checklist

### DIFF
--- a/tests/cypress/support/accessibility.js
+++ b/tests/cypress/support/accessibility.js
@@ -39,12 +39,6 @@ function getConfigurationForCLITests () {
 }
 function getConfigurationForGUITests () {
   return it('Check for possible accessibility errors at all logging levels set below in accordance with WCAG AA requirements', () => {
-    checkA11y(null, null, {
-      includedImpacts: ['minor', 'moderate', 'serious', 'critical'],
-      runOnly: {
-        type: 'tag',
-        values: ['wcag2aa']
-      }
-    })
+    checkA11y(null)
   })
 }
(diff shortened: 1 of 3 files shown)


tests/cypress/support/accessibility.js
```diff
@@ -39,7 +39,7 @@ function getConfigurationForCLITests () {
 }
 function getConfigurationForGUITests () {
   return it('Check for possible accessibility errors at all logging levels set below in accordance with WCAG AA requirements', () => {
-    checkA11y({
+    checkA11y(null, null, {
```
osma, line 42: Passing `null, null, {...}` is wrong as the arrow function is declared (above in the same file) to take just one parameter called `options`. So this change should be reverted.

The reason that the table no longer displays has to do with the options object declared here which is somehow wrong, but when passed as third parameter it never gets used.

PR #1509, opened 2023-08-31
Use dockerized Skosmos for Cypress tests

### LINKED ISSUE #1508: Use dockerized Skosmos for Cypress tests

We need a standardized Skosmos setup for running the Cypress tests. Currently this is a bit ad-hoc and every developer has their own Skosmos installations, which may have different configurations and vocabularies.

In #1488 / PR #1496, we are switching to a dockerized Fuseki for running the PHPUnit tests. In practice, this PR already uses `docker compose` to start up not just Fuseki but also a Skosmos container, even though it is not yet. The next logical step is to use that container for running Cypress tests.

Parts of this are already implemented or being worked on:

- PR #1498 switched to a more recent Ubuntu base image for the Skosmos container
- PR #1479 uses a similar Cypress/Skosmos/Docker setup under GitHub Actions CI

In addition to completing that work, these steps need to be done:

- [x] rename the `init_fuseki.sh` to something that covers Skosmos as well, e.g. `init_containers.sh`
- [x] adjust the Skosmos container so it uses the same configuration as for running PHPUnit tests
- [x] modify Cypress configuration so it uses the Skosmos container
- [x] adjust current Cypress tests so that they work with the test vocabularies we currently have
- [x] adjust documentation for running the tests (PHPUnit and Cypress) to correspond with the new situation

In the near future we probably need a new, larger and more diverse test vocabulary than the current very small ones, for example a cut down version of YSO, in order to make Cypress tests that test for advanced functionality. But that is outside the scope of this issue.

Mainly Skosmos developers.

Running Cypress tests in a standardized environment. This also enables writing Cypress tests that are specific to a particular, known test vocabulary.

This is a blocker for making substantial Cypress tests.

osma:

We want to run Cypress tests in a standard environment. See #1508

- Closes #1508

- define a custom `tests/docker-compose.yml` that extends `dockerfiles/docker-compose.yml` and overrides the skosmos configuration, so that Skosmos in the container uses the same `testconfig.ttl` configuration as the PHPUnit tests
- rename init_fuseki.sh script to init_containers.sh because it initializes also a Skosmos container; and make it use the customized docker-compose.yml from above
- make a separate skosmos configuration file for PHPUnit tests that involve overriding baseHref, and drop the baseHref setting from the regular `testconfig.ttl` (otherwise links aren't working in the skosmos container)
- configure Cypress to use the dockerized skosmos on localhost:9090
- (unrelated) fix the Cypress test for the error page that was failing due to a change in template HTML structure

After these changes, I can successfully run both the PHPUnit and Cypress tests locally. Instructions for running the tests:

Set up the Docker containers:

    cd tests
    ./init_containers.sh

Run the tests normally:

    vendor/bin/phpunit
    npx cypress open # or run

After running the tests, the containers can be shut down like this:

    cd tests
    docker compose down

Wiki documentation for running PHPUnit and Cypress tests should be updated when/after merging this.

This PR doesn't include running Cypress tests under GitHub Actions CI; we have a draft PR #1479 about that, but it needs some more work that is best done after this PR is merged.

## Checklist


osma:
I merged #1496 into skosmos-3, then rebased this PR branch on top of current skosmos-3 to clean up the commit history and diff. Now this PR shows only actual changes w.r.t. current skosmos-3.

miguelvaara (commented):
When running the command `./init_containers.sh`, I encountered an error:

```
Error response from daemon: Conflict. The container name "/skosmos-fuseki-cache" is already in use by container "0fcc35ef7ecc3336b294cb2e577de388e77e24885a866cf2a88c1b1489b9314c". You have to remove (or rename) that container to be able to reuse that name.
Waiting for Fuseki to get ready
...waiting...
...waiting...
...waiting...
```

I removed the previous containers (deleting just the Varnish container would likely have sufficed), after which all runs went smoothly. I am wondering whether this situation could be considered problematic or not. I was considering how much effort it would be to implement some adjustment that automatically removes a conflicting container. On the other hand, manually removing the conflicting container and re-running the script is not a significant task. In my opinion, both approaches are acceptable; having automation or having the tester manually remove the conflicting container.

miguelvaara (commented):
The execution of PHPUnit tests was successful. Of course, the tests reported errors, but that's okay, and the passing of tests is not within the scope of this PR.

!! The errors in PHPUnit tests were my own mistake. I forgot to run php composer update - once again :-D

miguelvaara (commented):
The Cypress tests are working well, meaning that the modified settings are functioning and the http://localhost:9090 URL is responding correctly.

miguelvaara (approved):
Everything seems to be working well. In think we can proceed with the merge.

PR #1510, opened 2023-08-31
Add feedback template

### LINKED ISSUE #1486: Feedback Page twig-templates

This issue is for making sure the initial version of the feedback page fulfills common requirements. It's OK if not all functionality is implemented.

- [ ] Layout follows [the static example](https://github.com/NatLibFi/Skosmos/blob/skosmos-3-html/feedbackpage.html)
 - [ ] Main .twig -template exists
 - [ ] .inc -templates exist
 - [ ] [cypress tests](https://github.com/NatLibFi/Skosmos/blob/skosmos-3/cypress/e2e/feedback.cy.js) exist

UnniKohonen:

Twig template for feedback page

- Closes #1486 

- Adding feedback.twig
- Adding cypress tests
- Minor changes to skosmos.css

## Checklist


src/view/feedback.twig
```diff
@@ -0,0 +1,51 @@
...
+    <h2>{{ 'Feedback has been sent!' | trans }}</h2>
+    <p id="feedback-thanks">{{ 'Thank you for your feedback' | trans }}</p>
+  {% else %}
+
+    <h2 class="fs-3 mb-5">{{ 'Contact us!' | trans}}</h2>
+    <p class="fs-5 mb-5">{{ 'feedback_enter_name_email' | trans }}</p>
+    
+    <form id="feedback-form" method="post" action="{% if request.vocab %}{{request.vocab.id}}/{% endif %}{{ request.lang }}/feedback">
+      <div class="mb-4">
+        <label class="form-label" for="vocab-select">Vocabulary</label> {# translation should be added, currently doesn't exist #}
```
osma, line 16: The translation should be added, along with the `trans` filter in this template

osma (changes requested):
Looks very good and works well when I tested this locally. The only issues I could find are related to translations and the use of colons.

We currently don't have a translation for the string "Vocabulary". This should be added as a msgid to `resource/translations/skosmos_en.po` along with the translation. And the same should be done to the `skosmos_fi.po`, `skosmos_sv.po` and `skosmos_se.po` files that we maintain (the other languages can wait).

Regarding the use of colons: now the headings on the form contain colons e.g. `Name:` and `Email:`, but the design spec for Skosmos 3 doesn't have them, so the colons should be removed both from the Twig template and the po files listed above.

resource/translations/skosmos_en.po
```diff
@@ -135,9 +135,12 @@ msgstr "Content language"
 msgid "Download this concept in SKOS format:"
 msgstr "Download this concept:"
 
+msgid "Vocabulary"
+msgstr "Vocabulary"
+
 #: /tmp/cache/587c83c09abaa/4a/4a8733799c4f4aa865ece4715881ffcbef92985ed08edc2acd445e7b0c129fab.php:144
 msgid "E-mail:"
```
osma, line 142: Please remove the colon also from msgid's (and the twig template)

osma (changes requested):
You have removed the colons from msgstr's (the strings displayed to the user), but for consistency, I think they should be removed from msgid's as well, and of course the twig template. I pointed out one example, the others are similar.

osma (approved):
Excellent!

PR #1512, opened 2023-09-08
Top-level `docker-compose.yml`

### LINKED ISSUE #1511: `docker-compose` setup unsuitable for multiple Skosmos deployments
## Description of the problem
We want to deploy several Skosmos instances on the same machine, but the current `docker-compose.yml` has several issues that prevents this:
* unnecessary `container_name` properties on services, which makes container names conflict when the second instance is started
* `docker-compose.yml` file in a `dockerfiles` subfolder, which is used to construct container names (after explicit `container_name` are removed), e.g. `dockerfiles_skosmos_1`

My suggestion:
* remove `container_name` properties
* move `docker-compose.yml` to the top-level dir

Then we can launch Skosmos from 2 different folders, e.g. `Skosmos-A` and `Skosmos-B`, and the container names would become `skosmos-a-skosmos-1`, `skosmos-a-virtuoso-1` etc. and `skosmos-b-skosmos-1`, `skosmos-b-virtuoso-1`, respectively. That way they will not conflict.

namedgraph:

`docker-compose` setup unsuitable for multiple Skosmos deployments

- Closes #1511

* moved `docker-compose.yml` to the top-level dir
* fixed relative paths accordingly
* removed `container_name` properties
* replaced host port numbers with env variables with default values

Not tested directly :) But these changes work on our private fork and I was able to deploy 2 Skosmos instances on different ports on the same machine.

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,10 +2,9 @@ version: '3.7'
 
 services:
   fuseki:
-    container_name: skosmos-fuseki
     hostname: fuseki
     build:
-      context: ../dockerfiles/jena-fuseki2-docker
+      context: ./dockerfiles/jena-fuseki2-docker
       dockerfile: Dockerfile
       args:
         JENA_VERSION: 4.8.0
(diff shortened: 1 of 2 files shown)


osma (changes requested):
The changes here look good and the reasons behind the changes were already discussed in issue #1511.

The documentation under `dockerfiles/README.md` needs to be updated as well. Currently it says that the `docker compose` command should be run under `dockerfiles/` which obviously isn't true anymore after these changes. Also, the environment variables for setting the ports should be documented there.

When that is done (@namedgraph can you do it?) then this can be merged to `master`, i.e. Skosmos 2.x.

After that the changes should be also ported to the `skosmos-3` branch because we want to keep the Docker configuration similar for both versions. On the `skosmos-3` some additional adjustments to the phpunit and cypress test suite setup need to be done as well, probably just fixing up paths in `tests/init_containers.sh` that currently assume that the `docker-compose.yml` file is under `dockerfiles/`. I can do that once we have this in `master`.

namedgraph:
I'll try this week.

osma:
What's the status @namedgraph ? Can you fix `dockerfiles/README.md` so it syncs with the changes in this PR? Then we could merge this.

namedgraph:
@osma made changes. Are the good enough?

osma:
Looks good now so I'm merging this. Thanks a lot!

The work still needs to be ported to Skosmos 3.x but I'll take care of that later.

PR #1520, opened 2023-09-21
Added trans filter to templates

### LINKED ISSUE #1503: Add trans filter to twig templates

Adding trans filters of Symfony to existing twig templates in [skosmos-3](https://github.com/NatLibFi/Skosmos/tree/skosmos-3l) branch.

All Skosmos developers

Translating text in twig templates

Trans filters using syntax of older twig extensions were removed and need to be replaced by new syntax.

Vainonen:
New Symfony translation component added to skosmos-3 branch. Translation tags/filters notation has changed from previously used.

- Closes #1503

trans filters added to Twig templates and inc files. Translation files containing variables in their identifiers are translated with notation `{% trans %} ... %variable% ... {% endtrans %}`

I could not find what causes and how to test this error message in file concept-card.inc:
`{% trans %}Error: Term "{{ term }}" not found in vocabulary!{% endtrans %}`

## Checklist

### DIFF
--- a/src/view/concept-card.inc
+++ b/src/view/concept-card.inc
@@ -4,12 +4,12 @@
     <h2 class="visually-hidden">Concept information</h2>
     {% if concept.deprecated %}
     <div class="alert alert-danger">
-      <p class="deprecated-alert">deprecated</p>
+      <p class="deprecated-alert">{{ "deprecated" | trans }}</p>
     </div>
     {% endif %}
     {% if concept.label.lang != request.contentLang %}
     <div class="alert alert-lang">
-      <p class="language-alert">There is no term for this concept in this language</p>
+      <p class="language-alert">{{ "There is no term for this concept in this language" | trans }}</p>
     </div>
     {% endif %}
     <div class="container concept-info{% if concept.deprecated %} deprecated-concept{% endif %}">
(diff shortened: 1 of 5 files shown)


src/view/concept-card.inc
```diff
@@ -183,7 +183,7 @@
 
   {% else %}
   <div class="alert alert-danger">
-    <p>Error: Term "{{ term }}" not found in vocabulary!</p>
+    <p>{% trans %}Error: Term "{{ term }}" not found in vocabulary!{% endtrans %}</p>
```
osma, line 186: I think you should use `"%term%"` here instead of the brace syntax (looking at other similar strings that use variable interpolation)

osma (changes requested):
Looks OK, except that the concept page is currently broken. Viewing http://localhost/Skosmos/yso/en/page/p12345 I get this error in the logs:

```
PHP Fatal error:  Uncaught Twig\\Error\\SyntaxError: A message inside a trans tag must be a simple text. in /var/www/html/Skosmos/src/view/concept-card.inc:186
```

The solution is to use `%term%` instead of `{{ term }}` in the template (see comment)

Other than this, LGTM!

src/view/concept-card.inc
Vainonen: Ok. I also fixed this in "Description of the changes in this PR":
trans filters added to Twig templates and inc files. Translation files containing variables in their identifiers are translated with notation {% trans %} ... %variable% ... {% endtrans %}

PR #1524, opened 2023-09-27
Issue 1521 hierarchy tab vue component

### LINKED ISSUE #1521: Implement hierarchical view in sidebar using Vue

Adding a Vue component for vocabulary hierarchy in the sidebar.

Initial plan is to implement the component from the ground up (possibly something like [this](https://vuejs.org/examples/#tree)). If this doesn't work, it's also possible to use an external library like [PrimeVue](https://primevue.org/) and its [Tree](https://primevue.org/Tree) component.

- [ ] Vue component for hierarchy
- [ ] Cypress tests for hierarchy

UnniKohonen:

Adding a Vue component for hierarchy tab in sidebar

- #1521

Requirements from #1521 that this PR addresses:
- 1 (basic functionality)
- 2a and 2b (partial page load)
- 5 (correct URLs for concepts outside of the vocabulary)
- 6a (not loading top concepts unnecessarily)

## Checklist

### DIFF
--- a/resource/js/tab-hierarchy.js
+++ b/resource/js/tab-hierarchy.js
@@ -1,7 +1,17 @@
+/* global Vue */
+/* global SKOSMOS */
+/* global partialPageLoad */
+
 const tabHierApp = Vue.createApp({
   data () {
     return {
-      hierarchy: []
+      hierarchy: [],
+      loading: false
+    }
+  },
+  provide () {
+    return {
+      partialPageLoad
     }
   },
   mounted () {
@@ -11,8 +21,14 @@ const tabHierApp = Vue.createApp({
     }
   },
   methods: {
+    handleClickHierarchyEvent () {
+      // only load hierarchy if hierarchy tab is available
+      if (!document.querySelector('#hierarchy > a').classList.contains('disabled')) {
+        this.loadHierarchy()
+      }
+    },
     loadHierarchy () {
-      // if we are on the concept page, load hierarchy for the concept, otherwise load top concepts
+      // if we are on a concept page, load hierarchy for the concept, otherwise load top concepts
       if (SKOSMOS.uri) {
         this.loadConceptHierarchy()
       } else {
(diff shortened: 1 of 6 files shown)


joelit:
Great start, and the tree is pretty fast! :)

* The functional specs for the hierarcy tree should be listed somewhere (here in this PR for example) since they can get pretty detailed, and it's good to keep track of where this PR fills the specs, and where it doesn't
* There should be a sanity check in place whether the top concept level should be shown - or should the hierarcy be shown only when a concept has been selected
   * E.g. https://finto.fi/mesh/fi/ vs https://finto.fi/mesh/fi/page/?uri=http://www.yso.fi/onto/mesh/D020216
   * In Skosmos 2 https://github.com/search?q=repo%3ANatLibFi%2FSkosmos%20disabledHierarchy&type=code
   * Now the tree is loading https://api.finto.fi/rest/v1/mesh/topConcepts/?lang=fi :lla 110 top concepts, but fails to display them on the vocab page
* When the tree is opened with a concept-uri, e.g. http://localhost/Skosmos/yso/fi/page/p1318
   * The tree is opened correctly at all the parent nodes where the concept appears
   * But parts of the tree that are open cannot be closed anymore
* loadConceptHierarcy is pointing to api.finto.fi - it should be amended for a relative url

resource/js/tab-hierarchy.js
```diff
@@ -0,0 +1,223 @@
...
+    },
+    handleClickConceptEvent (event, concept) {
+      concept.isOpen = true
+      this.$emit('loadChildren', concept)
+      this.partialPageLoad(event, this.getHref(concept.uri))
+    },
+    loadChildrenRecursive (concept) {
+      this.$emit('loadChildren', concept)
+    },
+    getHref (uri) {
```
osma, line 184: I think this method will eventually be needed in other tabs as well, so it could be extracted into a general utility method. But for now it can live here until we need it elsewhere.

osma (changes requested):
This looks like a very good start. The promised functionality seems to work, so I think this could be merged and more functionality added later in subsequent PRs.

The JS style check failed with this error:

    resource/js/partial-page-load.js:53:7: 'partialPageLoad' is assigned a value but never used. (no-unused-vars)

This should be fixed before merging because currently all CI tests (except Cypress) are failing.

Also, SonarCloud gave a few complaints about code smells, for example suggesting the use of `for...of` for some loops. These could be fixed as well if it's not too hard to do, but it's not crucial for merging.

I suggested extracting the `getHref` method into a generally usable utility method, but this can be left for later when it is actually needed in several components.

PR #1529, opened 2023-09-27
Show deprecated concepts without replacement in change list

### LINKED ISSUE #1513: Tab 'New and Deprecated' lists of concepts
Hi,

It seems that there are two issues concerning
"Deprecated Concepts
Deprecated concepts are concepts that are no more valid but are kept in the vocabulary for backward compatibility. Skosmos can detect deprecated concepts with :

    an owl:deprecated true metadata on the skos:Concept;
    optionnally, the dcterms:replacedBy pointing to the (possibly multiple) concepts that replace this one;

Deprecated concepts will not be visible in the alphabetical listing, in the hierarchical view, and cannot be searched. They can still be accessed through their URI."

1. when setting skosmos:showDeprecated to 'false' the deprecated concepts are not searchable or displayed in the alphabetical list, BUT there are still listed in the Hierarchy tree;
2. concepts that have the owl:depracted true element/value only, BUT do not have the (optional) dct:replacedBy element seem to be missed completely/altogether and do not show up in the 'New and Deprecated' lists at all. 

Could you please provide a solution for esp. issue 2. Thanks.

Best wishes,
Jeannine

osma:

It was reported in #1513 that 

> concepts that have the owl:depracted true element/value only, BUT do not have the (optional) dct:replacedBy element seem to be missed completely/altogether and do not show up in the 'New and Deprecated' lists at all.

This PR fixes the problem by making sure that deprecated concepts which do not have an isReplacedBy relationship are shown. For example "elevated tracks" in this screenshot is a deprecated concept which doesn't have a replacement (I removed it manually from the YSO version I used for testing):

image

I verified that the performance of the "New and Deprecated" page is unchanged; possibly it is actually slightly better (1.4 seconds instead of 1.5 for the test case I used).

- Closes #1513 (NB the issue mentions another problem about deprecated concepts shown in the hierarchy, but we need more clarification on that and a separate issue would be good)

* reformat the SPARQL query snippet in GenericSparql.generateChangeListQuery to make it more readable and easily editable
* change the querying and handling of deprecated concepts for the change list so that also concepts without isReplacedBy are included in the list; the deprecation status must be tracked separately from the possible replacement
* adjust the `changes` test vocabulary adding a deprecated concept without isReplacedBy and add tests to check that it is included as well (I had to adjust other unit tests that broke due to the changes in the test vocabulary)

The same or similar changes need to be repeated for Skosmos 3 as well.

## Checklist

### DIFF
--- a/model/sparql/GenericSparql.php
+++ b/model/sparql/GenericSparql.php
@@ -2262,13 +2262,17 @@ EOQ;
         $deprecatedVars = '';
         if ($showDeprecated) {
             $deprecatedVars = '?replacedBy ?deprecated ?replacingLabel';
-            $deprecatedOptions =
-            'UNION {'.
-                '?concept dc:isReplacedBy ?replacedBy ; dc:modified ?date2 .'.
-                'BIND(COALESCE(?date2, ?date) AS ?date)'.
-                'OPTIONAL { ?replacedBy skos:prefLabel ?replacingLabel .'.
-                    'FILTER (langMatches(lang(?replacingLabel), \''.$lang.'\')) }}'.
-                'OPTIONAL { ?concept owl:deprecated ?deprecated . }';
+            $deprecatedOptions = <<<EOQ
+UNION {
+    ?concept dc:isReplacedBy ?replacedBy ; dc:modified ?date2 .
+    BIND(COALESCE(?date2, ?date) AS ?date)
+    OPTIONAL {
+        ?replacedBy skos:prefLabel ?replacingLabel .
+        FILTER (langMatches(lang(?replacingLabel), '$lang'))
+    }
+}
+OPTIONAL { ?concept owl:deprecated ?deprecated . }
+EOQ;
         }
 
         $query = <<<EOQ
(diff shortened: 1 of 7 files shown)


jeanninebeeken:
Hi Osma,

Thank you for fixing the issue. May I ask whether the fix will be part of/visble in version 2.17 or only from version 3.0 onwards, thanks.

Best wishes,
Jeannine

joelit (approved):
Looks good (even though switching to Skosmos 2 for testing was a bit twiddly)

osma:
@jeanninebeeken 

> Thank you for fixing the issue. May I ask whether the fix will be part of/visble in version 2.17 or only from version 3.0 onwards, thanks.

This will first be fixed on the development branch (`master`) and on the Skosmos 2.x maintenance branch `v2.16-maintenance` which you can start using immediately; if you followed the recommended installation instructions, upgrading will be a simple `git pull` command. Later we may release either a 2.16.x point release or 2.17 containing the fix - not sure about these, since we're focusing on Skosmos 3 development, so 2.x releases are not a priority.

The same fix has to be done for Skosmos 3 as well so we don't lose the improved functionality. For now this means only porting the back-end parts, because the change list is not yet implemented in the new Skosmos 3 front-end.

PR #1549, opened 2023-11-14
Concept page template and Cypress tests

### LINKED ISSUE #1484: Concept Page twig-templates

This issue is for making sure the initial version of the concept page fulfills common requirements. It's OK if not all functionality is implemented.

- [ ] Layout follows [the static example](https://github.com/NatLibFi/Skosmos/blob/skosmos-3-html/termpage.html)
 - [ ] Base .twig -template exists
 - [ ] .inc -templates exist
 - [ ] cypress tests exist

osma:

Reimplement the concept page template so it follows the new Skosmos 3 design.
Also add Cypress tests verifying the existing functionality.

This implements just the basic functionality. Many features and special cases are not implemented yet; see checklist in https://github.com/NatLibFi/Skosmos/issues/1484#issuecomment-1739297570

- Implements first parts of #1484

- refactor Vocabulary.getConceptInfo() so it returns a single concept (instead of an array of size 1)
- rewrite the concept page template based on the static example (but adapting it heavily to use Bootstrap Grid instead of a HTML table)
- add some Cypress tests for the concept page / template

n/a

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -553,7 +553,7 @@ body {
     margin-bottom: 0;
   }
 
-  .term-other-languages a {
+  .term-other-languages .label {
     display: inline-block;
     width: 50%;
   }
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -276,7 +276,7 @@ body {
     font-size: 16px;
   }
 
-  #main-container h1, #main-container h2 {
+  #main-container h1 {
     font-family: var(--font-family-heading);
     font-weight: bold;
   }
@@ -506,9 +506,10 @@ body {
     padding: .5rem 0;
   }
 
-  .property-label {
+  .property-label h2 {
     font-weight: bold;
     font-size: 20px !important;
+    margin-bottom: 0;
   }
 
   .property-value {
@@ -537,7 +538,7 @@ body {
     padding-top: 1.25rem;
   }
 
-  #concept-label h2 {
+  #concept-label h1 {
     float: left;
   }
 
@@ -560,6 +561,10 @@ body {
   #concept-other-languages .label {
     display: inline-block;
     width: 50%;
+  }
+
+  #download-links li {
+    display: inline-block;
   }
 
   /***** Main content searchpage & searchpage multi vocab *****/
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -558,6 +558,10 @@ body {
     margin-bottom: 0;
   }
 
+  .prop-skos_altLabel .property-value li, .altlabel {
+    font-style: italic;
+  }
+
   .prop-foreignlabels h3 {
     font-size: 1rem;
     margin-bottom: 0;
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -534,7 +534,8 @@ body {
   /*** Main content termpage ***/
   #term-heading span {
     width: 30%;
-    align-self: end;
+    align-self: start;
+    padding-top: 1.2rem;
   }
 
   .copy-clipboard {
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -376,6 +376,10 @@ body {
     font-weight: bold;
   }
 
+  #sidebar, #main-content {
+    background-color: var(--vocab-bg);
+  }
+
   /*** Sidebar termpage & vocabpage ***/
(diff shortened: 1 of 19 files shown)


tests/cypress/template/concept.cy.js
kinow, line 1: :clap: :1st_place_medal:

kinow (approved):

osma:
I see that there's a little regression on the vocabulary home page, for example some background colors have changed. This is most likely due to shared CSS rules. I will try to fix that before merging.

resource/css/skosmos.css
```diff
@@ -376,6 +376,10 @@ body {
     font-weight: bold;
   }
 
+  #sidebar, #main-content {
+    background-color: var(--vocab-bg);
```
joelit, line 380: This affects now all main-content elements.
 - Maybe we need to either further divide the css definition to #main-content.vocabPage, #main-content.conceptPage
 - Or rename the vocab-bg to be something more generic, since leaving it like this would mean changing the color variable would change the lolours of at least vocab-info and error pages

src/view/concept-card.inc
```diff
@@ -1,191 +1,76 @@
...
-              </td>
-            </tr>
+
+      {% for property in concept.properties %}
+      <div class="row property prop-{{property.ID}}">
+        <div class="col-sm-4 px-0 property-label"><h2>{{ property.label}}</h2></div>
+        <div class="col-sm-8 align-self-center property-value">
+          <ul class="align-bottom">
+          {% for propval in property.values %}
+            {% if propval.uri %} {# resources with URI #}
```
joelit, line 22: This could have a clause for handling rdfs:type-property so that the property value would have only the textuar representatio of the class without a link to the rdf-representation of the class.

tests/cypress/template/concept.cy.js
joelit, line 1: This sets a high bar for future cypress tests! :)

joelit (changes requested):
I made a couple of suggestions for this PR. I also wrote further requirements for concept pages in the comments for the associated issue #1484 that would be out of scope for this PR. 

his is a solid start! Thanks especially for dismantling some of the old Skosmos 2.x template logic that is no longer in use. The cypress tests are also a sight worth seeing. Two michelin stars!

src/view/concept-card.inc
osma: Fixed

resource/css/skosmos.css
osma: I moved the background-color setting to `.main-content-section` class, which is more specific and allows for multiple sections within the main-content area. I think it solves the problem well enough for now.

osma:
Still TODO:

- [ ] fix headline styling regressions on feedback and about pages
- [ ] fix background color regressions on vocab-home page

tests/cypress/template/concept.cy.js
```diff
@@ -1,3 +1,120 @@
...
+    cy.get('.prop-skos_definition .reified-property-value').invoke('text').should('contain', 'The black sea bass')
+  })
+  it('contains broader concept', () => {
+    cy.visit('/yso/en/page/p21685') // go to "music research" concept page
+
+    // check the property name
+    cy.get('.prop-skos_broader .property-label').invoke('text').should('equal', 'Broader concept')
+
+    // check the broader concept
+    cy.get('.prop-skos_broader .property-value a').invoke('text').should('equal', 'research')
```
joelit, line 51: It seems this test is failing

tests/cypress/template/concept.cy.js
```diff
@@ -1,3 +1,120 @@
...
+    cy.visit('/yso/en/page/p38289') // go to "music archaeology" concept page
+
+    // check the property name
+    cy.get('.prop-skosmos_memberOf .property-label').invoke('text').should('equal', 'Belongs to group')
+
+    // check that we have the correct number of groups
+    cy.get('.prop-skosmos_memberOf .property-value').find('li').should('have.length', 1)
+
+    // check the first group value
+    cy.get('.prop-skosmos_memberOf .property-value a').invoke('text').should('equal', '51 Archaeology')
```
joelit, line 93: Same here, this test failed

joelit (changes requested):
There seems to be problems with two cypress tests that should be fixed

tests/cypress/template/concept.cy.js
osma: I'm on it

Issue #1514, opened 2023-09-18
Epic: Vocabulary Search Bar


joelit:
## Description of the enhancement

PR #1470  contains the first implementation of vocabulary search results page. This needst to be built upon, before we have a demo-ready feature for the search in whole.

## Who are the users that would benefit from the enhancement and how?

Getting this ready would be one of the first Skosmos demos, so the users would be testers and the audiences interested in Skosmos 3.0 development

### Vocabulary search field specification

1. General functionality
  - [x] 1.1 Twig-templates, Bootstrap-layout and CSS
  - [x] 1.2 Search field can be emptied by clicking X
  - [x] 1.3 Search can be executed with a chosen content language, or with all languages (cypress tests for this)
  - [x] 1.4 Dropdown menu styling should match design expectations (Bootstrap, if possible)
    - [x] 1.4.1 Dropdown menu scroll bar can be styled to colour scheme?
    - [x] 1.4.2 Menu item highlight elements should be designed according to unified scheme
  - [x] 1.5 Once user has stopped typing after 300 ms, submit the search
   - [x] 1.5.1 Process the search results into a stored object that facilitates caching and interrupting
  - [x] 1.6 Append an asterix after the search term, if the search term does not contain an asterix at some point
  - [x]  1.7 Once the search results are calculated, dispaly the dropdown menu
   - [x] 1.7.1 If there are no results, show the dropdown with no results -message
   - [x] 1.7.2 If the user writes more text in addition to previously given input, show the result of the latest input
  - [x] 1.8 Hide the dropdown list, if the user:
    - [x] 1.8.1 Clears the text box from the clear-button
    - [x] 1.8.2 Deletes the contentents of the input field
    - [x] 1.8.3 Clicks on somewhere outside of the search result dropdown
2. Language options
  - [x] 2.1 Default language is chosen from a cookie (cypress tests for this)
  - [x] 2.2 Changing the search language changes the content language
    - [x] 2.2.1 Udpating the content language updates the cookie (cypress tests for this)
    - [x] 2.2.2 Updating the content language executes a partial page load (cypress tests for this)
  - [x] 2.2 Contents for the language menu are defined in the  SKOSMOS JS-object (cypress tests for this)
  - [x] 2.3 All UI text is in English (in the form of translation message IDs)
  - [x] 2.4 Dropdown menu language codes should be translated to the UI language
  - [x] 2.5 The translation component provides the translated UI messages (cypress tests for this)
[...]

PR #1551, opened 2023-11-16
Issue1514 search bar

joelit:

Draft PR for search field

- #1514

## Checklist


src/view/scripts.inc
```diff
@@ -43,6 +43,14 @@ const SKOSMOS = {
...
 <script src="node_modules/bootstrap/dist/js/bootstrap.bundle.js"></script>
 
+<!-- Enable Bootstrap popovers -->
+<script>
+    var popoverTriggerList = [].slice.call(document.querySelectorAll('[data-bs-toggle="popover"]'))
+    var popoverList = popoverTriggerList.map(function (popoverTriggerEl) {
+      return new bootstrap.Popover(popoverTriggerEl)
+    })
+</script>
+
```
joelit, line 53: This is for the search tips to work as a clickable popover element in Bootstrap. I did not include the search tips box in this PR yet.

tests/cypress/template/vocab-search-bar.cy.js
```diff
@@ -0,0 +1,64 @@
...
+      
+      cy.get('#search-wrapper select').find('option').each(($option, index) => {
+        const optionValue = $option.attr('value');
+        const expectedLanguage = languageOrder[index];
+
+        // Check that the search language options values matche the languageOrder
+        cy.wrap(optionValue).should('eq', expectedLanguage);
+      });
+    })
+    */
```
joelit, line 62: Generally, this kind of commented test should not be in the PR. In this case, I left the WIP in here in case I cannot attend during the rest of the sprint, and somebody wants to have a go with writin the test.

osma (changes requested):
Looks like a good start.

Regarding the layout/style, this doesn't follow the design exactly. The border is too dark, and I don't think we want rounded corners in the dropdown menu: 

image

You mentioned that "Search field can be emptied by clicking X" is implemented, but I didn't see any such feature.

Here are problems that should either be fixed in this PR or added to the list of requirements in the issue #1514 so that they can be addressed later:

- border color should follow design spec
- dropdown style should match design expectations (no rounded borders)
- language names should be shown, not language codes
- in general, UI texts should be translated (currently a problem for all Vue components)

joelit:
It seems I need to adjust the behaviour of the search form. Chrome looks like this:

Screenshot from 2023-11-22 14-13-25
Screenshot from 2023-11-22 14-13-35

[Apparently the rendering of the options-menu of the form is up to the browser](https://www.456bereastreet.com/archive/200409/styling_form_controls/) ( -_- ), and different browsers yield different results. It could be that the menu should intead be implemented in a more ad-hoc manner to get more control over the visuals.

Same is probably true for the reset text 'x' icon. I'll investigate.

src/view/scripts.inc
osma: Let's remove it from this PR as it's not yet needed

osma (approved):
LGTM! Can be merged after CI tests have passed

PR #1557, opened 2023-11-22
Implement max length for input fields on the feedback page

rvdwxl:
…fields

This PR adds a maximum input length to the "name", "email", and "subject" input fields on the feedback page, both to the frontend and backend. A maximum length was not previously implemented, and allows users to input very large strings either in the form, or by hijacking the POST-request.

- none

This PR imposes a limit of 255 characters to the "name", "email", and "subject" input fields on the feedback page; both in the frontend (feedback.twig, by using attribute "maxlength") and the backend (WebController, by using "substr").

- The limit of 255 characters is hard coded.

## Checklist

### DIFF
--- a/controller/WebController.php
+++ b/controller/WebController.php
@@ -215,9 +215,9 @@ class WebController extends Controller
         if ($request->getQueryParamPOST('message')) {
             $feedbackSent = true;
             $feedbackMsg = $request->getQueryParamPOST('message');
-            $feedbackName = $request->getQueryParamPOST('name');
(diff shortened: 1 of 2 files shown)


osma:
Thanks for the PR @rvdwxl !

Hardcoding the limit to 255 characters is a little ugly, but I think this is still an improvement, so this can be merged to Skosmos 2.x and included in the upcoming 2.17 release (probably the last one in the Skosmos 2.x) series.

For Skosmos 3 I think we need a more detailed specification on how to handle this. I will open a separate issue about that shortly.

PR #1558, opened 2023-11-22
Issue1476 cypress tests for vocab home page tempate statistics

miguelvaara:
All tests pass (including the new ones related to statistics), except for the "alphabetical index letters are clickable" test, which fails randomly. It sometimes passes if the entire vocab-home test is run again, but after making changes to the test, it sometimes fails and sometimes does not when run separately. I suspect it is related to some clicking issue.

This is an early-stage draft PR, so it does not aim to meet code formatting requirements or have sufficient descriptions. The code will be commented later.

- Closes #

## Checklist

### DIFF
--- a/tests/cypress/e2e/fp-lang-yso-search-result-vocab.sy.js
+++ b/tests/cypress/e2e/fp-lang-yso-search-result-vocab.sy.js
@@ -1,0 +1,21 @@
+describe('Front page - lang - list - yso - search - result - vocab', () => {
+    it('should navigate through Finto application', () => {
+      cy.visit('/en/');
+  
+      cy.contains('Change language').click();
+      cy.contains('Suomeksi').click();
+  
+      // Select "YSO - Yleinen suomalainen ontologia" from the vocabulary list
+      cy.get('#tähän_sopiva_id').select('YSO - Yleinen suomalainen ontologia');
+  
+      cy.visit('/yso/fi/search?query=kissaeläimet');
+  
+      // Select the concept 'kissaeläimet' from the list and go to the concept page
+      cy.contains('joku sopiva käsite').click();
+  
+      // After transitioning from the search results page to the concept page, you should enter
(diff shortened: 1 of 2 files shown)


osma:
> All tests pass (including the new ones related to statistics), except for the "alphabetical index letters are clickable" test, which fails randomly.

I looked at this and was able to replicate the occasional text failure.

It looks like this test revealed a race condition in the alphabetical index Vue component. It starts loading the index entries for the letter A, but the user (here simulated by Cypress) clicks on D before that operation has finished. This leads to a race condition with both requests (for A and D) happening in parallel, and depending on the order in which they finish, the index will contain either 8 or 33 entries.

This is probably a more general problem with all kinds of asynchronous operations performed by Vue components. We should have some sort of queue mechanism, or a way to cancel/invalidate previous operations. In this case, when the user clicks on D before the request for A has finished, it would make sense to cancel the request for A.

osma:
In Skosmos 2, similar cancellations and race conditions were handled by using a queue for Ajax requests. This implementation of course relies on jQuery functionality, which doesn't exist in Skosmos 3, so another solution is needed.

https://github.com/NatLibFi/Skosmos/blob/88c2c781be7dc00a65a6167baf087b39354f8d13/resource/js/scripts.js#L86-L136

miguelvaara:
I think we now have four different ways to proceed with this problem:

- building now some kind of queuing mechanism to eliminate the root cause
- separating the statistical tests of the vocabulary page from this test
- accepting that errors will occur occasionally (then implementing TDD rather poorly)
- postponing the developing of the statistical tests (I think we should have them) until the competition situation issue is fixed

Which approach would you suggest @osma ?

osma:
I think we need a new issue for the race condition. It is unrelated to the tests in this PR, so we can proceed with finishing up this PR.

miguelvaara:
The PR is not needed anymore

PR #1562, opened 2023-11-23
Add logos for Skosmos 3

UnniKohonen:

Skosmos 3 is missing logos from all pages, this PR adds a logo to headerbar on landing page and to topbar on other pages.

- Closes #

- Add .svg files for logos
- Add logo to landing page
- Add logo to other pages

Headerbar (and thus logo) is visible on about and feedback pages, which should be fixed.

## Checklist


src/view/base-template.twig
```diff
@@ -28,7 +28,13 @@
   <a class="visually-hidden" id="skiptocontent" href="{{ request.langurl }}#maincontent">Skip to main</a>
   <header>
     <div class="container-fluid bg-dark d-flex my-auto py-3 px-4 text-bg-dark">
-      <span class="fs-6 py-3 text-light">Yhteishaku sanastoista v</span>
+      {% if request.vocabid == '' and request.page != 'about' and request.page != 'feedback' %}
+        <span class="fs-6 py-3 text-light">Yhteishaku sanastoista v</span>
+      {% else %}
+        <a id="skosmos-logo-top" href="{{ request.lang }}/{% if request.contentLang and request.contentLang != request.lang %}?clang={{ request.contentLang }}{% endif %}{% if request.queryParam('anylang') == 'on' %}{% if request.contentLang == request.lang %}?{% else %}&{% endif %}anylang=on{% endif %}">
```
osma, line 34: This is a terrible way of generating the link URL for the home page, but I see that it has been used in many other places as well (a few lines down in the same file) so I guess one more time doesn't hurt much.
However, I think a better mechanism should be developed. We already have the custom `link_url` filter, maybe it could be used instead, with some enhancements?

osma:
Looks good and gets the job done. I commented about the URL generation, but that is a more general problem, not specific to this PR. I'll merge this as it is.

PR #1568, opened 2023-12-05
Avoid showing vocabulary modified timestamp on concept page when the concept doesn't have a specific modification date

### LINKED ISSUE #1343: Skosmos shouldn't show the modified timestamp for a concept page unless such is explicitly stated for the concept data

http://www.yso.fi/onto/yso/p508596

1. The concept page mentions a modified date - "Last modified 3/5/21"
2. The concept data in turtle format does not contain dct:modified of dct:created timestamps
3. The last modified data comes from the vocabulary's main concept page's info

The last modified date should be only shown when the concept has such data. Skosmos can still use the vocabulary's last modified data if such exists, to cache http requests, but this can be separate from the data displayed for a concept.

Chromium

osma:

See #1343

- Closes #1343

- introduce a new method Concept.getConceptModifiedDate that only returns the specific modification date of the concept but doesn't fall back to the vocabulary-level modified timestamp
- use getConceptModifiedDate when showing modified timestamps on the concept page
- unit test to verify

Some unit tests will fail with deprecation warnings on PHP 8.1 for unrelated reasons - see PR #1567 for the explanation and fix.

## Checklist

### DIFF
--- a/model/Concept.php
+++ b/model/Concept.php
@@ -698,9 +698,9 @@ class Concept extends VocabularyDataObject implements Modifiable
     }
 
     /**
-     * @return DateTime|null the modified date, or null if not available
+     * @return DateTime|null the modified date of this concept, or null if not available
      */
-    public function getModifiedDate()
+    public function getConceptModifiedDate()
     {
         // finding the modified properties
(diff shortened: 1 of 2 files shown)


joelit (commented):
Looks good, works in my environment

Issue #1561, opened 2023-11-23
Race conditions in Skosmos 3 AJAX requests


osma:
## URL address of the page where you encountered the problem

http://localhost/Skosmos/yso/en/

## Description of the problem

When quickly clicking on entries in the alphabetical index, it's possible to cause a situation where more than one request is happening in parallel. This may cause a situation where the wrong result is shown to the user.

This happened recently with a Cypress test for the alphabetical index; see https://github.com/NatLibFi/Skosmos/pull/1558#issuecomment-1822927410

![image](https://github.com/NatLibFi/Skosmos/assets/1132830/c14a8650-22da-48e2-ade4-aa932f33d547)

Details from above:

> It looks like this test revealed a race condition in the alphabetical index Vue component. It starts loading the index entries for >the letter A, but the user (here simulated by Cypress) clicks on D before that operation has finished. This leads to a race >condition with both requests (for A and D) happening in parallel, and depending on the order in which they finish, the index will >contain either 8 or 33 entries. The test expects 8, so it will fail if there are 33 entries.

>This is probably a more general problem with all kinds of asynchronous operations performed by Vue components. We should >have some sort of queue mechanism, or a way to cancel/invalidate previous operations. In this case, when the user clicks on D >before the request for A has finished, it would make sense to cancel the request for A.

I think we need some kind of mechanism to ensure that there are not more than one similar AJAX-style requests happening simultaneously. In Skosmos 2, a queue for AJAX requests was used to solve similar problems; see https://github.com/NatLibFi/Skosmos/pull/1558#issuecomment-1822941901

Although this happened only in the alphabetical index component, the problem is more general and also applies to partial page loads triggered from e.g. the alphabetical index and the hierarchy component. Also in that case, it is possible to click on many entries so quickly that multiple requests are "in the air" at the same time and they could finish in any order. So we need a general mechanism to ensure that only the last action (last click) determines the outcome; probably the old request should be cancelled when an user action triggers a new request of the same kind.

osma:
I think this could be done using an AbortController to [abort the previous fetch operation](https://javascript.info/fetch-abort).

Here's what Phind.com suggested:

> To modify the partialPageLoad function to cancel the previous fetch that hasn't yet finished using an AbortController, follow these steps:
>
>   1. Create an AbortController instance outside the function scope to ensure it persists across multiple function calls. This allows you to abort the previous fetch when a new one is initiated.
[...]

PR #1589, opened 2024-01-31
Implement autocomplete for search

joelit:
Autocomplete functionality for #1514 

This PR introduces a Bootstrap 5 dropdown menu for autocomplete, and it aims to fulfill the basic functionality of the search bar's autocomplete part.

- [x] Once user has stopped typing aftes X ms, submit the search
- [x] Append an asterix after the search term
- [x] Autocomplete should fire if there is at least two characters in the input string
- [x] Process the search results into autoCompeteResults
- [x] If the user writes more text in addition to previously given input, show the result of the latest input
- [x] When the result list is calculated, display the dropdown
- [x] If there are no results, display a dropdown with no results -message
- [x] Hide the dropdown list, if the user
  - [x] Clears the text box from the clear-button
  - [x] Deletes the contentents of the input field
  - [x] Clicks on somewhere outside of the search result dropdown
- [x] Input element should be rectangular
- [x] Dropdown menu should be styled with Bootsrap

 - Bootstrap 5 dropdown menu is designed to fire on click. If and when this is not desired, the showing and hiding has to be triggered in JS, which is rather odd
 
 
## Checklist


resource/css/skosmos.css
joelit, line 1: The CSS is not in line with Bootstrap yet, it's just something I pulled from a tutorial.

resource/js/vocab-search.js
```diff
@@ -7,18 +7,119 @@ const vocabSearch = Vue.createApp({
...
     autoComplete () {
+      const delayMs = 300
+
+      // when new autocomplete is fired, empty the previous result
+      this.hideDropdown()
+
+      // cancel pending API calls when method is called
+      clearTimeout(this._timerId)
+
+      // is the search term is in cache, use the cache
```
osma, line 38: Should be marked with TODO

resource/js/vocab-search.js
```diff
@@ -7,18 +7,119 @@ const vocabSearch = Vue.createApp({
...
+    document.removeEventListener('click', this.onClickOutside)
   },
   methods: {
     autoComplete () {
+      const delayMs = 300
+
+      // when new autocomplete is fired, empty the previous result
+      this.hideDropdown()
+
+      // cancel pending API calls when method is called
```
osma, line 35: I first misunderstood this comment and thought this would cancel the API calls "in the air". But it actually cancels just the timer. Maybe change to "cancel timer for upcoming API call" (note singular, as there can be only one), or something along those lines.

resource/js/vocab-search.js
```diff
@@ -7,18 +7,119 @@ const vocabSearch = Vue.createApp({
...
+
+      // is the search term is in cache, use the cache
+
+      // delay call, but don't execute if the search term is not at least two characters
+      if (this.searchTerm.length > 1) {
+        this._timerId = setTimeout(() => { this.search() }, delayMs)
+      }
+    },
+    /*
+     * search should fetch the response and save it to cache
```
osma, line 46: Should be marked with TODO

resource/js/vocab-search.js
```diff
@@ -7,18 +7,119 @@ const vocabSearch = Vue.createApp({
...
+     * search calls renderResults for displaying the response
```
osma, line 47: Is this comment necessary? It confused me a little and doesn't seem to add much, the code is clear. Remove?

resource/js/vocab-search.js
```diff
@@ -7,18 +7,119 @@ const vocabSearch = Vue.createApp({
...
+     */
+    search () {
+      const mySearchCounter = this.searchCounter + 1 // make sure we can identify this search later in case of several ongoing searches
```
osma, line 50: This counter logic is one way of avoiding the problems with AJAX race conditions (#1561) seen in partial page loads. Maybe a more general mechanism could handle both these API calls and the partial page loads, as well as other types of dynamic HTTP calls?

resource/js/vocab-search.js
```diff
@@ -32,24 +133,57 @@ const vocabSearch = Vue.createApp({
     },
     changeLang () {
       SKOSMOS.content_lang = this.selectedLanguage
-      // TODO: Impelemnt partial page load to change content according to the new content language
+      // TODO: Impelement partial page load to change content according to the new content language
```
osma, line 136: You still didn't fix all the typos in the word "Implement" ;)

resource/js/vocab-search.js
```diff
@@ -7,18 +7,119 @@ const vocabSearch = Vue.createApp({
...
+    },
+    /*
+     * search should fetch the response and save it to cache
+     * search calls renderResults for displaying the response
+     */
+    search () {
+      const mySearchCounter = this.searchCounter + 1 // make sure we can identify this search later in case of several ongoing searches
+      this.searchCounter = mySearchCounter
+
+      fetch('rest/v1/' + SKOSMOS.vocab + '/search?query=' + this.formatSearchTerm() + '&lang=' + SKOSMOS.lang)
```
osma, line 53: I think the search term should be URL escaped here. For example, it's currently not possible to search for `r&` (which should match `r&b` and the altlabel `R&D` in YSO).

The best way to build URLs like this is nowadays to use the [URL class](https://www.valentinog.com/blog/url/), which nicely avoids all escaping issues.

resource/js/vocab-search.js
```diff
@@ -7,18 +7,119 @@ const vocabSearch = Vue.createApp({
...
+      this.renderedResultsList.forEach(result => {
+        const hitPref = this.renderMatchingPart(renderedSearchTerm, result.prefLabel)
+        const hitAlt = this.renderMatchingPart(renderedSearchTerm, result.altLabel)
+        const hitHidden = this.renderMatchingPart(renderedSearchTerm, result.hiddenLabel)
+        if ('uri' in result) { // change uris to Skosmos page urls
+          result.uri = SKOSMOS.vocab + '/' + SKOSMOS.lang + '/page?uri=' + encodeURIComponent(result.uri)
+        }
+        // render search result labels
+        if (hitHidden) {
+          result.rendered = '<a href="' + result.uri + '">' + result.prefLabel + '</a>'
```
osma, line 95: Building up the HTML like this seems fragile, prone to e.g. escaping issues and hard to maintain. Could we use the Vue templating mechanism here as well?

resource/js/vocab-search.js
```diff
@@ -7,18 +7,119 @@ const vocabSearch = Vue.createApp({
...
+    },
+    renderMatchingPart (searchTerm, label) {
+      const regex = new RegExp(searchTerm, 'i') // case insensitive matching
+      if (label && regex.test(searchTerm)) {
+        return label.replace(regex, (match) => `<b>${match}</b>`)
+      }
+      return label
+    },
+    translateType (type) {
+      return SKOSMOS.msgs[SKOSMOS.lang][type]
```
osma, line 75: This method of displaying human-readable type labels only works in the case where the types are defined in UI translations (for which the infrastructure in Skosmos 3 is still lacking). But Skosmos 2 has another mechanism as well: the frontend will call the `types` REST API method and use its results to display human-readable type labels, which are typically vocabulary-specific types that have `rdfs:label` properties in the vocabulary data.

I don't think this PR needs to use the `types` method or otherwise know how to display vocabulary-specific type labels, but this missing feature should be put on the TODO list if it's not yet there. (Though it doesn't have to be implemented the same way as in Skosmos 3; maybe the type information could be passed to the frontend code already in the SKOSMOS object, because it's going to be needed in most cases anyway. This would reduce the number of REST API calls.)

resource/js/vocab-search.js
```diff
@@ -7,18 +7,119 @@ const vocabSearch = Vue.createApp({
...
+    renderResults () {
+      // get the results list form cache if it is implemented
+      console.log(this.renderedResultsList)
+      const renderedSearchTerm = this.searchTerm // save the search term in case it changes while rendering
+      this.renderedResultsList.forEach(result => {
+        const hitPref = this.renderMatchingPart(renderedSearchTerm, result.prefLabel)
+        const hitAlt = this.renderMatchingPart(renderedSearchTerm, result.altLabel)
+        const hitHidden = this.renderMatchingPart(renderedSearchTerm, result.hiddenLabel)
+        if ('uri' in result) { // change uris to Skosmos page urls
+          result.uri = SKOSMOS.vocab + '/' + SKOSMOS.lang + '/page?uri=' + encodeURIComponent(result.uri)
```
osma, line 91: This is maybe a nitpick, but I think it would be cleaner to assign this to `const result_url`. It's better to avoid mutating already existing variables and data structures as it makes code logic easier to inspect.

resource/js/vocab-search.js
```diff
@@ -7,18 +7,119 @@ const vocabSearch = Vue.createApp({
...
+      return SKOSMOS.msgs[SKOSMOS.lang][type]
+    },
+    /*
+     * renderResults is used when the search string has been indexed in the cache
+     * it also shows the autocomplete results list
+     * TODO: Showing labels in other languages, extra concept information and such goes here
+     */
+    renderResults () {
+      // get the results list form cache if it is implemented
+      console.log(this.renderedResultsList)
```
osma, line 84: Should be removed eventually

resource/js/vocab-search.js
```diff
@@ -7,18 +7,119 @@ const vocabSearch = Vue.createApp({
...
+    translateType (type) {
+      return SKOSMOS.msgs[SKOSMOS.lang][type]
+    },
+    /*
+     * renderResults is used when the search string has been indexed in the cache
+     * it also shows the autocomplete results list
+     * TODO: Showing labels in other languages, extra concept information and such goes here
+     */
+    renderResults () {
+      // get the results list form cache if it is implemented
```
osma, line 83: Should be marked with TODO

resource/css/skosmos.css
```diff
@@ -275,6 +279,55 @@ body {
...
+
+  #headerbar .autocomplete-result {
+    list-style: none;
+    text-align: left;
+    padding: 4px 2px;
+    cursor: pointer;
+  }
+
+  #headerbar .autocomplete-result:hover {
+    background-color: #4AAE9B;
```
osma, line 325: Why this green shade? It doesn't seem to fit in with the rest of the current palette. I would expect dark blue or gray... Also the contrast between the blue link text and the greenish background is quite low.

Maybe we just need a proper visual design for the autocomplete list...

osma (changes requested):
This is a great start for the autocomplete functionality. I added specific comments to the code. General comments:

- The PR should have a human readable name (e.g. "Implement autocomplete for search")
- It seems that keyboard navigation hasn't been implemented. This is critical for accessibility. It should either be added on this branch (along with cypress tests) or put on a TODO list for later implementation.
- Some Cypress tests have been left out; these should probably be implemented as well.
- There are quite a few commits now and some of them are unrelated comments that nowadays also exist on the main branch. The diff view is quite messy. I suggest rebasing this on current main. After rebase, the commits specific to this branch could perhaps be squashed to a single commit (or a small number of commits). Alternatively, the PR could eventually be merged with the squash option so it becomes just a single commit on the main branch.

resource/js/vocab-search.js
joelit: I added this to the relevant issue as "3.5.1 Search result types should be fetched from rest/v1/types and stored in SKOSMOS object".

resource/js/vocab-search.js
joelit: Yeah, I was thinking of an abort controller -based general solution.

joelit:
I changed the name of the PR and added missing tests. I found one bug with enter search related to encoding the search term, which I also fixed. I also fixed the comments and wrote further specs in the the main issue #1514 .

The PR should be ready for another round! I didn't squash the commits here, fearing I'd mess it up.

osma:
FYI, I changed the base branch of this PR briefly to `skosmos-2` and then back to `main`. This made the diff view much cleaner (before: 14 changed files, after: only 4 changed files) as irrelevant changes that had already been merged to main were dropped from the diff view.

Currently reviewing.

resource/js/vocab-search.js
```diff
@@ -7,18 +7,116 @@ const vocabSearch = Vue.createApp({
...
+        .then(data => data.json())
+        .then(data => {
+          if (mySearchCounter === this.searchCounter) {
+            this.renderedResultsList = data.results // update results (update cache if it is implemented)
+            this.renderResults() // render after the fetch has finished
+          }
+        })
+    },
+    formatSearchTerm () {
+      if (this.searchTerm.includes('*')) { return this.searchTerm }
```
osma, line 60: encodeURIComponent is not called in this case, I think it should be?

Example: Try searching for `r&*` in YSO. It should match two concepts, but doesn't show any results.

resource/js/vocab-search.js
```diff
@@ -7,18 +7,116 @@ const vocabSearch = Vue.createApp({
...
+      const formatted = this.searchTerm + '*'
+      return encodeURIComponent(formatted)
+    },
+    renderMatchingPart (searchTerm, label) {
+      const regex = new RegExp(searchTerm, 'i') // case insensitive matching
```
osma, line 65: The search term isn't technically a RegEx, so e.g. characters that have a special meaning in regular expressions will probably not work properly and may even cause breakage. I suggest changing this to plain substring matching instead.

Example: Try entering `A(` in the search field (YSO, Finnish or English). This should match e.g. `A(H1N1) virus`. But instead this error shows up in the JS console:

    Uncaught (in promise) SyntaxError: unterminated parenthetical

resource/js/vocab-search.js
```diff
@@ -7,18 +7,116 @@ const vocabSearch = Vue.createApp({
...
+
+      const renderedSearchTerm = this.searchTerm // save the search term in case it changes while rendering
+      this.renderedResultsList.forEach(result => {
+        const hitPref = this.renderMatchingPart(renderedSearchTerm, result.prefLabel)
+        const hitAlt = this.renderMatchingPart(renderedSearchTerm, result.altLabel)
+        const hitHidden = this.renderMatchingPart(renderedSearchTerm, result.hiddenLabel)
+        if ('uri' in result) { // change uris to Skosmos page urls
+          result.pageUrl = SKOSMOS.vocab + '/' + SKOSMOS.lang + '/page?uri=' + encodeURIComponent(result.uri)
+        }
+        // render search result labels
```
osma, line 90: I think I commented on this in the previous review, but here goes again:

> Building up the HTML like this seems fragile, prone to e.g. escaping issues and hard to maintain. Could we use the Vue templating mechanism here as well?

resource/js/vocab-search.js
osma: OK good. I edited the wording slightly to this: "Labels for search result types should be provided by the backend in the SKOSMOS object and displayed in search results"
...because the types don't need to be fetched from the REST API if they are already given in the SKOSMOS object.

osma (changes requested):
Much better than the previous version, but still I found some issues; see the detailed comments.

Regarding the UI: Now only the prefLabel is a clickable link. I think the whole row should be clickable, as it is in Skosmos 2 and most other autocomplete implementations (e.g. Google search box).

joelit:
Changed the way special characters are handled for the search itself, and how the matching part of the search results are displayed. Also changed URL parameters into URLSearchParams objects, and adjusted cypress tests to reflect these changes. :+1:

resource/js/vocab-search.js
```diff
@@ -6,49 +6,246 @@ const vocabSearch = Vue.createApp({
       languages: [],
       selectedLanguage: null,
       searchTerm: null,
-      autoCompeteResults: [],
-      languageStrings: null
+      searchCounter: null,
+      renderedResultsList: [],
+      languageStrings: null,
+      msgs: null
```
UnniKohonen, line 12: I would add a variable to control the visibility of the dropdown list instead of manipulating DOM directly
```suggestion
      msgs: null
      showDropdown: false
```

resource/js/vocab-search.js
```diff
@@ -6,49 +6,246 @@ const vocabSearch = Vue.createApp({
...
     autoComplete () {
+      const delayMs = 300
+
+      // when new autocomplete is fired, empty the previous result
+      this.renderedResultsList = []
+
+      // cancel the timer for upcoming API call
+      clearTimeout(this._timerId)
+      const element = document.getElementById('search-autocomplete-results')
+      element.classList.remove('show')
```
UnniKohonen, line 37: `hideDropdown` method could be called here

resource/js/vocab-search.js
```diff
@@ -6,49 +6,246 @@ const vocabSearch = Vue.createApp({
...
+      })
+
+      if (this.renderedResultsList.length === 0) { // show no results message
+        this.renderedResultsList.push({
+          prefLabel: this.msgs['No results'],
+          lang: window.SKOSMOS.lang
+        })
+      }
+      const element = document.getElementById('search-autocomplete-results')
+      element.classList.add('show')
```
UnniKohonen, line 127: `showAutoComplete` method could be called here

resource/js/vocab-search.js
```diff
@@ -6,49 +6,246 @@ const vocabSearch = Vue.createApp({
...
+    },
+    hideDropdown () {
+      const element = document.getElementById('search-autocomplete-results')
+      element.classList.remove('show')
```
UnniKohonen, line 131: Instead of removing class directly, I would use a variable to control dropdown visibility, for example:
```suggestion
      this.showDropdown = false
```

resource/js/vocab-search.js
```diff
@@ -6,49 +6,246 @@ const vocabSearch = Vue.createApp({
...
+      if (listener && !listener.contains(event.target)) {
+        this.hideDropdown()
+      }
+    },
+    /*
+     * Show the existing autocomplete list if it was hidden by onClickOutside()
+     */
+    showAutoComplete () {
+      const element = document.getElementById('search-autocomplete-results')
+      element.classList.add('show')
```
UnniKohonen, line 167: Instead of adding class directly, I would use a variable to control dropdown visibility, for example:
```suggestion
      this.showDropdown = true
```

resource/js/vocab-search.js
```diff
@@ -6,49 +6,246 @@ const vocabSearch = Vue.createApp({
...
+            aria-expanded="false"
+            autocomplete="off"
+            data-bs-toggle=""
+            aria-label="Text input with dropdown button"
+            placeholder="Search..."
+            v-model="searchTerm"
+            @input="autoComplete()"
+            @keyup.enter="gotoSearchPage()"
+            @click="showAutoComplete()">
+          <ul id="search-autocomplete-results" class="dropdown-menu"
```
UnniKohonen, line 192: I would set the class of `search-autocomplete-results` with Vue instead of manipulating DOM directly, for example like:
```suggestion
          <ul id="search-autocomplete-results" class="dropdown-menu" :class="{ 'show': showDropdown }"
```

resource/js/vocab-search.js
```diff
@@ -6,49 +6,246 @@ const vocabSearch = Vue.createApp({
...
   },
   mounted () {
     this.languages = window.SKOSMOS.languageOrder
     this.selectedLanguage = window.SKOSMOS.content_lang
-    this.languageStrings = window.SKOSMOS.language_strings[window.SKOSMOS.lang]
+    this.searchCounter = 0
+    this.languageStrings = window.SKOSMOS.language_strings[window.SKOSMOS.lang] ?? window.SKOSMOS.language_strings.en
+    this.msgs = window.SKOSMOS.msgs[window.SKOSMOS.lang] ?? window.SKOSMOS.msgs.en
+    this.renderedResultsList = []
+    document.addEventListener('click', this.onClickOutside)
```
UnniKohonen, line 22: This (and line 25) could be done like in for example hierarchy tab using a custom directive: https://github.com/NatLibFi/Skosmos/blob/3925c7162660286cb20903e73c129dbed4a1669f/resource/js/tab-hierarchy.js#L170-L180

It is probably not necessary and could be left as it is, though

UnniKohonen (commented):
Added some minor suggestions for CSS.

In the Vue component, I would change the way the visibility of the dropdown list is handled. Instead of removing and adding the list's class directly, I would use a variable and bind the class in the template. It works the way it's done now, but IMO it would be a little cleaner to use Vue's own functionality.

The event listener for clicking outside the component could be done using a custom directive like in the other Vue components, but this isn't probably necessary as it seems to work fine now.

Otherwise the code looks good and the component seems to work as intended.

Issue #1585, opened 2024-01-25
Upgrade to Twig 3 for Skosmos 3


osma:
## Description of the enhancement

Twig 2 was EOL'd in December 2023. Skosmos 2 and 3 are both still using Twig 2. Skosmos 3 should upgrade to Twig 3 because it doesn't make sense to build a new application on an obsolete version.

Fortunately the required changes should be small. It's possible that only the PHP side initialization will need some adjustment. See https://symfony.com/blog/preparing-your-applications-for-twig-3

## Who are the users that would benefit from the enhancement and how?

Skosmos developers and users

## What new functionalities would the enhancement make possible?

E.g. security and other updates to Twig.

## Why is the enhancement important?

This has to be done anyway because Twig 2 is EOL. It's better to do it early to avoid extra work later.

PR #1590, opened 2024-01-31
Copy to clipboard functionality for concept page

### LINKED ISSUE #926: Improvements to the copy to clip board buttons
@kouralex
> I'd appreciate if the tool tip text could be translated, though. Could you consider implementing this? For sure, this functionality should be replicated to the URI copying "button" (span in reality) as well. Additionally, expressing the actual copying in a transparent fashion could be done with a pop up-ish message such as "Copied!" or similar

These should be a separate issue on their own. As part of accessability enhancements, the copy-to-clipboard elements Skosmos has for various labels, IRIs and notations should contain alt text explaining what they do with appropriate translations to various UI languages. As it is primarily an accessability improvement, this and possible messages after copying the target literal should be tested with a screen reader or an accessability evaluation tool like http://wave.webaim.org/ .

### LINKED ISSUE #1504: New copy-to-clipboard component for Skosmso 3.0

Concept URLs and notation codes need a copy-to-clipboard component in concept.twig, global-search.twig and vocac-search.twig . This can be just vanilla JS located in the view/scripts.inc .
 - [ ] Layout follows the [static examples](https://github.com/NatLibFi/Skosmos/blob/skosmos-3-html/vocabpage.html)
 - [ ] cypress tests exist

osma:

Implement copy to clipboard functionality, which is one of the required features for the concept page (#1484).

- part of #1484 
- Closes #1504
- Closes #926

- add a vanilla JS function copyToClipboard
- use the copyToClipboard method for the prefLabel and URI on the concept page
- add translations (fi, sv, en) for the "Copy to clipboard" alt text used for the copy buttons
- remove the copy to clipboard icons from the search results page (this was probably a misunderstanding during layout design)
- add Cypress tests for copy to clipboard functionality

- the Cypress tests that access the clipboard may fail when run interactively (`npx cypress open`) if the Cypress window isn't focused (for example when working in an IDE) due to browser security policy; they always work when run headless (`npx cypress run`) and under CI

## Checklist

### DIFF
--- a/resource/js/copy-to-clipboard.js
+++ b/resource/js/copy-to-clipboard.js
@@ -17,14 +17,18 @@ async function copyToClipboard (id) {
 // register the copyToClipboard function as event an handler for the copy buttons
 const copyPrefElem = document.getElementById('copy-preflabel')
 if (copyPrefElem) {
-  copyPrefElem.addEventListener('click', async () => {
-    await copyToClipboard('concept-preflabel')
+  copyPrefElem.addEventListener('click', () => {
+    (async () => {
+      await copyToClipboard('concept-preflabel')
+    })()
   })
 }
 
 const copyUriElem = document.getElementById('copy-uri')
 if (copyUriElem) {
-  copyUriElem.addEventListener('click', async () => {
-    await copyToClipboard('concept-uri')
+  copyUriElem.addEventListener('click', () => {
+    (async () => {
+      await copyToClipboard('concept-preflabel')
+    })()
   })
 }
--- a/resource/js/copy-to-clipboard.js
+++ b/resource/js/copy-to-clipboard.js
@@ -17,10 +17,14 @@ async function copyToClipboard (id) {
 // register the copyToClipboard function as event an handler for the copy buttons
 const copyPrefElem = document.getElementById('copy-preflabel')
 if (copyPrefElem) {
-  copyPrefElem.addEventListener('click', () => copyToClipboard('concept-preflabel'))
+  copyPrefElem.addEventListener('click', async () => {
+    await copyToClipboard('concept-preflabel')
+  })
 }
 
 const copyUriElem = document.getElementById('copy-uri')
 if (copyUriElem) {
-  copyUriElem.addEventListener('click', () => copyToClipboard('concept-uri'))
+  copyUriElem.addEventListener('click', async () => {
+    await copyToClipboard('concept-uri')
+  })
 }
--- a/resource/js/copy-to-clipboard.js
+++ b/resource/js/copy-to-clipboard.js
@@ -1,0 +1,26 @@
+// function for copying the content from a specific element (by id) to the clipboard
+async function copyToClipboard (id) {
+  const copyElem = document.getElementById(id)
+  const sel = window.getSelection()
+  const range = document.createRange()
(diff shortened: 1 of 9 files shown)


osma:
Rebased on current `main` since PR #1585 was merged.

osma:
TODO:

- [ ] fix: buttons don't work after partial page load
- [ ] clicking on prefLabel or URI should select it
- [ ] copy button next a long prefLabel should stay on the same line if possible

osma:
SonarCloud complains about duplicated code in Cypress tests. Well, duh, tests are repetitive...

resource/js/copy-to-clipboard.js
```diff
@@ -0,0 +1,30 @@
...
+  const copyUriElem = document.getElementById('copy-uri')
+  if (copyUriElem) {
+    copyUriElem.addEventListener('click', () => copyToClipboard('concept-uri'))
+  }
+}
+
+// register the copyToClipboard function as event an handler for the copy buttons
+registerCopyToClipboardEvents()
+
+// re-register the event handlers after partial page loads
```
joelit, line 29: This line fills me with dread. We should make sure all event handlers are re-registered collectively. Partial page loading as a specific test case hasn't been used as of yet. Should we have an issue about it?

tests/cypress/template/concept.cy.js
```diff
@@ -95,6 +95,18 @@ describe('Concept page', () => {
...
+  it('concept preflabel can be copied to clipboard', () => {
+    cy.visit('/yso/en/page/p21685') // go to "music research" concept page
+
+    // click the copy to clipboard button next to the prefLabel
+    cy.get('#copy-preflabel').click()
+
+    // check that the clipboard now contains "music research"
+    // NOTE: This test may fail when running Cypress interactively in a browser.
+    // The reason is browser security policies for accessing the clipboard.
+    // If that happens, make sure the browser window has focus and re-run the test.
```
joelit, line 107: The things you learn..! -__-

joelit:
I would really want to see a hightlight for the copy button when used by keyboard.

Test case:
* Go to http://localhost/Skosmos/yso/fi/page/p23406
* Select "Uudet ja poistetut" form the side bar for ease switchng through interavtive elements
* Use tab to switch through active elements
* _After the breadcrumb, (and if using shift+tab, before the broader concept link) the point where the preferred label with copy-to-clipboard button is active, there is no visual hint like visible border to indicate this_
* When you press enter while the copy-button is actice, the relevant element is highlighted and the copy works as expected

joelit (changes requested):
Keyboard navigation needs a highlight hint, otherwise it looks good

resource/js/copy-to-clipboard.js
osma: I agree that this is a more general issue, but currently we don't have that many event handlers and AFAICT only the concept-mappings component needs to re-register its event handler after a partial page load, and it already does that.

tests/cypress/template/concept.cy.js
osma: :grin: Ideally the test would just be skipped when this happens. Not sure if that's easy to implement.

joelit (approved):
Tested, and it works for me - we can ask for a graphic designer's opinion later.

PR #1594, opened 2024-02-01
removed trailing commas from SKOSMOS object

joelit:

Further formatting of SKOSMOS object was needed for JSON parsing.

## Checklist


osma:
I think we need a Cypress test to catch these kinds of problems.

osma (approved):

PR #1599, opened 2024-03-05
fix problem introduced by docker compose 2.24.6 

osma:
This PR fixes a problem that appeared with the update of docker compose 2.24.6. Docker compose no longer supports a configuration where a base service declares a dependency on other services, and the base service is extended by another service. We had in the top level `docker-compose.yml` a base service `skosmos` that depends_on services `fuseki` and `fuseki-cache`, but for executing the test suite, there was another version of the `skosmos` service declared in `tests/docker-compose.yml` that extended the base service with the same name but with slightly different settings (e.g. environment variables). This used to work fine, now it breaks with the error message

    service "skosmos" can't be used with `extends` as it declare `depends_on`

This was reported as a bug in docker compose in https://github.com/docker/compose/issues/11544 but swiftly closed as apparently this kind of configuration was never allowed by the [docker compose specification](https://github.com/compose-spec/compose-spec/blob/master/05-services.md#restrictions).

This PR "solves" the problem by dropping the `depends_on` declarations from the base `skosmos` service. It's not ideal, because the dependencies are real (Skosmos doesn't work without Fuseki), but a typical `docker compose up` command will start up all three containers in any case and the Skosmos container doesn't really suffer badly from not having access to Fuseki right after startup, so the order of starting up the containers (which is affected by depends_on) doesn't matter much.

Make it possible to start up the test containers for running the test suite.

n/a

- drop the `depends_on` declarations from the base service `skosmos` in top level `docker-compose.yml`

This isn't ideal, but I can't think of a better solution right now.

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -40,9 +40,6 @@ services:
       dockerfile: dockerfiles/Dockerfile.ubuntu
     ports:
       - ${SKOSMOS_PORT:-9090}:80
-    depends_on:
-      - fuseki
-      - fuseki-cache
     volumes:
       - type: bind
         source: ./dockerfiles/config/config-docker-compose.ttl


miguelvaara (approved):
I ran the _./init_containers.sh_ script and all containers started correctly and the error "_service "skosmos" can't be used with extends as it declare depends_on_" no longer occurred. Additionally, Cypress tests worked correctly (_npx cypress open_), so the background system required for the tests also works well.

We can proceed with the merge.

Thank you, @osma!

osma:
The Cypress test for concept page mappings is failing under CI. But that seems unrelated to the docker compose problem that this PR is targeting, so I will merge anyway.

PR #1600, opened 2024-03-05
Scheduled Actions CI run every Monday morning

osma:

We want to catch problems with the test suite early. Recently, our test container setup has broken twice due to changes in docker compose. This PR introduces a scheduled CI run every Monday morning at 6am UTC, regardless of whether any changes have been made in the repository. It should catch situations where the test suite is breaking due to changes in the surrounding ecosystem.

- follow-up enhancement after #1599 

* define a scheduled CI run every Monday morning at 6am UTC

It's only once per week, so could miss some events. We can change the frequency later if this turns out to be a problem. Running too frequently could create noise as well.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,12 @@
 name: CI
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+  schedule:
+    # run the CI workflow at 6:00am UTC every Monday
+    - cron: '0 6 * * MON'
 
 jobs:
   build-test:


miguelvaara (commented):
I was going to setting up GitHub to run CI actions in this feature branch as well and I planned to schedule a test run. However, I then realized that I may have gone too far in my plans.

I have checked the syntax correctness from multiple sources and it seems correct. E.g. [https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions)

However, what I think we could do is to schedule a test run for tomorrow morning and merge into the main branch. That way we can make sure in advance that the desired automation works correctly. Then we would also have time to make changes if the run does not work as expected.

I think there is no a need for documentation, because the ci.yaml documents itself in this regard.

osma:
Removed the back-and-forth commits by force pushing.

osma:
Merging this as is, as we will anyway find out next Monday whether this will work or not.

PR #1601, opened 2024-03-05
Order mappings on concept page first by source vocabulary, then by label

osma:

Cypress tests for the mappings on the concept page were failing, because the two mappings on the example page to "musicology" from Wikidata and "Musicology" from LCSH were not in the expected order. It turned out that the order of mappings depended on the labels only and wasn't stable in a case like this were the labels are the same (case insensitive).

This PR causes the mappings on the concept page (of each type) to be displayed in a consistent order: first by source vocabulary, then by label. This is implemented on the backend side, so that the data returned by the `mappings` REST API method is sorted. (An alternative approach would have been to do the sorting in the frontend only.)

Result can be seen e.g. on the YSO concept page for "musicology":

image

Note that the closeMatch mappings are ordered by vocabulary (LCSH < Wikidata) and likewise the exactMatch mappings are now ordered by vocabulary (Allärs < KOKO < YSA). Before this PR, the order could be different as it was based only on the labels (e.g. music research < musiikintutkimus < musikforskning) and could be arbitrary when labels are the same (musicology vs Musicology).

- part of #1484

* add a getSortKey method to the classes ConceptPropertyValue, ConceptPropertyValueLiteral and ConceptMappingPropertyValue (for the two former, based on label only; for the mappings, based first on vocabulary name, then by label)
* use the getSortKey method when sorting property values in ConceptProperty.sortValues()
* add PHPUnit tests for the getSortKey methods
* adjust the Cypress test for the concept page mappings so that it matches the new order of mapping values

n/a

## Checklist

### DIFF
--- a/src/model/ConceptMappingPropertyValue.php
+++ b/src/model/ConceptMappingPropertyValue.php
@@ -55,6 +55,11 @@ class ConceptMappingPropertyValue extends VocabularyDataObject
         return $label;
     }
 
+    public function getSortKey()
+    {
+        return strtolower($this->getVocabName() . ": " . $this->getLabel());
+    }
+
     private function queryLabel($lang = '', $queryExVocabs = true)
     {
         if ($this->clang) {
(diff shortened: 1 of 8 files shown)


joelit (approved):
LGTM

Issue #1563, opened 2023-11-23
Implement aphabetical view in sidebar using Vue


UnniKohonen:
## Description of the enhancement

Adding a Vue component for alphabetical index in the sidebar.

## What new functionalities would the enhancement make possible?

Most functionality in alphabetical index is complete but some features are still missing (more can be added as they come up):

- [ ] 1. CSS
  - [ ] a. Highlighting currently open concept (removing `#tab-hierarchy` from [this line](https://github.com/NatLibFi/Skosmos/blob/82873ab992ba7608c2cb81cee285485df60d42ad/resource/css/skosmos.css#L535) should fix it)
  - [ ] b. Correct sidebar height
- [ ] 2. Loading new concepts as list is scrolled

PR #1603, opened 2024-03-06
Fix height of sidebar

UnniKohonen:

Sidebar height is not set correctly, this PR fixes the height for alphabetical and hierarchical views.

- Part of #1521 and #1563

- Correct height for alphabetical view
- Correct height for hierarchical view

This PR addresses requirements 3a in #1521 and 1b in #1563

Setting the height is done with JS in Vue components and needs to be done separately for all tabs. This might cause problems later in e.g. mobile layout.

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -423,7 +423,7 @@ body {
   #main-container.vocabpage #sidebar-col, #main-container.termpage #sidebar-col {
     background-color: var(--vocab-bg);
     background-clip: content-box, padding-box;
-    min-height: 1000px; /* What should this be? */
+    min-height: 1000px;
   }
 
   #main-container.vocabpage #sidebar, #main-container.termpage #sidebar {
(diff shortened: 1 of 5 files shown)


resource/js/tab-hierarchy.js
```diff
@@ -184,6 +194,7 @@ tabHierApp.component('tab-hier', {
       this.$emit('loadChildren', concept)
     },
     handleClickConceptEvent (event, concept) {
+      console.log(concept)
```
joelit, line 197: These should be removed eventually

resource/css/skosmos.css
```diff
@@ -423,6 +423,7 @@ body {
   #main-container.vocabpage #sidebar-col, #main-container.termpage #sidebar-col {
     background-color: var(--vocab-bg);
     background-clip: content-box, padding-box;
+    min-height: 1000px;
```
joelit, line 426: It looks OK to me (even though the concept card can be quite a bit smaller).
Screenshot from 2024-03-07 13-23-16

joelit (approved):
LGTM, I'd maybe start cleaning console.log lines.

PR #1605, opened 2024-03-07
Add Cypress tests for partial page load

### LINKED ISSUE #1598: Cypress tests for partial page load

Add Cypress tests for the already existing partial page load functionality. The sidebar components (alphabetical, hierarchy) already support this (clicking on concepts in either of these performs a partial page load of the concept information) but current Cypress test do not verify that all functionality is working after partial page loads.

Initial TODO:

- [x] Ensure that there are Cypress tests for partial page loads in all situations where this happens (all page types that show a sidebar)
- [x] The old draft PR #1474 by @UnniKohonen contains proposed Cypress tests for partial page load; see if it can be reused, or close it if it cannot.
- [x] Adjust the Cypress tests for the concept page; make sure they verify functionality both with and without partial page loads (perhaps using [dynamic tests](https://learn.cypress.io/advanced-cypress-concepts/using-data-to-build-dynamic-tests); each test should be executed twice - with and without partial page load)

Skosmos developers. It's easy to forget testing after partial page loads.

Ensure that functionality is not broken by accident. 

Breakage after partial page loads has happened several times in recent development sprints, so the problem is urgent.

osma:

This PR adds more Cypress tests for verifying functionality during and after partial page loads.

- Closes #1598

- modify the Cypress tests for the concept page so that the most important tests related to dynamic functionality (title, URI, copy to clipboard, mappings) are run both with full and partial page loads
- change the declaration of the SKOSMOS global object from `const SKOSMOS` to `window.SKOSMOS`, and all references to it to `window.SKOSMOS` (this makes it accessible on the Cypress side)
 - add tests to the vocab-home cypress test suite for verifying partial page load functionality when clicking concepts in the sidebar (hierarchy and alphabetical index); also verify that `window.SKOSMOS` has been properly updated after partial page load

Cypress/Electron appears to have problems getting stuck with `cy.visit(url)` in case the URL has already been loaded by the tested application using AJAX style requests. The code in the PR avoids this situation by using different example concept pages for the partial page load tests than for the subsequent tests that perform full page loads. See [this comment](https://github.com/cypress-io/cypress/issues/27501#issuecomment-1980986170) for some details.

## Checklist

### DIFF
--- a/tests/cypress/template/vocab-home.cy.js
+++ b/tests/cypress/template/vocab-home.cy.js
@@ -54,14 +54,11 @@ describe('Vocabulary home page', () => {
   it('clicking on alphabetical index entries performs partial page load', () => {
     cy.visit('/yso/en/') // go to the YSO home page in English language
 
-    // click on the third letter (C)
-    cy.get('#tab-alphabetical .pagination :nth-child(3) > .page-link').click()
+    // click on the the letter C
+    cy.get('#tab-alphabetical').contains('a', 'C').click()
 
-    // check that the first entry is "care institutions"
-    cy.get('#tab-alphabetical .sidebar-list').children().first().invoke('text').should('equal', 'care institutions')
-
-    // click on the first entry (should trigger partial page load)
-    cy.get('#tab-alphabetical .sidebar-list').children().first().click()
+    // click on the link "care institutions" (should trigger partial page load)
+    cy.get('#tab-alphabetical').contains('a', 'care institutions').click()
 
     // check the concept prefLabel
     cy.get('#concept-heading h1').invoke('text').should('equal', 'care institutions')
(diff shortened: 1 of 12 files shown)


tests/cypress/template/vocab-home.cy.js
```diff
@@ -50,4 +50,41 @@ describe('Vocabulary home page', () => {
 
     // check that the first entry is "östliga handelsvägar"
     cy.get('#tab-alphabetical .sidebar-list .list-group').children().first().children().first().invoke('text').should('equal', 'östliga handelsvägar')
-  })})
+  })
+  it('clicking on alphabetical index entries performs partial page load', () => {
```
UnniKohonen, line 54: Should this check mappings too?

tests/cypress/template/vocab-home.cy.js
```diff
@@ -50,4 +50,41 @@ describe('Vocabulary home page', () => {
...
+    // check the concept prefLabel
+    cy.get('#concept-heading h1').invoke('text').should('equal', 'care institutions')
+
+    // check that the SKOSMOS object matches the newly loaded concept
+    cy.window().then((win) => {
+      expect(win.SKOSMOS.uri).to.equal('http://www.yso.fi/onto/yso/p6034');
+      expect(win.SKOSMOS.prefLabels[0]['label']).to.equal("care institutions");
+    })
+  })
+  it('clicking on hierarchy entries performs partial page load', () => {
```
UnniKohonen, line 72: Should this check mappings too?

tests/cypress/template/vocab-home.cy.js
osma: Added check

osma:
@UnniKohonen I added tests for mappings as well. All good now?

UnniKohonen (approved):
LGTM!

PR #1607, opened 2024-03-07
Enhanced Axe configuration to ensure compliance with WCAG 2.1 AA requirements (Issue1606)

### LINKED ISSUE #1606: The accessibility requirements of Axe tests must be explicitly set up in the Axe settings.
Ensure that we are performing accessibility tests at the required level of WCAG 2.1 level AA.

All Skosmos users who require accessible services.

More precisely defined settings

A) Providing accessible services to users is important and B) the organization we operate in requires accessibility from services.

miguelvaara:
To ensure that WCAG 2.1 AA level requirements are met.

- Closes #1606 

Created a JavaScript object that contains Axe's specifications according to WCAG 2.1 AA-level (wcag21aa) and the object is passed as an argument to the Axe functions that require it.

-
## Checklist


tests/cypress/support/accessibility.js
```diff
@@ -1,3 +1,9 @@
+const a11yOptions = {
+  runOnly: {
+    type: 'tag',
+    values: ['wcag21aa']
```
osma, line 4: Axe nowadays supports `wcag22aa` as well, shouldn't we use that?

osma (changes requested):
Looks good. I suggest changing to `wcag22aa` rules unless we have good reason not to.

After considering the above suggestion and implementing it, this can be merged.

The commit history is quite messy, please use the squash-and-merge option.

tests/cypress/support/accessibility.js
miguelvaara: The idea of using WCAG 2.2 rules is good, but if it is not certain whether such a change might cause additional work compared to aiming to the current level of rules, it is better to leave this potentially for the next iteration. The University of Helsinki requires a level of accessibility that meets WCAG 2.1 AA, and the changes made now fulfill this requirement. In summary: a good idea and worth implementing, but better to do it later.

PR #1609, opened 2024-03-13
implement fetchWithAbort and use it for partial page load, mappings & alphabetical index

### LINKED ISSUE #1561: Race conditions in Skosmos 3 AJAX requests

osma:

Trying to address the AJAX race conditions related to partial page load (#1561).

This PR defines a new JS function `fetchWithAbort` that wraps the `fetch` method but also keeps track of pending fetches (per category - there can be different categories) and aborts them early when it's called again.

- Closes #1561

- define function `fetchWithAbort`
- use `fetchWithAbort` when performing partial page load and also when loading mappings for the concept page
- use `fetchWithAbort` when loading entries for the alphabetical index

- The PR doesn't interact with the search field autocomplete widget, which is still WIP in PR #1589. I think the fetchWithAbort function would be useful there as well.
- Should there be Cypress tests? How would they work?

## Checklist

### DIFF
--- a/resource/js/concept-mappings.js
+++ b/resource/js/concept-mappings.js
@@ -12,7 +12,8 @@ const conceptMappingsApp = Vue.createApp({
   },
   methods: {
     loadMappings () {
-      fetchWithAbort('rest/v1/' + window.SKOSMOS.vocab + '/mappings?uri=' + window.SKOSMOS.uri + '&external=true&clang=' + window.SKOSMOS.lang + '&lang=' + window.SKOSMOS.content_lang)
+      const url = 'rest/v1/' + window.SKOSMOS.vocab + '/mappings?uri=' + window.SKOSMOS.uri + '&external=true&clang=' + window.SKOSMOS.lang + '&lang=' + window.SKOSMOS.content_lang
+      fetchWithAbort(url, 'concept')
         .then(data => {
           return data.json()
         })
--- a/resource/js/concept-mappings.js
+++ b/resource/js/concept-mappings.js
@@ -1,4 +1,5 @@
 /* global Vue */
+/* global fetchWithAbort */
 
 const conceptMappingsApp = Vue.createApp({
   data () {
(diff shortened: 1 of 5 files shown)


osma:
You are right @UnniKohonen . I will increase the timeouts. I think it makes sense to do this on this PR branch rather than makkng a separate PR because the timeout problem is closely related even though it's probably not caused by new changes in this PR.

Issue #1611, opened 2024-03-13
Implementation of WCAG rules requires corrections/improvements.


miguelvaara:
## Description of the enhancement

Even though the default level of accessibility rules covered by implemented Cypress Axe tests (en-301-549) aligns with the WCAG 2.1 AA level required by public organizations, it is beneficial to explicitly express these rules in the configuration.

The previous fix (#1607) aimed at making the rules more explicit, does not fully function and the tests are running with too low level rules.

In the initial phase, a basic implementation that "handles" the rules right is sufficient and later the implementation can be improved. The needed fix will likely focus on how the previously defined rules are called. Previously only the presentation of the rules was addressed.

## Who are the users that would benefit from the enhancement and how?
Developers and later the end users of the application.

(Name specific organizations, roles and/or communities if possible, and/or point to relevant discussion elsewhere)

## What new functionalities would the enhancement make possible?
More explicit definition of WCAG rules.

## Why is the enhancement important?
To ensure that the desired level of accessibility is achieved.

(Why should this be prioritized, instead of spending energy on something else? Who would do the work to make this happen?)

PR #1612, opened 2024-03-13
Reverts back to the original implicit usage of WCAG rules and changes the parameterization of CLI and GUI functions to their original state (#1611)

miguelvaara:
To ensure that the required level of accessibility is used in tests

#1611
- Closes #

Only the parameters of the functions _getConfigurationForCLITests_ and _getConfigurationForGUITests_ were modified. Both functions implicitly use, among others level EN 301 549, which corresponds to the officially required WCAG 2.1 AA level by public organizations (https://www.deque.com/en-301-549-compliance/).

The achieved WCAG levels (verified from the information reported in the test browser console) are:

cat.aria
wcag2a
wcag131
EN-301-549
EN-9.1.3.1

Guidance for the reviewer:

Copy the following code into the axe-vocab-home.cy.js file and run the tests normally. For example in the browser's inspector, you should now see a table containing accessibility issues and the test itself should report any problems.

```
import { accessibilityTestRunner } from '../support/accessibility.js'
import 'cypress-axe';

/* If you want the test to be skipped, add a skip command after the describe part:
    - test enabled: describe('Check accessibility of ...
    - test to be skipped: describe.skip('Check accessibility of ... */
describe('Check accessibility of the vocab-home page', () => {
  before(() => {
    cy.visit('/yso/fi/')
    cy.injectAxe()
  })

  accessibilityTestRunner()
})
```

The implementation is not very sophisticated, but it works. However, later we need to focus on this case and make a better implementation. For now, the most important thing is that the checks work fine.

There are no direct issues, but the use of explicitly defined rules is not working, even though the overall functionality works (the test setup accurately identifies errors).

## Checklist

### DIFF
--- a/tests/cypress/support/accessibility.js
+++ b/tests/cypress/support/accessibility.js
@@ -1,7 +1,7 @@
 const a11yOptions = {
   runOnly: {
     type: 'tag',
-    values: ['wcag21aa']
+    values: ['wcag21aa'] // WCAG 2.1 Level AA
   }
 }
 const checkA11y = options => {
@@ -38,11 +38,11 @@ export function accessibilityTestRunner () {
 }
 function getConfigurationForCLITests () {
   return it('Logs (CLI)', () => {
-    cy.checkA11y(null, a11yOptions, terminalLog)
+    cy.checkA11y(null, null, terminalLog)
   })
 }
 function getConfigurationForGUITests () {
   return it('Check for possible accessibility errors at all logging levels set below in accordance with WCAG AA requirements', () => {
-    checkA11y(a11yOptions)
+    checkA11y(null, a11yOptions)
   })
 }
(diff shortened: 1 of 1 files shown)


tests/cypress/support/accessibility.js
```diff
@@ -1,7 +1,7 @@
 const a11yOptions = {
```
osma, line 1: It's pointless and potentially confusing to define an a11Options object that is never used. I suggest deleting it instead if we're not going to use it anyway.

osma: My bad, misread the code.

osma (approved):
Changes look good. Please change the PR title to "revert to default Axe rules" or something along those lines, because that's what this PR actually does.

miguelvaara:
Thank you for the review :+1:

PR #1613, opened 2024-03-13
Concept page mappings: style and visibility fixes

osma:

The layout and styling of concept mappings was wrong after opening the vocabulary home page and clicking on links that perform a partial page load. Also, when switching between concepts, the mappings didn't update immediately, so the wrong (stale) set of mappings could be shown.

This PR addresses the above problems by adjusting the template structure of the concept-mappings Vue component and making the display of the `div` element with the actual mappings conditional on the existence of mappings.

- part of #1484

- make the display of the mappings `div` conditional on the existence of mappings (using a computed value for clarity)
- when switching between concepts, clear the old mappings before starting to load new ones so that stale information is not displayed
- adjust CSS classes of the concept-mappings component

- This PR is built on top of PR #1609 and includes the commits from that PR, so that should be merged first.
- Should there be Cypress tests specific to these changes or are the existing tests enough?
- There is now an extra useless `div` in the DOM structure. I couldn't figure out how to avoid it. I tried changing it to `template` instead in the Vue template, but it just ended up in the DOM as `<template>` and broke the functionality.

image

## Checklist

### DIFF
--- a/resource/js/concept-mappings.js
+++ b/resource/js/concept-mappings.js
@@ -4,7 +4,7 @@
 const conceptMappingsApp = Vue.createApp({
   data () {
     return {
-      mappings: []
+      mappings: {}
     }
   },
   provide: {
@@ -17,7 +17,7 @@ const conceptMappingsApp = Vue.createApp({
   },
   methods: {
     loadMappings () {
-      this.mappings = [] // clear mappings before starting to load new ones
+      this.mappings = {} // clear mappings before starting to load new ones
       const url = 'rest/v1/' + window.SKOSMOS.vocab + '/mappings?uri=' + window.SKOSMOS.uri + '&external=true&clang=' + window.SKOSMOS.lang + '&lang=' + window.SKOSMOS.content_lang
       fetchWithAbort(url, 'concept')
         .then(data => {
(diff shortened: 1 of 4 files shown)


resource/js/concept-mappings.js
```diff
@@ -9,15 +10,29 @@ const conceptMappingsApp = Vue.createApp({
...
   },
+  computed: {
+    hasMappings () {
+      return Object.keys(this.mappings).length > 0
+    }
+  },
   methods: {
     loadMappings () {
-      fetch('rest/v1/' + window.SKOSMOS.vocab + '/mappings?uri=' + window.SKOSMOS.uri + '&external=true&clang=' + window.SKOSMOS.lang + '&lang=' + window.SKOSMOS.content_lang)
+      this.mappings = [] // clear mappings before starting to load new ones
```
UnniKohonen, line 20: Mappings variable should be an object, same on line 7.

UnniKohonen (approved):
Looks good to me! Some kind of indication that the mappings are being loaded could be added but not necessarily in this PR.

There is also a very minor issue with the `mappings` variable but it doesn't affect functionality.

osma:
Fixed the mappings initialization and added the spinner to the TODO list in the conccept page issue. Merging this now.

PR #1614, opened 2024-03-14
Copy to clipboard button for notations

joelit:

This PR replicates Skosmos 2 -style function of using the copy to clipboard button in the title to copy the notation code if one exists.

#1484 

Accordding to Skos spec, [a concept can have several notations](https://www.w3.org/TR/skos-reference/#L2637). We don't have such a vocabulary, but in such case there is no spec for Skosmos to follow - which notation(s) are shown in the title, and what should the copy to clipboard buttom return in such case.

## Checklist

- [X] phpUnit tests pass locally with my changes
- [X] I have added tests that show that the new code works, or tests are not relevant for this PR (e.g. only HTML/CSS changes)
- [X] The PR doesn't reduce accessibility of the front-end code (e.g. tab focus, scaling to different resolutions, use of `.sr-only` class, color contrast)
- [X] The PR doesn't introduce unintended code changes (e.g. empty lines or useless reindentation)

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -232,7 +232,7 @@ body {
     height: 100%;
   }
 
-  #headerbar .dropdown-toggle {
+  #search-wrapper .dropdown-toggle {
     background-color: var(--search-bg);
     color: var(--headerbar-text-2);
     border: 1px solid var(--search-border);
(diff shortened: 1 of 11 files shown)


joelit:
Rebased and force-pushed

osma (changes requested):
The new functionality works very well and the implementation looks good.

The Cypress test "concept notation can be copied to clipboard / partial" failed for me once when I ran it with `npx cypress open`. When comparing notation codes, there was a mismatch: `33.10` != `33.2`. But when running the test again a few times it worked fine. I suspect there may be some kind of race condition within the test: it might copy the notation code before the partial load has finished. For now we can just keep an eye on this but if it happens again, the test should be fixed.

Regarding the visual layout: The notation code is now displayed using the same font and color as the prefLabel. In previous Skosmos/Finto versions, we have used a different color for the notation code:

image

It looks like we don't have a visual spec for how the notation code should look next to the prefLabel. I suggest adding this as a question to the visual errata document: should the notation code look different from the prefLabel next to it?

Also, I think it might be better to place the copy button right next to the notation code instead of placing it to the right of the prefLabel/heading. When class headings are long, the button can be very far away from the code it is copying:

image

But these questions about visual layout are not a blocker for merging this PR.

Regarding this:
> Accordding to Skos spec, [a concept can have several notations](https://www.w3.org/TR/skos-reference/#L2637). We don't have such a vocabulary, but in such case there is no spec for Skosmos to follow - which notation(s) are shown in the title, and what should the copy to clipboard buttom return in such case.

I suggest opening a separate issue about these concerns.

I gave a few comments about small details in the code that could be fixed before merging.

PR #1623, opened 2024-04-18
Search result tweaks for autocomplete

joelit:

- #1514 

* Filtered duplicate results with `unique=true` parameter. This changes the amount of search results recieved
* Alternate labels are displayed in italics
* Notation codes are shown if showNotation configuration parameter is set for the vocabulary
* Notation code results are returned and bolded if they match the search term

The language for the search is not set yet. That fuctionality should come in the follow-up pull request.

## Checklist

### DIFF
--- a/resource/js/vocab-search.js
+++ b/resource/js/vocab-search.js
@@ -10,7 +10,8 @@ const vocabSearch = Vue.createApp({
       renderedResultsList: [],
       languageStrings: null,
       msgs: null,
-      showDropdown: false
+      showDropdown: false,
+      showNotation: null
     }
   },
   mounted () {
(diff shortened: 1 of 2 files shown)


resource/js/vocab-search.js
```diff
@@ -96,9 +103,15 @@ const vocabSearch = Vue.createApp({
...
           result.hit = this.renderMatchingPart(renderedSearchTerm, result.altLabel)
           result.hitPref = this.renderMatchingPart(renderedSearchTerm, result.prefLabel)
-        } else if ('prefLabel' in result) {
-          result.hitType = 'pref'
-          result.hit = this.renderMatchingPart(renderedSearchTerm, result.prefLabel)
+        } else {
+          if (this.notationMatches(renderedSearchTerm, result.notation)) {
+            const notationMatch = this.renderMatchingPart(renderedSearchTerm, result.notation)
+            result.hitType = 'notation'
+            result.hit = notationMatch
```
UnniKohonen, line 110:
```suggestion
            result.hit = this.renderMatchingPart(renderedSearchTerm, result.notation)
```
Hit could be set like with other result types

UnniKohonen (approved):
Everything seems to work as intended. Added a very minor suggestion to code but this can be merged IMO.

resource/js/vocab-search.js
joelit: Ah, that was a left over that I forgot to clean up, from having a buggy way of checking whether the hit was on notation or on preflabel.

Issue #1525, opened 2023-09-27
Switch to JSON message format for UI translations


osma:
## Description of the enhancement

We are currently using traditional gettext `.po` files for storing translated messages. For Skosmos 3, it would make sense to switch to JSON as the message file format because the same files could then be used both by the backend and the frontend.

The Skosmos 3 backend now uses [Symfony Translations](https://symfony.com/doc/current/translation.html) for handling translated messages in PHP code and Twig templates. For the frontend components implemented using Vue there is currently no mechanism for UI translations but we need one - see #1523.

Symfony Translations supports many message formats, among them a very simple JSON key-value format - see [JSON loader implementation](https://github.com/symfony/translation/blob/6.3/Loader/JsonFileLoader.php) and [example message file from the test suite](https://github.com/symfony/translation/blob/6.3/Tests/fixtures/resources.json). The translations for each language has to be stored in a separate JSON file.

Vue i18n, the most likely candidate for handling the translations in Vue components, reads the translated messages from a JavaScript object usually called `messages` which contains the translations for all languages - see [example from the new Finto AI](https://github.com/NatLibFi/FintoAI/blob/visual-update-2023-vuejs/ai.finto.fi/static/js/i18n.js).

If we stored the messages in simple JSON files that are accessible both to the backend and frontend, the exact same files could be used both by Symfony Translations and (after merging the separate languages into a single object) Vue i18n. There would be no need to pass the necessary translations from the backend to the frontend.

What has to be done to make this happen:

* convert `resource/translations/*.po` to JSON format
* switch Symfony Translations so it uses JsonFileLoader

A possible complication is how this would affect Transifex. I haven't checked whether it's able to read and write JSON files. (We could also consider switching to another translation platform - Symfony [supports](https://symfony.com/doc/current/translation.html#translation-providers) Crowdin, Loco and Localise - but that's beyond the scope of this issue).

## Who are the users that would benefit from the enhancement and how?

Skosmos developers; indirectly all users as well

## What new functionalities would the enhancement make possible?

It would simplify implementing Vue component UI translations (#1523)

## Why is the enhancement important?

I think this is a good moment to switch formats. There is almost nothing that requires us to use .po files anymore, now that we've switched away from traditional gettext.

osma:
> (We could also consider switching to another translation platform - Symfony [supports](https://symfony.com/doc/current/translation.html#translation-providers) Crowdin, Loco and Localise - but that's beyond the scope of this issue).

[...]

Issue #1523, opened 2023-09-26
Translate messages shown by Vue components


osma:
For Skosmos 3, we have already implemented UI translations for PHP backend code (PR #1466) and Twig templates (PR #1520). But text in Vue components is not yet translated to the UI language, as can be seen for example in the vocab-counts and term-counts components on the vocabulary home page (the texts "Resource counts by type", "Term counts by language" etc. are currently hardcoded and always shown in English):

![image](https://github.com/NatLibFi/Skosmos/assets/1132830/20609617-ce1e-4d64-aa93-4d486d9c19a2)

To fix this, we need to add internationalization support to the Vue components, most likely using [Vue i18n](https://kazupon.github.io/vue-i18n/).

We also need to provide the Vue components with the correct messages for each language. For Skosmos 2, this was done using global JS variables, like this:

```js
<!-- translations needed in javascript -->
var noResultsTranslation = "Ei tuloksia";
var loading_text = "Ladataan sisältöä";
var loading_failed_text = "Virhe: Lisätulosten haku epäonnistui!";
var loading_retry_text = "Yritä uudelleen";
var jstree_loading = "Ladataan ...";
var results_disp = "Näytetään kaikki %d tulosta";
var all_vocabs  = "kaikista sanastoista";
var n_selected = "sanastosta";
var missing_value = "Kenttä on pakollinen";
var expand_paths = "näytä kaikki # polkua";
var expand_propvals = "näytä kaikki # arvoa";
var hiertrans = "Hierarkia";
var depr_trans = "Käytöstä poistettu käsite";
var sr_only_translations = {
  hierarchy_listing: "Hierarkian mukainen lista käsitteistä",
  groups_listing: "Hierarkian mukainen lista käsitteistä ja ryhmistä",
};
```

A similar but hopefully less ugly mechanism for passing the messages from the backend to the frontend is needed. One option would be to add the necessary messages to the SKOSMOS global object (see #1438 where it was implemented).

osma:
If we switched to a JSON based message format, it would simplify the implementation of Vue translations as there would be no need to pass specific messages from the backend to the frontend. I opened issue #1525 proposing this.

osma:
Issue #1656 is a near-duplicate of this. But we can consider this issue to be the broader task of translations in all Vue components, while #1656 is about providing the framework for doing this.

joelit:
Closing this issue as completed. #1774 #1776 and #1777 implement the latest round of translations to Vue components.

PR #1624, opened 2024-04-18
Add infinite scrolling to alphabetical index

UnniKohonen:

Alphabetical index does not load more concepts as it's scrolled. This PR adds functionality to address this.

- Part of #1563

- Add an event listener for scrolling alphabetical list
- Detect when scrolling reaches the bottom of the list and load more concepts
- Add messages for loading content to `SKOSMOS` variable
- Add spinners when loading content

Addresses requirement 2 in #1563

- There could still be some bugs with loading (i.e. loading happening in the wrong order, concepts being loaded twice, some offsets being skipped, etc).

## Checklist

### DIFF
--- a/resource/js/tab-alpha.js
+++ b/resource/js/tab-alpha.js
@@ -104,7 +104,7 @@ const tabAlphaApp = Vue.createApp({
         })
     },
     handleScrollEvent () {
-      listElement = this.$refs.tabAlpha.$refs.list
+      const listElement = this.$refs.tabAlpha.$refs.list
       if (listElement.scrollTop + listElement.clientHeight >= listElement.scrollHeight - 1) {
         this.loadMoreConcepts()
       }
--- a/resource/js/tab-alpha.js
+++ b/resource/js/tab-alpha.js
@@ -11,6 +11,7 @@ const tabAlphaApp = Vue.createApp({
       loadingLetters: false,
       loadingConcepts: false,
       loadingMoreConcepts: false,
+      loadingMessage: '',
       currentOffset: 0
     }
   },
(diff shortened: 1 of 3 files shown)


resource/js/tab-alpha.js
```diff
@@ -130,13 +177,13 @@ tabAlphaApp.component('tab-alpha', {
...
-      <div>
-        Loading...
-      </div>
-    </template>
-    <template v-else>
-      <div class="sidebar-list" :style="getListStyle()">
+    <div class="sidebar-list" :style="getListStyle()" ref="list">
+      <template v-if="loadingConcepts">
+        <div>
+          Loading...
```
joelit, line 183: This could be a value from the window.SKOSMOS object, e.g. "Ladataan sisältöä", "Laddar mera resultat" & "Loading more items". Also if you want to use a spinner, something like `<i class="fa-solid fa-spinner fa-spin-pulse"></i>` could be considered.

resource/js/tab-alpha.js
```diff
@@ -147,9 +194,12 @@ tabAlphaApp.component('tab-alpha', {
               :href="getConceptURL(concept.uri)" @click="loadConcept($event, concept.uri)"
             >{{ concept.prefLabel }}</a>
           </li>
+          <template v-if="loadingMoreConcepts">
+            Loading...
```
joelit, line 198: Same here as on L183

joelit (changes requested):
It works well now! A minor thing, but I'd get rid of the hardcoded loading message at this point already.

resource/js/tab-alpha.js
osma: Not sure if it's worth adding more translations to the SKOSMOS object, because the translation mechanism for the frontend may still change to something completely different that doesn't rely on the SKOSMOS object (e.g. reading the translations directly from a JSON file). See #1525 and https://github.com/NatLibFi/Skosmos/issues/1523#issuecomment-1736810725 for some thoughts

joelit (approved):
LGTM!

PR #1625, opened 2024-04-18
SEO: improved titles for Skosmos 3 generated HTML pages

### LINKED ISSUE #1533: Search Engine Optimization for Skosmos 3.0
Skosmos pages have generally pretty good visibilty on search engines. However, according to [Finto.fi](http://finto.fi/) use statistics, the number of visit directly from search engine searches has decreased quite a bit in the past few years.

Now that the front end of Skosmos in getting a major overhaul with version 3.0, it might be a good idea to try to optimize the visibility of Skosmos pages to search engines.

The current (Skosmos 2.X) SEO methods include… 

Let’s use this issue to spec out the concrete measures needed for Skosmos 3.0 SEO.

For related issues see at least #583

osma:

Improve search engine optimization in Skosmos 3 by making the `<title>` values and related metadata (`title`, `twitter:title`, `og:title`) conform to the specification in https://github.com/NatLibFi/Skosmos/issues/1533#issuecomment-2063290242 .

This PR only handles the titles, not descriptions. The descriptions will have to be implemented in a subsequent PR.

- Part of #1533

- refactor: pass the GlobalConfig instance to the Twig templates, instead of individual configuration settings in separate variables
- adapt the Twig templates to use the GlobalConfig instance, and set their title metadata to match the specification
- add Cypress tests to verify the title values

- this PR doesn't implement long titles (skosmos:serviceNameLong setting) for the landing page, but just uses regular serviceName for the landing page title; implementing long titles is left to another PR

## Checklist

### DIFF
--- a/src/view/error.twig
+++ b/src/view/error.twig
@@ -1,6 +1,6 @@
 {% set pageType = 'error' %}
 {% extends "base-template.twig" %}
-{% block title %}: "Error"{% endblock %}
+{% block title %}{{ "Error"| trans }} - {{ GlobalConfig.serviceName }}{% endblock %}
 {% block content %}
 {% set pageError = request.vocabid != '' and request.page == 'page' %}
 {% if pageError %}
--- a/resource/translations/skosmos_en.po
+++ b/resource/translations/skosmos_en.po
@@ -882,3 +882,6 @@ msgstr "Breadcrumbs"
 
 msgid "Copy to clipboard"
 msgstr "Copy to clipboard"
+
+msgid "Error"
+msgstr "Error"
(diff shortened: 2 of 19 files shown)


src/view/error.twig
```diff
@@ -1,6 +1,6 @@
 {% set pageType = 'error' %}
 {% extends "base-template.twig" %}
-{% block title %}: "Error"{% endblock %}
+{% block title %}{{ "Error"| trans }} - {{ GlobalConfig.serviceName }}{% endblock %}
```
joelit, line 3: There does not seem to be a translation string for 'Error':
* I tried with pages like https://localhost/Skosmos/yso/fi/p%C3%A5ge/p23406
* Then again, `grep -C 1 'Error' skosmos_fi.po` didn't show a msg-id that would be just 'Error'

The 'error' msgid should probably be added to the translation templates.

osma:
Converting back to draft, because I still need to:
- add the missing "Error" msgid pointed out by @joelit 
- implement long title (skosmos:serviceNameLong setting) for the landing page

joelit (approved):
LGTM!

PR #1626, opened 2024-04-23
Sort concepts in hierarchical view by label and locale

UnniKohonen:

Concepts in hierarchical view are not currently sorted. This PR add functionality to sort concepts by their label according to locale.

- Part of #1521 

Sort concepts in hierarchy by label and locale when they are added to hierarchy tree

Addresses requirement 6d in #1521 

Concepts are only sorted by label and locale, not notation codes or notation sorting style

## Checklist

### DIFF
--- a/resource/js/tab-hierarchy.js
+++ b/resource/js/tab-hierarchy.js
@@ -155,8 +155,9 @@ const tabHierApp = Vue.createApp({
       }
     },
     compareLabels (a, b) {
-      const labelA = a.label ? a.label : a.prefLabel
-      const labelB = b.label ? b.label : b.prefLabel
+      // Set labels as label, prefLabel or an empty string if there are no lables
+      const labelA = a.label ? a.label : (a.prefLabel ? a.prefLabel : "")
+      const labelB = b.label ? b.label : (b.prefLabel ? b.prefLabel : "")
       const lang = window.SKOSMOS.content_lang ? window.SKOSMOS.content_lang : window.SKOSMOS.lang
 
       return labelA.localeCompare(labelB, lang)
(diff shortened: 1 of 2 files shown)


resource/js/tab-hierarchy.js
```diff
@@ -148,6 +153,13 @@ const tabHierApp = Vue.createApp({
         height: 'calc( 100% - ' + height + 'px)',
         width: width + 'px'
       }
+    },
+    compareLabels (a, b) {
+      const labelA = a.label ? a.label : a.prefLabel
+      const labelB = b.label ? b.label : b.prefLabel
+      const lang = window.SKOSMOS.content_lang ? window.SKOSMOS.content_lang : window.SKOSMOS.lang
+
+      return labelA.localeCompare(labelB, lang)
```
osma, line 162: This may fail if labelA happens to be null, which currently happens for example in some of the Cypress tests. I suggest checking for the case where either labelA or labelB is null and using another value instead such as "" (empty string).

tests/cypress/template/vocab-home.cy.js
```diff
@@ -61,7 +61,7 @@ describe('Vocabulary home page', () => {
     cy.get('#tab-alphabetical').contains('a', 'care institutions').click()
 
     // check the concept prefLabel
-    cy.get('#concept-heading h1').invoke('text').should('equal', 'care institutions')
+    cy.get('#concept-heading h1', {'timeout': 15000}).invoke('text').should('equal', 'care institutions')
```
osma, line 64: Is this change necessary?

osma (changes requested):
null labels should be handled, otherwise this looks good!

tests/cypress/template/vocab-home.cy.js
UnniKohonen: It was failing before without the timeout but now it isn't anymore. I'll take it out

osma (approved):
LGTM, please check the SonarCloud suggestions

PR #1628, opened 2024-04-24
Redesign initialization of Model and GlobalConfig to avoid chicken-and-egg situation

### LINKED ISSUE #1627: Initialization problems in Model and GlobalConfig
## Description of the problem

While working on other issues, I noticed that the initialization of the Model and GlobalConfig objects is a bit broken in current Skosmos (both versions 2 and 3). There is a chicken-and-egg problem when creating these fundamental objects that form the core of the application:

- The Model constructor takes a GlobalConfig object as parameter
- GlobalConfig is a subclass of BaseConfig, which is a subclass of DataObject
- The DataObject constructor takes a Model object as parameter, because a DataObject should always know the current Model

So Model cannot be constructed without first creating a GlobalConfig object. But GlobalConfig cannot be constructed without first creating a Model object. Chicken, meet egg!

However, this has still worked because the GlobalConfig and BaseConfig constructors never call the constructor of the DataObject superclass. So the GlobalConfig can in fact be constructed without a Model, but then it will not know the current Model like a proper DataObject should. The end result is that GlobalConfig expects to know the Model but in fact doesn't have a reference to it. This becomes a problem when it actually needs to access Model, for example when it needs to know the current language of the application (model.getLocale()) when reading multilingual values from the Skosmos configuration file.

This can be resolved in multiple ways, but it's clear that the initialization needs to be redesigned. I'm currently working on a fix and will soon open a PR.

osma:

There is a chicken-and-egg problem in initializing the Model and GlobalConfig objects that are fundamental to the Skosmos application (see #1627). This PR changes the way these are initialized to avoid the circular dependency.

Previously, GlobalConfig was created first (with optionally a path to the configuration file); then Model was created, giving it the GlobalConfig object as a constructor parameter. But this meant that GlobalConfig didn't have access to the Model even though it should, since it's a subclass of DataObject and DataObjects are expected to know the current Model. A similar problem affects VocabularyConfig that also needs to be aware of the Model as it is also a DataObject.

This PR changes the situation:

- Model is always created first, and its constructor (which is optionally given a path to the configuration file) creates the GlobalConfig object. This way both objects are aware of each other from the beginning. It also means that they always go hand-in-hand; it's not possible to create one without the other.
- VocabularyConfig is now given the Model as a constructor parameter.

These changes are not very big, but since many PHPUnit tests initialize one or more of these objects, a lot of changes had to be made to the test suite.

Also, the exception handling was a bit broken/misplaced, so this PR changes the way exceptional situations are handled. Exceptions are thrown by GlobalConfig but they are no longer immediately caught within GlobalConfig. Instead the exception is allowed to propagate to the outermost calling code. `index.php` catches and reports these exceptions to the user because this helps to diagnose common problems encountered when Skosmos is being installed. (This matches the old behavior, but the catching code has been moved.)

In the initial version of this PR, I forgot to fix entity.php; it was not covered by any PHPUnit or Cypress tests. So I added a few simple Cypress tests to verify that entity.php is working.

- Closes #1627

* change the constructors of Model, GlobalConfig and VocabularyConfig for the new initialization style
* modify the Skosmos application code to use the new initialization style
* adapt the PHPUnit tests to the changes in the initialization
* add a new unit test VocabularyTest.testGetShortNameLang that verifies that GlobalConfig can access current language information from the Model (this was not possible until this PR)
* add a new Cypress E2E test for serving linked data via entity.php

none

## Checklist

### DIFF
--- a/src/entity.php
+++ b/src/entity.php
@@ -8,8 +8,7 @@ require_once '../vendor/autoload.php';
 header("Access-Control-Allow-Origin: *"); // enable CORS
 
 try {
-    $config = new GlobalConfig();
-    $model = new Model($config);
+    $model = new Model();
     $controller = new EntityController($model);
     $request = new Request($model);
 
--- a/src/index.php
+++ b/src/index.php
@@ -13,8 +13,7 @@ try {
     return;
 }
 
-$config = new GlobalConfig();
-$model = new Model($config);
+$model = new Model();
 $controller = new WebController($model);
 $request = new Request($model);
 
@@ -29,7 +28,7 @@ if (sizeof($parts) <= 2) {
     $lang = sizeof($parts) == 2 && $parts[1] !== '' ? $parts[1] : $controller->guessLanguage($request);
     header("Location: " . $lang . "/");
 } else {
-    if (array_key_exists($parts[1], $config->getLanguages())) { // global pages
+    if (array_key_exists($parts[1], $model->getConfig()->getLanguages())) { // global pages
         $request->setLang($parts[1]);
         $model->setLocale($parts[1]);
         $content_lang = $request->getQueryParam('clang');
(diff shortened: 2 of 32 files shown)


osma:
Good point. I had fixed index.php and rest.php but forgotten entity.php

joelit (approved):
Looks good to me!

PR #1629, opened 2024-04-24
Setting the language in the search bar

joelit:

#1514 

- Rendering the search results when "All languages" is chosen
- Choosing the correct search language
  - First from the url clang parameter
  - Second from cookie
  - If all else fails, from the SKOSMOS object

Needs more cypress tests

## Checklist

### DIFF
--- a/resource/js/vocab-search.js
+++ b/resource/js/vocab-search.js
@@ -16,7 +16,7 @@ const vocabSearch = Vue.createApp({
   },
   mounted () {
     this.languages = window.SKOSMOS.languageOrder
-    this.selectedLanguage = window.SKOSMOS.content_lang
+    this.selectedLanguage = this.parseSearchLang()
     this.searchCounter = 0
     this.languageStrings = window.SKOSMOS.language_strings[window.SKOSMOS.lang] ?? window.SKOSMOS.language_strings.en
     this.msgs = window.SKOSMOS.msgs[window.SKOSMOS.lang] ?? window.SKOSMOS.msgs.en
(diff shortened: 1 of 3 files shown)


resource/js/vocab-search.js
```diff
@@ -67,8 +68,39 @@ const vocabSearch = Vue.createApp({
...
       return false
     },
-    renderMatchingPart (searchTerm, label) {
+    parseSearchLang () {
+      // if content language can be found from uri params, use that and update it to SKOSMOS object and to search lang cookie
+      const urlParams = new URLSearchParams(window.location.search)
+      const paramLang = urlParams.get('clang')
+      const anyLang = urlParams.get('anylang')
+      if (anyLang) {
+        this.setSearchLangCookie('all')
```
UnniKohonen, line 77: Maybe this should call `changeLang` like on line 81

resource/js/vocab-search.js
```diff
@@ -163,14 +212,22 @@ const vocabSearch = Vue.createApp({
...
       this.showDropdown = true
       this.$forceUpdate()
+    },
+    setSearchLangCookie (lang) {
+      // The cookie path should be relative if the baseHref is known
+      let cookiePath = '/'
+      if (window.SKOSMOS.baseHref && window.SKOSMOS.baseHref.replace(window.origin, '')) {
+        cookiePath = window.SKOSMOS.baseHref.replace(window.origin, '')
+      }
+      document.cookie = `SKOSMOS_SEARCH_LANG=${this.selectedLanguage};path=${cookiePath}`
```
UnniKohonen, line 222: The value of `this.selectedLanguage` is null when the `anylang` URL parameter is set because `this.selectedLanguage` is updated only after the cookie is set on line 78.

UnniKohonen (changes requested):
When the `anylang` URL parameter is set, the value of the cookie is set to null. This should be fixed.

There is also a cypress test that fails but I don't know if the problem originates from this PR.

Otherwise everything looks good

UnniKohonen (approved):
LGTM!

PR #1638, opened 2024-07-23
Hierarchical view style fixes

UnniKohonen:

Hierarchical view has several minor style errors, this PR fixes them.

- Part of #1521

- Fix width of concepts
- Remove unnecessary borders
- Set `last` class correctly
- Use images instead of unicode characters in buttons

Addresses 3g, 3h and 3c in #1521

## Checklist


osma (changes requested):
I tested this briefly and the changes look very good! But there are currently two problems:

1. There is a conflict that needs to be resolved
2. The Cypress Axe test for the concept page fails for some reason

If you can fix these, then this can be merged!

PR #1639, opened 2024-07-24
Fix sidebar width on window resize

UnniKohonen:

Sidebar Vue components do not currently resize correctly with window resizing, this PR fixes that.

- Part of #1521 and #1563

- Adds an event listener for window resizing in alphabetical and hierarchical components
- Change the width of the components on window resize event

Addresses requirement 3i in #1521 and 1d in #1563

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -554,6 +554,7 @@ body {
   #tab-hierarchy .sidebar-list .list-group-item {
     border-left: 1px dashed var(--vocab-text);
     border-radius: 0;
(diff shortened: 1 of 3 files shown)


joelit (approved):
LGTM!

The resize listener works as expected. However, the layout itself is not the best in every window width - with window size just under 990px, the right mardin disappears and the left margin is set pretty wide, reducing the room for deep hierarchy.
990px

joelit (approved):
Looks good to me!

PR #1644, opened 2024-08-14
SEO: long title on landing page (skosmos:serviceNameLong configuration setting)

osma:

This PR makes it possible to set a long name for the service using a new setting `skosmos:serviceNameLong`. This name will be used as the `<title>` (and related metadata elements important for SEO) on the landing page. The setting is multilingual, so it is possible to set different names for different UI languages. If not set, the value of the `skosmos:serviceName` setting is used as a fallback.

I also adjusted the `BaseConfig.getLiteral()` fallback mechanism in case there is no configuration value in the requested language. In that case it picked an arbitrary value that could be in any other language. I adjusted the mechanism so it only falls back to a value without a language tag, but not to another language which could be surprising.

There's also a small change to drop `twitter:title` metadata because it turns out it's completely redundant according to [Twitter/X developer docs](https://developer.x.com/en/docs/twitter-for-websites/cards/guides/getting-started#opengraph) when we already have `og:title`.

- Part of #1533 

- add support for skosmos:serviceNameSetting in GlobalConfig
- adjust BaseConfig.getLiteral so it uses a better fallback mechanism for multilingual values
- PhpUnit tests for the above
- use the serviceNameLong value in the Twig template of the landing page
- drop use of `twitter:title` because it's redundant
- Cypress test for the above

## Checklist

### DIFF
--- a/src/view/base-template.twig
+++ b/src/view/base-template.twig
@@ -11,7 +11,6 @@
 <meta name="generator" content="Skosmos {{ request.version }}">
 <meta name="title" content="{{ block('title') }}">
 <meta property="og:title" content="{{ block('title') }}">
-<meta name="twitter:title" content="{{ block('title') }}">
 <meta name="twitter:card" content="summary">
 <link href="node_modules/bootstrap/dist/css/bootstrap.min.css" media="screen, print" rel="stylesheet" type="text/css">
 <link href="resource/css/fonts.css" media="screen, print" rel="stylesheet" type="text/css">
--- a/src/model/BaseConfig.php
+++ b/src/model/BaseConfig.php
@@ -54,7 +54,14 @@ abstract class BaseConfig extends DataObject
             return $literal->getValue();
         }
 
-        // not found with selected language, try any language
+        // not found with selected language, try to find one without a language tag
+        foreach ($this->getResource()->allLiterals($property) as $literal) {
+            if ($literal->getLang() === null) {
+                return $literal->getValue();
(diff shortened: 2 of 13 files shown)


src/view/base-template.twig
```diff
@@ -11,7 +11,6 @@
 <meta name="generator" content="Skosmos {{ request.version }}">
 <meta name="title" content="{{ block('title') }}">
 <meta property="og:title" content="{{ block('title') }}">
-<meta name="twitter:title" content="{{ block('title') }}">
 <meta name="twitter:card" content="summary">
```
joelit, line 14: https://developer.x.com/en/docs/twitter-for-websites/cards/overview/markup#overview-of-all-twitter-card-tags

Now that we're getting rid of twitter:title in favour of og:title, we could also change twitter:card to og:type. I reckon the domain of og:type fits skosmos pages better than the domain of twitter:card.

joelit (approved):
Looks good to me!

I would use this PR (or make a new one) to also change twitter:card to og:type to make sure the SE crawler can parse the metadata in the future.

osma:
> I would use this PR (or make a new one) to also change twitter:card to og:type to make sure the SE crawler can parse the metadata in the future.

I think it's out of scope for this PR, which is about titles/names. But I added this to the checklist in the main issue: https://github.com/NatLibFi/Skosmos/issues/1533#issuecomment-2077229334

Merging this as-is.

PR #1645, opened 2024-08-14
Fix sidebar Vue component sizing

UnniKohonen:

Setting the height and width of sidebar Vue components is broken and needs to be fixed.

Set the height and width of sidebar-list on `beforeUpdate` lifecycle hook instead of `mounted`. This sets the components to the correct size.

## Checklist


joelit (approved):
LGTM

Issue #1527, opened 2023-09-27
Switch to a new translation platform (instead of Transifex)


osma:
## Description of the enhancement

We have been using Transifex to manage UI translations maintained by the community. But it has always been a bit difficult to work with and recently the API access has stopped working as well. With Skosmos 3 we have switched to Symfony Translations that [provides direct support](https://symfony.com/doc/current/translation.html#translation-providers) for three translation platforms: Crowdin, Loco (localise.biz) and Lokalise. Obviously we would need to evaluate these and pick one. We could also ask about this on skosmos-users as part of the evaluation.

## Who are the users that would benefit from the enhancement and how?

- Skosmos developers (making it easier to work with contributed translations)
- individuals or organizations that contribute UI translations, e.g. ZBW, ZPID, GBV

## What new functionalities would the enhancement make possible?

It would enable simple pushing and pulling translations from Skosmos to the chosen platform.

## Why is the enhancement important?

Now would be a good time to switch. We can keep using Transifex for Skosmos 2 but switch to something else for Skosmos 3.

miguelvaara:
Are the limits of the free option on Localise sufficient for us?
[https://localise.biz/plans](https://localise.biz/plans)

miguelvaara:
Loco has been contacted and discussions (pricing and technical issues) are ongoing. A more detailed discussion can be found in out team's shared email.

miguelvaara:
## Post-sprint report

In regards to Lokalise, one major question is how much we need to modify and change the Symfony framework, which we already use very minimally. Another important question is whether, as open-source advocates, we have the option to use Lokalise free of charge.

So far, we have explored other translation frameworks, such as Loco and have discussed pricing with them. However, initially we are investigating the possibilities of using Lokalise, especially since Lokalise is also used with VuFind, which is utilized by Finna (https://github.com/NatLibFi/NDL-VuFind2). It is always better if we can find synergies with the Finna team.

**So far, the following has been done:**

A demo account has been set up and,
- the PO translation files for the 21 languages we use have been uploaded to the service (although some languages are still incomplete)
- we have tested the user interface https://app.lokalise.com (in terms of our user community this is important and we will report on this in more detail later).

**The use of the API has been tested, including:**

- Generating the necessary tokens
- Listing all used keys (assets used to fetch translations)
- Retrieving a list of all translation files
- Fetching a link from which the translation files can be downloaded in zip format.

### Method 1 / API:

[...]

PR #1648, opened 2024-08-14
WIP: Use Symfony Translations to synchronize translations with Lokalise

osma:

WIP.

See #1527 

- Closes #

Running PHP-CS-Fixer under GitHub Actions fails on PHP 8.2.

## Checklist


composer.json
```diff
@@ -16,18 +16,29 @@
     "ext-mbstring": "*",
     "ext-xsl": "*",
     "monolog/monolog": "1.27.*",
-    "symfony/translation": "6.0.*",
+    "symfony/translation": "6.4.*",
     "symfony/translation-contracts": "3.0.*",
-    "symfony/twig-bundle": "6.0.*",
-    "symfony/config": "6.0.*"
+    "symfony/twig-bundle": "6.4.*",
+    "symfony/config": "6.4.*",
```
osma, line 22: The new symfony dependencies are only needed to get the console working, so they could perhaps be moved into dev dependencies

config/packages/twig.yaml
```diff
@@ -0,0 +1,4 @@
+twig:
+    default_path: '%kernel.project_dir%/src/view'
+    file_name_pattern: ['*.twig', '*.inc']
```
osma, line 3: Could be dropped, we no longer use .inc

osma:
TODO:

- [ ] change domain to "messages" in Lokalise so that pull works
- [ ] document the push/pull commands, maybe write a wrapper script?

osma:
It was pretty difficult to properly initialize the project on the Lokalise side so that it works with Symfony Translate push and pull commands. One crucial thing is that all the translation keys need to have the filename `messages.xliff` on the Lokalise side (which is completely separate from the filenames used on the application side, which are `messages.XX.json`). The push and pull commands assume this filename and will not work if it is something else. Also, the push command is very unreliable and breaks easily. Here is a procedure that seems to work:

1. Initialize the Lokalise project, or delete all keys, so that it has no keys
2. Push the English language translations using this command: `bin/console translation:push lokalise --domains messages --locales en --force`. This will define the necessary keys and they will all have the filename `messages.xliff` as required later.
3. Use the Upload function in Lokalise to upload all the other translation files. (This will take a while to process, check the Activity view)
4. Make sure that all keys have been assigned the filename `messages.xliff` (some keys that were not in the English language file could have another filename). In the Multilingual view, look for keys that have been assigned to a filename other than `messages.xliff` using a Filter. Then click on each key to open the details and on the Advanced tab, change the filename to `messages.xliff` under Assigned to file. You may have to refresh a few times to change them all...

Now the project should have all keys and all available translations on the Lokalise side. After this point, only English language translations need to be pushed (e.g. after extracting new keys from templates). Pulling is more reliable and should be done for all languages regularly.

Push command for English:

    bin/console translation:push lokalise --domains messages --locales en --force

Pull command for a single language (the format parameter is important!): 

    bin/console translation:pull lokalise --domains messages --format json --locales fi

Pull command for all languages: 

    bin/console translation:pull lokalise --domains messages --format json

_NOTE_: Pulling the English language translations causes all entries that have an "Empty" translation on the Lokalise side (i.e. the message is identical to the key, or used to be) to be dropped from the file entirely.

osma:
Steps for extracting new message ids, adding them to Lokalise, translating them to other languages, then pulling back:

1. Make a change in the code that adds a new translated message into the twig templates
2. Check that it can be extracted: `bin/console translation:extract --dump-messages en` (should be shown in the list) or `bin/console debug:translation en` (should show it as missing)
3. Extract the message(s) and add them to the English language translation file: `bin/console translation:extract --force en --format json --prefix ""`
4. Push the translation file to Lokalise: `bin/console translation:push lokalise --domains messages --locales en --force`
5. In the Lokalise UI, look up the message and add translations in other languages
6. Pull the translations back to the codebase: `bin/console translation:pull lokalise --domains messages --format json`

osma:
Preliminary documentation: https://github.com/NatLibFi/Skosmos/wiki/Translation-for-Skosmos-3

miguelvaara (approved):
At first, there was some uncertainty about whether the system was actually updating the data correctly (deletion in Lokalise -> deletion locally as well), but it seems that it worked correctly after all. Although the feature branch can be merged, further testing should still be done just to be sure.

PR #1650, opened 2024-08-15
Bootstrap search bar

joelit:

- Related to #1514

- [x] Bootstrap dropdown list and JS functions for search language selection
- [x] Updated cypress tests
- [x] CSS for the new implementation
- [x] Accessibility tests for the new implementation

## Checklist

### DIFF
--- a/resource/js/vocab-search.js
+++ b/resource/js/vocab-search.js
@@ -245,7 +245,7 @@ const vocabSearch = Vue.createApp({
               @click="changeContentLangAndReload(key)"
               @keydown.enter="changeContentLangAndReload(key)"
               role="menuitem"
-              tabindex=1 >
+              tabindex=0 >
               {{ value }}
             </a>
           </li>
--- a/resource/js/vocab-search.js
+++ b/resource/js/vocab-search.js
@@ -227,7 +227,7 @@ const vocabSearch = Vue.createApp({
       <div class="d-flex justify-content-end input-group ms-auto" id="search-wrapper">
 
       <div class="dropdown" id="language-selector">
-        <a class="btn btn-outline-secondary dropdown-toggle"
+        <button class="btn btn-outline-secondary dropdown-toggle"
            href="#"
            role="button"
            data-bs-toggle="dropdown"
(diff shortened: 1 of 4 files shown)


resource/js/vocab-search.js
```diff
@@ -225,12 +225,26 @@ const vocabSearch = Vue.createApp({
...
       <div class="d-flex justify-content-end input-group ms-auto" id="search-wrapper">
-        <select class="btn btn-outline-secondary dropdown-toggle" data-bs-toggle="dropdown-item"
-          v-model="selectedLanguage"
-          @change="changeContentLangAndReload($event.target.value)"
-          aria-label="Select search language">
-          <option class="dropdown-item" v-for="(value, key) in languageStrings" :value="key">{{ value }}</option>
-        </select>
+
+      <div class="dropdown" id="language-selector">
+        <a class="btn btn-outline-secondary dropdown-toggle"
```
UnniKohonen, line 230: Should this be a button like here? https://getbootstrap.com/docs/5.2/components/dropdowns/#single-button

Using a button seems to at least fix the size of the dropdown toggle

UnniKohonen (commented):
Looks good so far. The only thing I would maybe change is using a `<button>` element instead of a `<a>` element in dropdown toggle.

CSS changes are still missing but I noticed some things that need to be fixed at least:
- color of the dropdown menu text should be dark blue
- hover highlight should be added to dropdown menu items
- cursor should probably be something different in dropdown menu
- text disappears when clicking on dropdown menu items
- the arrow on the button should be changed

resource/js/vocab-search.js
joelit: Button is fine too. It can be done with either - the next example on the linked document is with <a> -elements. I'll have a look at how the choise of the element affects accessibiliy, mobile user agents etc.

resource/js/vocab-search.js
```diff
@@ -225,12 +225,33 @@ const vocabSearch = Vue.createApp({
...
-        <select class="btn btn-outline-secondary dropdown-toggle" data-bs-toggle="dropdown-item"
-          v-model="selectedLanguage"
-          @change="changeContentLangAndReload($event.target.value)"
-          aria-label="Select search language">
-          <option class="dropdown-item" v-for="(value, key) in languageStrings" :value="key">{{ value }}</option>
-        </select>
+
+      <div class="dropdown" id="language-selector">
+        <button class="btn btn-outline-secondary dropdown-toggle"
+           href="#"
```
osma, line 231: I don't think the `href` attribute is useful for a `button` element?

osma (approved):
I tested this and it looks good. I found one minor issue in the HTML template: the `button` element probably shouldn't have a `href` attribute.

I noticed that one Cypress test failed during CI, but the failing test was unrelated to this PR. In fact, I just fixed those flaky tests today in PR #1652 which is already merged so all should be well.

resource/js/vocab-search.js
joelit: Missed that from the a-element style implementation. I'll fix it and merge.

PR #1651, opened 2024-08-15
Order hierarchy by notation

UnniKohonen:

Hierarchy is currently only ordered by labels. It should be possible to order it by notation codes as well

- Part of #1521 

Sort concepts based on notation codes using both lexical and natural sorting.

Addresses requirement 6d in #1521

## Checklist

### DIFF
--- a/resource/js/tab-hierarchy.js
+++ b/resource/js/tab-hierarchy.js
@@ -52,7 +52,7 @@ const tabHierApp = Vue.createApp({
 
           this.hierarchy = []
 
-          for (const c of data.topconcepts.sort((a, b) => this.compareLabels(a, b))) {
+          for (const c of data.topconcepts.sort((a, b) => this.compareStrs(a, b))) {
             this.hierarchy.push({ uri: c.uri, label: c.label, hasChildren: c.hasChildren, children: [], isOpen: false, notation: c.notation })
           }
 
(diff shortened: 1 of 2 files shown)


resource/js/tab-hierarchy.js
```diff
@@ -161,13 +161,35 @@ const tabHierApp = Vue.createApp({
         width: width + 'px'
       }
     },
-    compareLabels (a, b) {
-      // Set labels as label, prefLabel or an empty string if there are no lables
-      const labelA = a.label || a.prefLabel || ''
-      const labelB = b.label || b.prefLabel || ''
-      const lang = window.SKOSMOS.content_lang ? window.SKOSMOS.content_lang : window.SKOSMOS.lang
+    compareStrs (a, b) {
```
osma, line 164: I see you've renamed the function from `compareLabels` to `compareStrs`, to reflect that it's no longer comparing just labels. However, since it takes two concepts (`a` and `b`) as parameters, I'd suggest calling it `compareConcepts` instead.

osma (changes requested):
Seems to work very well on the example vocabularies I tested (YKL, YSO-aika, okm-tieteenala). However, we don't really have a recent use case for numeric sorting of notations (the PTVL classification that required this was removed from Finto some time ago) so I couldn't test that with a real example. Anyway, I suppose it's good enough given that we don't even need it at the moment :)

I'd like to see Cypress tests that verify that the order in the hierarchy follows the notations, could you add some? I know that we're currently lacking Cypress tests for a lot of sidebar functionality, so adding tests could also be done as a separate effort if you prefer.

I made a suggestion to rename the compareStrs function to compareConcepts.

osma (approved):
LGTM!

PR #1652, opened 2024-08-20
Increase HTTP timeout in test configuration to fix flaky Cypress mapping tests

osma:

Some Cypress runs were failing under GitHub Actions CI. In particular the "concept mappings / full" and/or "concept mappings / partial" tests for the concept page often failed. For example [this run](https://github.com/NatLibFi/Skosmos/actions/runs/10404988178/job/28814722532#step:6:700).

Adjusting the Cypress timeouts doesn't help, because the problem is that the underlying Skosmos resolver code (LinkedDataResolver) times out. The reason is that we've set a 2 second httpTimeout in testconfig.ttl, used to configure Skosmos for the PHPUnit and Cypress tests. This is too short to reliably resolve some of the mappings (LCSH, KOKO, YSA etc.) especially under the GitHub Actions CI environment.

The fix is simple: increase the timeout to 8 seconds. (Sidenote: this needs to be different from the default value of 5 seconds, because it is tested in a PHPUnit test that needs to verify that the non-default setting is effective. The test has been changed accordingly)

- n/a, but failing tests were seen in e.g. PR #1651 

- increase httpTimeout value from 2 to 8 in testconfig.ttl
- modify GlobalConfigTest accordingly

Not sure if 8 is the ideal value here but it's better than 2 for sure.

## Checklist

### DIFF
--- a/tests/GlobalConfigTest.php
+++ b/tests/GlobalConfigTest.php
@@ -46,7 +46,7 @@ class GlobalConfigTest extends PHPUnit\Framework\TestCase
 
     public function testGetHttpTimeout()
     {
-        $this->assertEquals(2, $this->config->getHttpTimeout());
+        $this->assertEquals(8, $this->config->getHttpTimeout());
     }
 
     public function testGetServiceName()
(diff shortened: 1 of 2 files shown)


osma:
Re-running Actions jobs twice, so that the Cypress tests have been run six times under CI in total. If they pass every time, I will merge this.

osma:
All tests were successful, merging this.

PR #1657, opened 2024-08-21
SEO: vocab description, site name, type, canonical URL

osma:

This PR implements most of the remaining SEO functionality from #1533 :

* vocabulary description on vocab home page
* og:site_name on all pages
* og:type="website" on all pages
* og:url and link rel=canonical on all pages
* drop twitter:card (shouldn't be needed since twitter/X falls back on og:type)

In addition, there is a bug fix to the Request.getLangUrl method that broke when the Skosmos 3 code was restructured by moving backend code under `src/`. This fix could have been a separate PR but since the bug was causing problems for the canonical URL generation, I decided to fix it here in this PR. The fix is in a separate commit 9bdb8948d4246a410181db32cdc7dcf0efd2f309 .

- part of #1533

* add VocabularyConfig.getDescription() method + PHPUnit test
* add new metadata elements to base-template.twig (and drop twitter:card)
* add blocks in page type templates to populate the metadata
* add cypress tests
* fix bug in Request.getLangUrl() method + correct the PHPUnit tests accordingly

None so far!

## Checklist

### DIFF
--- a/src/model/Request.php
+++ b/src/model/Request.php
@@ -210,7 +210,7 @@ class Request
      */
     public function getLangUrl($newlang=null)
     {
-        $script_name = str_replace('/index.php', '', $this->getServerConstant('SCRIPT_NAME'));
+        $script_name = str_replace('/src/index.php', '', $this->getServerConstant('SCRIPT_NAME'));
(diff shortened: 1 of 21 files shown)


joelit (approved):
As far as I can understand, looks good to me!

PR #1658, opened 2024-08-21
Override default property labels on concept page

UnniKohonen:

It is not currently possible to override default property labels on the concept page. This PR adds this functionality.

- part of #1484

- Override default property labels in concept card and mappings
- Cypress tests
- Remove `skosmos:sparqlGraph` config option from `conceptPropertyLabels` test vocab
- Refactor mappings template
- Fix bug in mappings URL

## Checklist

### DIFF
--- a/resource/js/concept-mappings.js
+++ b/resource/js/concept-mappings.js
@@ -9,12 +9,14 @@ const conceptMappingsApp = Vue.createApp({
     }
   },
   provide: {
-    content_lang: window.SKOSMOS.content_lang,
-    customLabels: window.SKOSMOS.customLabels
+    content_lang: window.SKOSMOS.content_lang
   },
   computed: {
     hasMappings () {
       return Object.keys(this.mappings).length > 0
+    },
+    customLabels () {
+      return window.SKOSMOS.customLabels
     }
   },
   methods: {
@@ -64,7 +66,7 @@ const conceptMappingsApp = Vue.createApp({
       </template>
       <template v-else-if="hasMappings">
         <div class="main-content-section p-5">
-          <concept-mappings :mappings="mappings"></concept-mappings>
+          <concept-mappings :mappings="mappings" :customLabels="customLabels"></concept-mappings>
         </div>
       </template>
     </div>
@@ -85,8 +87,8 @@ conceptMappingsApp.directive('load-concept-page', {
 })
 
 conceptMappingsApp.component('concept-mappings', {
-  props: ['mappings'],
-  inject: ['content_lang', 'customLabels'],
+  props: ['mappings', 'customLabels'],
+  inject: ['content_lang'],
   template: `
     <div class="row property prop-mapping" v-for="(mapping, label) in mappings">
       <template v-if="customLabels">
--- a/resource/js/concept-mappings.js
+++ b/resource/js/concept-mappings.js
@@ -9,7 +9,8 @@ const conceptMappingsApp = Vue.createApp({
     }
   },
   provide: {
-    content_lang: window.SKOSMOS.content_lang
+    content_lang: window.SKOSMOS.content_lang,
+    customLabels: window.SKOSMOS.customLabels
   },
   computed: {
     hasMappings () {
@@ -20,7 +21,7 @@ const conceptMappingsApp = Vue.createApp({
     loadMappings () {
       this.mappings = {} // clear mappings before starting to load new ones
       this.loading = true
-      const url = 'rest/v1/' + window.SKOSMOS.vocab + '/mappings?uri=' + window.SKOSMOS.uri + '&external=true&clang=' + window.SKOSMOS.lang + '&lang=' + window.SKOSMOS.content_lang
+      const url = 'rest/v1/' + window.SKOSMOS.vocab + '/mappings?uri=' + window.SKOSMOS.uri + '&external=true&lang=' + window.SKOSMOS.lang + '&clang=' + window.SKOSMOS.content_lang
       fetchWithAbort(url, 'concept')
         .then(data => {
           return data.json()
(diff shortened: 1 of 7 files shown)


tests/testconfig.ttl
```diff
@@ -342,8 +342,7 @@
           rdfs:label "Caption"@en, "Luokka"@fi ],
         [ skosmos:property skos:notation ;
           rdfs:label "UDC number"@en, "UDC-numero"@fi, "UDC-nummer"@sv ;
-          rdfs:comment "Class Number"@en ] ;
-	skosmos:sparqlGraph <http://www.skosmos.skos/conceptPropertyLabels/> .
```
osma, line 346: Why did you remove this setting? (I'm sure it's OK to remove, I'd just like to understand the reason)

osma (changes requested):
I tested this using the YKL property label override configuration from dev.finto.fi, i.e. this:

```
	skosmos:propertyLabelOverride
    [ skosmos:property skos:prefLabel ;
          rdfs:label "luokka"@fi, "klass"@sv, "class"@en ] ,
    [ skosmos:property skos:notation ;
          rdfs:label "luokkatunnus"@fi, "klassindex"@sv, "notation"@en ] ,
    [ skosmos:property skos:altLabel ;
          rdfs:label "Hakusanat"@fi, "nyckelord"@sv, "keywords"@en ; rdfs:comment "Luokan hakusanat."@fi, "Nyckelord för klassen."@sv, "Keywords for this class."@en ] ,
    [ skosmos:property skos:broader ;
          rdfs:label "yläluokka"@fi, "överordnad klass"@sv, "superclass"@en ; rdfs:comment "Merkitykseltään laajempi luokka."@fi, "Klass vars betydelse är vidare."@sv, "Superclass."@en ] ,
    [ skosmos:property skos:narrower ;
          rdfs:label "alaluokat"@fi, "underordnade klasser"@sv, "subclasses"@en ; rdfs:comment "Merkitykseltään rajatummat luokat."@fi, "Klasser vars betydelse är snävare."@sv, "Subclasses."@en ] ,
    [ skosmos:property skos:related ;
          rdfs:label "assosiatiiviset luokat"@fi, "relaterade klasser"@sv, "related classes"@en ; rdfs:comment "Läheisesti luokan aiheeseen liittyvät luokat."@fi, "Klasser som är nära relaterade till klassens ämne."@sv, "Classes related to this class."@en ] ,
    [ skosmos:property skos:closeMatch ;
          rdfs:label "lähes vastaavat luokat"@fi, "nästan motsvarande klasser"@sv, "closely matching classes"@en ] ;
```

Here is how the result looks on the concept page:

image

The property labels are working like they should! However, they are not automatically capitalized, so they look different from regular properties. While capitalization of property labels is technically a separate feature, I think it would make sense to implement it as part of this PR, if it's not too much work, because it is often needed for custom property labels like these.

The configuration above also defines property label overrides for the mapping property skos:closeMatch ("closely matching classes"), but it is not used when displaying mappings:

image

I noticed that it's not working in Skosmos 2 either, so maybe it's not super critical; however, it would be great for consistency if the property label overrides could be applied also for mapping properties.

resource/js/concept-mappings.js
```diff
@@ -20,7 +21,7 @@ const conceptMappingsApp = Vue.createApp({
     loadMappings () {
       this.mappings = {} // clear mappings before starting to load new ones
       this.loading = true
-      const url = 'rest/v1/' + window.SKOSMOS.vocab + '/mappings?uri=' + window.SKOSMOS.uri + '&external=true&clang=' + window.SKOSMOS.lang + '&lang=' + window.SKOSMOS.content_lang
+      const url = 'rest/v1/' + window.SKOSMOS.vocab + '/mappings?uri=' + window.SKOSMOS.uri + '&external=true&lang=' + window.SKOSMOS.lang + '&clang=' + window.SKOSMOS.content_lang
```
osma, line 24: Good catch!

src/view/scripts.inc
```diff
@@ -24,6 +24,14 @@ window.SKOSMOS = {
   {%- if request.plugins.callbacks ~%}
   "pluginCallbacks": [{% for function in request.plugins.callbacks %}{% if not loop.first %}, {% endif %}"{{ function }}"{% endfor %}],
   {%- endif ~%}
+  {%- if custom_labels -%}
+  "customLabels": { 
+    {%- for k, v in custom_labels -%}
+      {%- set label = v['label'][request.lang] -%}
+      {%- set description = v['description'][request.lang] -%}
+      {%- if not loop.first -%}, {%- endif -%}"{{k}}": ["{{label[:1]|upper ~ label[1:]}}", "{{description[:1]|upper ~ description[1:]}}"]{%- endfor -%}
```
osma, line 32: I wonder if there's a danger here that the custom labels may contain special characters like `"` or `\` which could break the JS syntax?

OTOH the same problem also applies to other values generated above, for example concept labels under `"prefLabels"`. They too could in principle contain special characters.

It's an unlikely scenario though...

The way to avoid it would be to generate the data structure on the PHP side and return a JSON encoded string that could be inserted here using a single Twig expression.

tests/testconfig.ttl
osma: I still think another solution is needed here. I can take a look...

osma (changes requested):
I gave a few comments for consideration.

While testing this, I noticed that the mapping label overrides didn't always seem to work. This seems to depend on how you navigate to the concept page - partial page loads are maybe causing issues? Here is a test sequence that shows the problem:

1. Configure YKL property name overrides (see above comment)
2. Go to YKL home page; for me this is http://localhost/Skosmos/ykl/en/
3. In the hierarchy sidebar, click on the arrow next to the first class "0 GENERAL WORKS ..." to open it
4. Click on "00 BOOKS AND PUBLISHING" in the sidebar
5. Look at the mappings, they don't display the overridden label "Closely matching classes" (but if you reload the page, it will be applied)

image

tests/testconfig.ttl
osma: Solved.

osma (approved):
It's good as it is. The JS syntax problem is not likely to happen and anyway it's a wider issue than just this PR/feature.

PR #1664, opened 2024-08-22
Fix flaky getResourceFromUri test by increasing httpTimeout (Skosmos 2)

osma:

One of the Skosmos 2 PHPUnit tests (ModelTest::testGetResourceFromUri) was frequently failing, e.g. [this run](https://github.com/NatLibFi/Skosmos/actions/runs/10504426502/job/29099839944?pr=1663).

The problem is that this test triggers a HTTP request to Finto, which can sometimes take a bit longer than 2 seconds. This causes a HTTP timeout in the LinkedDataResolver code. The fix is to increase the httpTimeout in testconfig.ttl to 8 seconds.

A very similar fix for another flaky test was done recently for Skosmos 3 in PR #1652 .

none

- increase httpTimeout from 2 to 8 seconds in testconfig.ttl
- change the PHPUnit test that checks the httpTimeout value accordingly

8 seconds should be enough for anyone, right?

## Checklist

### DIFF
--- a/tests/GlobalConfigTest.php
+++ b/tests/GlobalConfigTest.php
@@ -43,7 +43,7 @@ class GlobalConfigTest extends PHPUnit\Framework\TestCase
 
     public function testGetHttpTimeout()
     {
-        $this->assertEquals(2, $this->config->getHttpTimeout());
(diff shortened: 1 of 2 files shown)


osma:
CI tests passed, merging this.

PR #1667, opened 2024-08-22
Css fixes for background for different page types

joelit:
This PR is built upon #1661 

## Reasons for creating this PR

 - Fixes background colours for different Skosmos page types
   - White for the landing page
   - Light hue for search result page, about page & feedback page
   - Medium hue for vocabulary page & concept page
 - Spacing after the header bar and before the footer bar
 - Removing the large Skosmos logo from about page & feedback page

## Description of the changes in this PR

Twig logic and bootstrap class changes

## Known problems or uncertainties in this PR

None

## Checklist

- [x] phpUnit tests pass locally with my changes
- [x] I have added tests that show that the new code works, or tests are not relevant for this PR (e.g. only HTML/CSS changes)
- [x] The PR doesn't reduce accessibility of the front-end code (e.g. tab focus, scaling to different resolutions, use of `.sr-only` class, color contrast)
- [x] The PR doesn't introduce unintended code changes (e.g. empty lines or useless reindentation)

PR #1668, opened 2024-08-22
Css fixes for the header bar

joelit:
This PR is built upon #1667 

Fixes for the header bar
 * Increase the Skosmos logo size by 20%
 * Include the navigation link icons

None

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -165,30 +165,6 @@ body {
     padding-left: 30px;
   }
 
-  #topbar.frontpage #topbar-nav-home, #topbar.frontpage-one-vocab #topbar-nav-home {
-    background-image: var(--topbar-home-1);
-  }
-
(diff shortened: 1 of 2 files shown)


UnniKohonen (changes requested):
Othewise looks good to me, but I think everything between lines 161 and 190 should be deteled from skosmos.css. Lines 114-148 look to me like they are unused right now, but might become relevant in the future.

joelit:
Good call! The skosmos.css lines 161-165 are used, but the 166-190 seem like some kind of legacy CSS that should be removed.

UnniKohonen:
The lines 114-148 are from the [static html version of Skosmos-3](https://github.com/NatLibFi/Skosmos/tree/skosmos-3-html) and have mainly to do with the different color scales which we don't have yet.

PR #1671, opened 2024-08-28
Upgrade EasyRdf to fix Turtle fragment URIs

### LINKED ISSUE #1591: Turtle output with fragment URIs is invalid
## Description of the problem
I copy-pasted Skosmos Turtle output into http://ttl.summerofcode.be and it does parse, nor does it with Jena's riot

The error:
```
Expected punctuation to follow "https://data.beamery.com/aaa/bbb/ccc/1287" on line 14.
```
The relevant bit of data looks like this:
```
@prefix smth: <https://data.beamery.com/aaa/bbb/ccc/> .

... skos:relatedMatch smth:1287#this

```

I don't think the `prefix:local-name#fragment` syntax is valid. You should look into the RDF I/O library that is producing this.

osma:

We were using EasyRdf 1.7.x which is quite old and produces invalid Turtle fragment URIs in some cases, see #1591. The fix is to upgrade to easyrdf 1.10.x or newer. This PR upgrades to EasyRdf 1.13.x which is the most recent release that still works without breaking Skosmos 2 PHPUnit tests; EasyRdf 1.14.1 is currently the newest release. I don't think it makes sense to do extensive debugging of the failing Skosmos 2 unit tests, so I'm leaving the dependency at 1.13.x.

We need to drop support for PHP 7.x because EasyRdf 1.10 and above require at least PHP 8.0. This leaves only two supported PHP versions for Skosmos 2: PHP 8.0 and 8.1.

- Fixes #1591 (for skosmos-2)

- upgrade to sweetrdf/easyrdf 1.13.x
- drop PHP 7.3 and 7.4 from the GitHub Actions CI test because they're incompatible with recent EasyRdf versions
- update README section on supported versions

I didn't investigate closer why some PHPUnit tests fail when using easyrdf 1.14.x.

We need to do a similar EasyRdf upgrade for Skosmos 3 as well (main branch).

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php_version: ["7.3", "7.4", "8.0", "8.1"]
+        php_version: ["8.0", "8.1"]
         experimental: [false]
         include:
           - php_version: "8.2"
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Skosmos is used as a basis for the [Finto](http://finto.fi) vocabulary service.
 The latest development version is also available at 
 [dev.finto.fi](http://dev.finto.fi).
 
-Skosmos is implemented using PHP (supported versions: 7.3, 7.4 and 8.0), with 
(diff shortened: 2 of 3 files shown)


joelit (approved):
LGTM

PR #1682, opened 2024-10-01
Add PHP 8.3 support + change Docker base image to Ubuntu 24.04 with PHP 8.3

### LINKED ISSUE #1665: PHP 8.3 support for Skosmos 3

PHP 8.3 was released in November 2023. By now I'd expect that most PHP libraries that we use for Skosmos 3 have been updated to work with PHP 8.3.

We've tried to support at least three consecutive PHP versions in Skosmos. However, PR #1648, when merged, will remove support for PHP 8.0, bringing the number of supported PHP releases to only two (8.1 and 8.2). We should at least investigate how far we are from supporting 8.3. It's possible that everything is working already, we just haven't tested running under 8.3. It's possible that some dependencies need to be upgraded to newer versions that work under PHP 8.3.

Skosmos developers benefit from staying up to date with a modern codebase.

Skosmos installations may use different PHP versions depending on which Linux distribution/environment they have. Supporting several PHP versions makes it easier to deploy and maintain Skosmos in different environments.

It would allow deploying Skosmos under PHP 8.3. For example Ubuntu 24.04 comes with PHP 8.3 and installing an earlier version requires extra work. 

PHP 8.3 should also be faster than earlier versions.

This is general maintenance. We need to do it anyway at some point.

osma:

This PR adds support for PHP 8.3 in Skosmos 3 and upgrades the Docker base image to Ubuntu 24.04 which includes PHP 8.3.

- Closes #1665

- enable PHP 8.3 build in GitHub Actions CI
- upgrade PHP-CS-Fixer to 3.64.0 to support PHP 8.3
- change the PHP code style to match new defaults in PHP-CS-Fixer
- upgrade Docker base image to Ubuntu 24.04 with PHP 8.3

None

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php_version: ["8.0", "8.1", "8.2"]
+        php_version: ["8.0", "8.1", "8.2", "8.3"]
         experimental: [false]
 
     steps:
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     "phpunit/phpunit": "9.5.*",
     "symfony/dom-crawler": "5.4.*",
     "mockery/mockery": "1.5.1",
-    "friendsofphp/php-cs-fixer": "3.16.*"
+    "friendsofphp/php-cs-fixer": "3.64.*"
   },
   "autoload": {
     "classmap": ["src/controller/", "src/model/", "src/model/sparql/"]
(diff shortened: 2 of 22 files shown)


osma:
> The [Skosmos install tutorial](https://github.com/NatLibFi/Skosmos/wiki/InstallTutorial#install-apache-and-php) should be updated to PHP 8.x (8.2?).

The tutorial is written for Skosmos 2, so it's not relevant to this PR, which is about Skosmos 3. But you are right that it should be updated, because PHP 7.4 no longer works. PHP 8.1 (in Ubuntu 22.04) would be the obvious choice as it's the newest PHP version that Skosmos 2 supports.

PR #1688, opened 2024-10-02
Add cypress tests for alphabetical index Vue component

UnniKohonen:

Alphabetical index has no cypress tests, this PR adds them.

- Part of #1563

- Added cypress tests for tab-alpha Vue component
- Added a new test vocab for testing loading on scroll

Addresses requirement 6 in #1563

## Checklist


osma (changes requested):
The existing tests look good. A few thoughts below.

I think there could also be a test that checks for altLabels that should appear in the index. For example in the YSO Archaeology test vocabulary, there is this altLabel entry: `artefacts → articles (inanimate objects)`

Also, would it be a good idea to check the links to concepts as well? For example verify that the target URL is correct and/or simulate a click and verify that the concept is loaded. OTOH, there are tests like this already in vocab-home.cy.js so maybe there is no need to duplicate those. Maybe the vocab-home.cy.js tests could be moved alongside these, since they exercise the hierarchy sidebar?

All the new sidebar-related tests are now in sidebar.cy.js. This file may grow quite big since the sidebar has several components and more will be added later. Would it make sense to split it already now into e.g. sidebar-alpha.cy.js, sidebar-hierarchy.cs.js etc?

osma (approved):
LGTM!

Issue #1656, opened 2024-08-21
Bring translations managed by Symfony into use in Vue templates


miguelvaara:
## Description of the enhancement
Symfony manages the translations of different languages used in Skosmos and the messages they contain are imported into the code using language tags (tags in the language files / json). This is the case elsewhere in the front-end code (Twig templates), but not yet in Vue templates. Therefore to streamline development, the same features must be available in Vue code as well.

It is recommended to aim to utilize the Vue i18n library.

## Who are the users that would benefit from the enhancement and how?
Developers / end users

(Name specific organizations, roles and/or communities if possible, and/or point to relevant discussion elsewhere)

## What new functionalities would the enhancement make possible?
Quick and easy addition of translations to the front-end code, including Vue.

## Why is the enhancement important?
This improvement makes it easier to deploy multilingual Skosmos installations and significantly streamlines the implementation of the language features expected by users.

(Why should this be prioritized, instead of spending energy on something else? Who would do the work to make this happen?)

PR #1690, opened 2024-10-03
The first version of translation functionality in Twig templates, whe…

miguelvaara:
FOR DESIGN REVIEW! This is not a PR for a final review

I would improve the implementation as follows:

Refer directly to Skosmos's `lang` variable instead of maintaining an unnecessary list of them on the [frontend](https://github.com/NatLibFi/Skosmos/blob/cdd654e4bebe3bef763cf33095ccfd6367c3ca82/resource/js/vocab-counts.js#L2).

...a bit of code brainstrorming:

translationService.js
```
export async function loadTranslationMessages() {
    const response = await fetch(`http://localhost/skosmos/resource/translations/messages.${locale}.json`);
        const data = await response.json();
        messages[`${lang}`] = data;
        etc etc ...
        
export async function createI18nInstance() {
    const messages = await loadTranslationMessages();
    ...
```

And in the component:

```
import { createI18nInstance } from './translationService';

async function initializeApp() {
    const i18n = await createI18nInstance();

    const resourceCountsApp = Vue.createApp({
    ...
    resourceCountsApp.use(i18n);
    resourceCountsApp.mount('#resource-counts');
}
initializeApp();
```

**Finally the same in the test**

Translations are now working well in the example for both the vocab-counts and term-counts components.

I made significant changes to the previous implementation (moving entire blocks inside new ones etc.). One major change is the use of an anonymous function.

The basic idea is that there is a single JavaScript component that provides translations as a service, which is then imported and used in the components where translations are needed. The array containing the list of languages is now removed and SKOSKOS.langi is used.

Translations also need to work on the frontend side and Vue components

https://github.com/NatLibFi/Skosmos/issues/1656

- Closes #
#1656

- Created a translation service component
- The translation service is invoked and used in two JS/Vue components
- The method of invoking JS/Vue components in the Twig template has been changed to a module-based approach
- Code has been modified towards the use of anonymous functions

## Checklist


resource/js/translation-service.js
```diff
@@ -0,0 +1,24 @@
+export async function loadLocaleMessages(locale) {
+    const messages = {};
+    try {
+      const response = await fetch(`http://localhost/skosmos/resource/translations/messages.${locale}.json`);
```
osma, line 4: This didn't work for me, because my local Skosmos installation lives on a slightly different URL. The URL should be changed to a relative one. This worked:

    const response = await fetch(`resource/translations/messages.${locale}.json`);

miguelvaara:
Now there is an implementation where both statistics (vocab-counts & term-counts) on the vocabulary homepage are displayed at the same time, but the translation initialization is only done once (see image). It was a long and frustrating process where only one of the Vue apps would load at a time :sweat_smile:. The issue is that I had to build a separate timing function since Vue now relies on the translation service. The timing conflict is now resolved with a separate timing function :face_with_peeking_eye:. The setup works, but generally speaking, when using frameworks, it feels a bit unsophisticated. I will not commit yet -  I eill clean up the code first and m
Screenshot from 2024-10-09 15-41-02
ake sure I fully understand what "spaghetti" have appearead through my own hands! :joy:

UnniKohonen (commented):
This solution works and seems fine for now. In the future, we can think about using the `vue-i18n` library if it adds any value. It would eliminate the need to check the availability of the `$t` function, but there would be some blinking with the component texts. Also, it might be worth investigating if promises could be used instead of `setTimeout` when checking for `$t` and starting Vue components. Though, based on my brief testing and googling it seems that `setTimeout` is the most straightforward way to wait for the function to be available.

osma (approved):
LGTM, we can continue from here later

PR #1693, opened 2024-10-09
Fix setting reading the search language cookie

### LINKED ISSUE #1692: Content language setting is not working consistently in Skosmos 3

http://localhost/Skosmos/ykl/fi/

## Description of the problem

1. Clear cookies for localhost (just in case)
2. Go to http://localhost/Skosmos/ykl/en/
3. In the content language dropdown, select Finnish (this will set the SKOSMOS_SEARCH_LANG cookie to `fi` and reload the page with the `?clang=fi` attribute)
4. Again go to http://localhost/Skosmos/ykl/en/ (without the clang parameter)
5. Look at the content language selection: it shows Finnish, while the hierarchy shows class names in English.
6. Open one of the branches of the hierarchy (e.g. 0 GENERAL...)
7. The classes in the branch now have class names in Finnish

image

I think the problem happens because of the way the SKOSMOS_SEARCH_LANG cookie is handled. The idea is to remember the content language chosen by the user. But in step 4 above, by the time the hierarchy renders, the SKOSMOS.content_language setting is still `en` and thus the top level of the hierarchy is rendered in English. When SKOSMOS.content_language is changed later by the search component, it's already too late, although subsequent request (for example opening new branches in step 6) will use the new setting.

Either we should drop the SKOSMOS_SEARCH_LANG cookie completely (it seems to me that having two mechanisms for this - the clang parameter and the cookie - is a recipe for problems) or if we need to keep it, then at least it should be processed very early in the frontend code (or possibly already on the backend side) so that SKOSMOS.content_language always matches the language set in the cookie.

Firefox

joelit:

There was a bug reading the content language cookie, due to a method not being updated to follow the new cookie convention. This PR is built upon #1689 .

- Closes #1692 

Changed the name of the cookie

* Should there be an additional test to catch the bug described in #1692 ?
* Does this bug warrant a closer inspection of the content language & UI language management logic?

## Checklist


osma (changes requested):
This is not the right way to fix #1692. There are, or should be, two cookies: SKOSMOS_LANGUAGE to remember the UI language and SKOSMOS_SEARCH_LANG to remember the content/search language. Whether we actually need the latter is debatable (as explained in #1692) but mixing up these two doesn't help.

osma:
> I don't think nothing sets the SKOSMOS_LANGUAGE cookie anymore, though. It looks like it's becoming dead code in 3.0.

You're right that it's not being set by the current Skosmos 3 code. In the Skosmos 2 code, we set event handlers for the UI language selection links:

https://github.com/NatLibFi/Skosmos/blob/6b9b380a62949f1715223ea018b1201c56498177/resource/js/docready.js#L488-L494

I think the same should be done for Skosmos 3. I can open a new issue about this.

osma:
Opened a new issue #1696

joelit:
Closing this PR, as the preferred solution was to get rid of the search language cookie. For example, the content language state of one vocabulary might not make sense when switching to another vocabulary with different supported languages. Better to handle search language with clang url parameter.

Issue #1482, opened 2023-08-22
Vocabulary Home twig-templates


joelit:
## Description of the enhancement

This issue is for making sure the initial version of the vocabulary home page template fulfills common requirements. It's OK if not all functionality is implemented.

- [ ] Layout follows [the static example](https://github.com/NatLibFi/Skosmos/blob/skosmos-3-html/vocabpage.html)
 - [ ] Base .twig -template exists
 - [ ] .inc -templates exist
 - [ ] cypress tests exist

PR #1697, opened 2024-10-09
Reimplement vocab home using Bootstrap Grid instead of tables

osma:

The vocabulary home page looks ugly and is using twig template structures copied from Skosmos 2, for example HTML tables. This PR reimplements the main part and the download links using Bootstrap Grid, similar to the concept page.

Also contains a minor fix to the concept page. The template code is now aligned between these two so that partial page loads work better.

- part of #1482

- reimplement vocab info twig template using Bootstrap Grid (+ corresponding CSS changes)
- align concept and vocab home template structure

- property-specific Cypress tests needed; but these are difficult to do because of the simplistic way that the backend provides vocab info properties as a key-value lists

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -677,7 +677,7 @@ body {
   }
 
   /*****  Main content vocabpage & termpage *****/
-  #main-content h2 {
+  #main-content h1 {
     font-size: 2.5rem;
   }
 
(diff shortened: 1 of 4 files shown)


joelit (approved):
LGTM! I unchecked one box from the corresponding issue, since the PR didn't implement the feature (show title only once, even if it is present twice).

Screenshot from 2024-10-10 15-22-42

PR #1706, opened 2024-10-29
Don't HTML encode quotes in REST API callback, to fix ONKI Selector (Skosmos 3)

osma:

See PR #1705. This is the same fix but for Skosmos 3.

n/a

- when sanitizing the callback parameter, apply the [FILTER_FLAG_NO_ENCODE_QUOTES](https://www.php.net/manual/en/filter.constants.php#constant.filter-flag-no-encode-quotes) option which prevents encoding quotes

n/a

## Checklist

### DIFF
--- a/src/controller/RestController.php
+++ b/src/controller/RestController.php
@@ -25,7 +25,7 @@ class RestController extends Controller
         // wrap with JSONP callback if requested
         if (filter_input(INPUT_GET, 'callback', FILTER_SANITIZE_FULL_SPECIAL_CHARS)) {
             header("Content-type: application/javascript; charset=utf-8");
-            echo filter_input(INPUT_GET, 'callback', FILTER_SANITIZE_FULL_SPECIAL_CHARS) . "(" . json_encode($data) . ");";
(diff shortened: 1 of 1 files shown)


joelit:
Stupid question, but why should the quotes be encoded differently when the response is optionally wrapped in a callback? [In the following block](https://github.com/NatLibFi/Skosmos/blob/84b37b4cbd7bab294f3a34f4c9f01e7d9bd0b08e/src/controller/RestController.php#L35) of the same function.

osma:
@joelit I think the idea of encoding is to try to prevent various kinds of injection attacks by encoding special characters. These need to be handled case by case. All special characters that should not be needed at all in the context are best encoded, i.e. make the encoding as comprehensive as possible without breaking functionality. However, in this case, the encoding of quotes in the callback parameter actually broke things, so it was relaxed a little bit to prevent that problem.

The line you refer to is for processing the HTTP Accept header. I don't think that header is supposed to contain quotes, so encoding them does no harm and there is no need to prevent encoding of quote characters with a flag.

PR #1711, opened 2024-12-10
Set HTTP user-agent to 'Skosmos' when performing requests

### LINKED ISSUE #1691: Set user agent when performing HTTP requests

Currently when Skosmos performs HTTP requests (e.g. SPARQL queries and requests for resolving Linked Data), it uses EasyRdf library to perform the actual request. EasyRdf sets its own User-Agent header `EasyRdf HTTP Client`. This may look suspicious; for example recently CloudFlare blocked some requests from EasyRdf.

The suggestion is to set a custom User-Agent header. The most logical value would be `Skosmos X.Y` where X.Y is the version number that is already generated for example for the `<meta name="generator">` tag in the HTML pages. 

Any organization using Skosmos. 

The HTTP requests performed by Skosmos would be more reliable and easier to trace.

It reduces potential problems.

osma:

See #1691. It's good practice to use a custom HTTP user agent instead of relying on library defaults. This code sets the user-agent to `Skosmos` every time EasyRdf is used to perform HTTP requests (including SPARQL). There is one more place in the code (LOCResolver) that performs HTTP requests directly using stream operations, but in that code the user-agent was already set to `Skosmos`.

I decided not to include the Skosmos version number in the user-agent string based on discussion with other Skosmos developers. 

- Closes #1691

- set the HTTP user agent (via EasyRdf config setting) in all places where the Skosmos backend code performs HTTP requests

No unit tests, but I don't think it's easy to test this with PHPUnit. However, I tested manually that the changes do have an effect by looking at HTTP server logs.

## Checklist

### DIFF
--- a/src/model/resolver/LinkedDataResource.php
+++ b/src/model/resolver/LinkedDataResource.php
@@ -16,7 +16,7 @@ class LinkedDataResource extends RemoteResource
         try {
             // change the timeout setting for external requests
             $httpclient = EasyRdf\Http::getDefaultHttpClient();
-            $httpclient->setConfig(array('timeout' => $timeout));
+            $httpclient->setConfig(array('timeout' => $timeout, 'useragent' => 'Skosmos'));
             EasyRdf\Http::setDefaultHttpClient($httpclient);
 
             $graph = EasyRdf\Graph::newAndLoad(EasyRdf\Utils::removeFragmentFromUri($this->uri));
--- a/src/model/resolver/WDQSResource.php
+++ b/src/model/resolver/WDQSResource.php
@@ -11,7 +11,7 @@ class WDQSResource extends RemoteResource
             EasyRdf\Format::unregister('json');
             // change the timeout setting for external requests
             $httpclient = EasyRdf\Http::getDefaultHttpClient();
-            $httpclient->setConfig(array('timeout' => $timeout));
+            $httpclient->setConfig(array('timeout' => $timeout, 'useragent' => 'Skosmos'));
             $httpclient->setHeaders('Accept', 'text/turtle');
             EasyRdf\Http::setDefaultHttpClient($httpclient);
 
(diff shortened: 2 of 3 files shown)


miguelvaara (commented):
**How I tested**


In the command line:

`sudo ngrep -q -W byline 'User-Agent' tcp port 80`

In the browser's address bar:
http://localhost/skosmos/yso/en/page/p26179 (A-clinics)

Result:

```
interface: wlp0s20f3 (192.168.XXX.XXX/255.255.255.0)
filter: ( tcp port 80 ) and ((ip || ip6) || (vlan && (ip || ip6)))
match (JIT): User-Agent

T 192.168.XXX.XXX:53316 -> 128.214.XXX.XXX:80 [AP] #6
Host: api.dev.finto.fi.
Connection: close.
User-Agent: Skosmos.
Accept: application/sparql-results+json,application/sparql-results+xml;q=0.8.
.


T 192.168.XXX.XXX:53320 -> 128.214.XXX.XXX:80 [AP] #17
Host: api.dev.finto.fi.
Connection: close.
User-Agent: Skosmos.
Accept: application/sparql-results+json,application/sparql-results+xml;q=0.8.
.


T 192.168.XXX.XXX:53322 -> 128.214.XXX.XXX:80 [AP] #28
Host: api.dev.finto.fi.
Connection: close.
User-Agent: Skosmos.
Accept: application/sparql-results+json,application/sparql-results+xml;q=0.8.
.


T 192.168.XXX.XXX:53326 -> 128.214.XXX.XXX:80 [AP] #39
Host: api.dev.finto.fi.
Connection: close.
User-Agent: Skosmos.
Accept: application/sparql-results+json,application/sparql-results+xml;q=0.8.
```
...etc (a longer list is not meaningful because it works in all cases)

The code also looks meaningful.

Well done! :+1:  In my opinion, it can be merged.

PR #1715, opened 2024-12-10
Issue1709 frontend dependencies

### LINKED ISSUE #1709: Upgrade front-end dependencies
Bootstrap and Vue need to be updated on the production side. Cypress, Axe and StandardJS need to be updated on the development side.

joelit:

* We should update software library dependencies periodically, e.g. once a year
* Vue was already at a version that didn't satisfy our requirements for translation templates

- Closes #1709 

Some of the cypress tests are flaky, notably:
* concept.cy.js:
concept notation can be copied to clipboard / partial:
AssertionError: expected '33.10' to equal '33.2'
* sidebar-alpha.cy.js:
clicking on alphabetical index entries performs partial page load:
AssertionError: Timed out retrying after 4000ms: Expected to find content: 'vårdinrättningar (sv)' within the element: <div.col-sm-5.prop-mapping-label> but never did.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,9 +16,9 @@
     "vue": "^3.5.x"
   },
   "devDependencies": {
-    "axe-core": "^4.7.x",
-    "cypress": "^13.6.x",
-    "cypress-axe": "^1.4.x",
+    "axe-core": "^4.10.x",
+    "cypress": "^13.16.x",
+    "cypress-axe": "^1.5.x",
     "standard": "^17.0.x"
   }
 }
--- a/package.json
+++ b/package.json
@@ -12,13 +12,13 @@
     "test-accessibility-cli-vocab-search": "npx cypress run --env cli=true --spec 'tests/cypress/accessibility/axe-vocab-search.cy.js'"
   },
   "dependencies": {
-    "bootstrap": "^5.2.3",
-    "vue": "^3.2.47"
+    "bootstrap": "^5.2.x",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -12,13 +12,13 @@
     "test-accessibility-cli-vocab-search": "npx cypress run --env cli=true --spec 'tests/cypress/accessibility/axe-vocab-search.cy.js'"
   },
   "dependencies": {
-    "bootstrap": "^5.2.3",
-    "vue": "^3.2.47"
+    "bootstrap": "^5.3.x",
```
osma, line 15: Very minor remark: Using `x` here looks a bit weird to me, although it seems to be a valid convention.

We are using `*` in composer.json, would it be good to be consistent and use it here as well?

joelit: Will do!

osma (approved):
LGTM. I tested running Cypress tests from CLI and interactively. I also looked at the various page types and couldn't find any issues.

I agree that the flaky tests are occasionally a problem.

The clipboard test fails if the browser window (or the window used for running `npx cypress run`) doesn't have focus. This is documented in a comment within that test. I don't think we can do anything about this, because it's a security feature in the browser engine.

The other flaky test related to the alphabetical index could possibly be fixed by increasing the timeout so that the request for mappings will be able to finish. However, I think that's outside the scope of this PR.

PR #1717, opened 2024-12-11
Set the UI language cookie

### LINKED ISSUE #1696: Skosmos 3: Remember UI language using a SKOSMOS_LANGUAGE cookie

In Skosmos 2, when you change an UI language, the JS code sets a SKOSMOS_LANGUAGE cookie with the selected language (e.g. `en`). Next time you visit the site (via the base URL that doesn't include the language), the UI will be shown in that language.

Skosmos 3 should do this as well. We have the backend side code in WebController.php, but the frontend side JS code to set the cookie is missing. Here's how it has been implemented in Skosmos 2:

https://github.com/NatLibFi/Skosmos/blob/6b9b380a62949f1715223ea018b1201c56498177/resource/js/docready.js#L488-L494

https://github.com/NatLibFi/Skosmos/blob/6b9b380a62949f1715223ea018b1201c56498177/resource/js/scripts.js#L182-L185

Note: This should not be confused with the SKOSMOS_SEARCH_LANG cookie which is used in a similar way to remember the search/content language. (See #1692 for related discussion)

All Skosmos users. Especially in case they have misconfigured language settings in their browser.

Remember the UI language chosen previously.

It's simple to do and already works in Skosmos 2.

joelit:

- Closes #1696 

 - Fixed the navbar language menu
 - Added functionality for setting the language cookie
 - Added tests for setting and reading the language cookie

- The functionality in index.php and web controller seemed to be OK, they might need a tweak in this

## Checklist

### DIFF
--- a/resource/js/navbar-ui-languages.js
+++ b/resource/js/navbar-ui-languages.js
@@ -12,10 +12,10 @@ function addLanguageEventListeners() {
   languageLinks.forEach(function(link) {
     link.addEventListener('click', function(event) {
       event.preventDefault()
-      setLangCookie(this.getAttribute('href'))
+      const langValue = this.getAttribute('id').substring(9) //strip the 'language-' part of the id
+      setLangCookie(langValue)
     })
   })
 }
-
 // register the language link event listeners
 document.addEventListener('DOMContentLoaded', addLanguageEventListeners)
--- a/resource/js/navbar-ui-languages.js
+++ b/resource/js/navbar-ui-languages.js
@@ -1,4 +1,4 @@
-function setLangCookie(lang) {
+function setLangCookie (lang) {
   // The cookie path should be relative if the baseHref is known
   let cookiePath = '/'
   if (window.SKOSMOS.baseHref && window.SKOSMOS.baseHref.replace(window.origin, '')) {
@@ -6,11 +6,11 @@ function setLangCookie(lang) {
   }
   document.cookie = `SKOSMOS_LANGUAGE=${lang};path=${cookiePath}`
 }
-function addLanguageEventListeners() {
-  const languageLinks = document.querySelectorAll('.nav-item.language a');
-  languageLinks.forEach(function(link) {
-    link.addEventListener('click', function(event) {
-      const langValue = this.getAttribute('id').substring(9) //strip the 'language-' part of the id
+function addLanguageEventListeners () {
+  const languageLinks = document.querySelectorAll('.nav-item.language a')
+  languageLinks.forEach(function (link) {
+    link.addEventListener('click', function (event) {
+      const langValue = this.getAttribute('id').substring(9) // strip the 'language-' part of the id
       setLangCookie(langValue)
     })
   })
--- a/resource/js/navbar-ui-languages.js
+++ b/resource/js/navbar-ui-languages.js
@@ -1,0 +1,21 @@
+function setLangCookie(lang) {
+  // The cookie path should be relative if the baseHref is known
+  let cookiePath = '/'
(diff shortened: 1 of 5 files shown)


resource/js/navbar-ui-languages.js
```diff
@@ -0,0 +1,19 @@
+function setLangCookie (lang) {
+  // The cookie path should be relative if the baseHref is known
+  let cookiePath = '/'
+  if (window.SKOSMOS.baseHref && window.SKOSMOS.baseHref.replace(window.origin, '')) {
+    cookiePath = window.SKOSMOS.baseHref.replace(window.origin, '')
+  }
+  document.cookie = `SKOSMOS_LANGUAGE=${lang};path=${cookiePath}`
```
osma, line 7: This only sets the cookie for the duration of the session. I think the intent was to remember the language setting for the next time, so keeping it only for this session isn't enough. Skosmos 2 sets the cookie for 365 days, the same should be done here. Also, Skosmos 2 uses the "SameSite=Lax" parameter, is that needed here as well?

tests/cypress/template/navbar-languages.cy.js
osma, line 1: Does it make sense to have a separate test file just for navbar languages? Will it ever contain more tests than just this one?

Would it make more sense to call this e.g. `navbar.js` so that we can later add more navbar-related tests into the same file?

tests/cypress/template/navbar-languages.cy.js
```diff
@@ -0,0 +1,12 @@
...
+
+  it('Sets and reads SKOSMOS_LANGUAGE cookie correctly', () => {
+    cy.visit('/')
+
+    cy.get('.nav-item.language a').first().click()
+    cy.getCookie('SKOSMOS_LANGUAGE').should('exist')
+    cy.getCookie('SKOSMOS_LANGUAGE').should('have.property', 'value', 'fi')
+
+    cy.url().should('include', '/fi/')
+  })
```
osma, line 11: How about also visiting the site root `/` again to verify that the cookie had an effect?

osma (changes requested):
I added some notes. At least the TTL of the cookie needs to be changed.

tests/cypress/template/navbar-languages.cy.js
```diff
@@ -0,0 +1,15 @@
+describe('Navbar languages', () => {
+
+  it('Sets and reads SKOSMOS_LANGUAGE cookie correctly', () => {
+    cy.visit('/')
```
osma, line 4: I think there should be a test here to verify that the URL does *not* include `/fi/`. Then the test would demonstrate that the cookie setting actually had an effect on `/`.

osma (approved):
Looks good now. I made a couple of minor additional suggestions for you to consider.

resource/js/navbar.js
```diff
@@ -0,0 +1,29 @@
...
+  const date = new Date()
+  const msPerDay = 24 * 60 * 60 * 1000
+  const days = 365
+  date.setTime(date.getTime() + days * msPerDay)
+  const expires = '; expires=' + date.toGMTString()
+
+  const langValue = 'SKOSMOS_LANGUAGE=' + lang
+
+  document.cookie = langValue + expires + path + '; SameSite=Lax'
+}
```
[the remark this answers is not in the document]
joelit: Thanks for the tip! Should have had a look at SonarCloud.

tests/cypress/template/navbar-languages.cy.js
joelit: Good point!

PR #1718, opened 2024-12-11
SKOS XL popup on concept page

### LINKED ISSUE #1714: SKOS XL functionality for Skosmos 3 concept page

Show SKOS XL information on the Concept Page for Skosmos 3. Part of #1484.

Users of e.g. YSO.

Make the existing SKOS XL information available to the user e.g. in the form of tooltips.

The SKOS XL information (e.g. term source and notices about outdated terminology) is there for a reason.

## Task list

SKOS XL information, when available, should be shown for:

- [ ] prefLabel heading
- [ ] properties with literal values
- [ ] foreign language labels

osma:

See #1714. This PR implements a popup display for SKOS XL information for 

1. the concept prefLabel heading
2. properties with literal values
3. the "terms in other languages" section. 

The popup is implemented in pure CSS. There are also Cypress tests to verify that the popup is triggered when clicking the icons for all three cases.

The Cypress test file `concept.cy.js` was starting to become too unwieldy, so I split off some of the tests into a separate `concept-full-vs-partial.cy.js` file. The PR also contains a minor typo fix for the concept page template.

- Closes #1714

- popup displays for the SKOS XL information (all 3 cases listed above)
- cypress tests for the popup (all 3 cases)
- split off some tests from concept.cy.js to concept-full-vs-partial.cy.js for easier maintenance and faster test cycles
- typo fix for the download link (not related to other stuff in the PR, but I included it here because it's so small)

Here are a couple of examples showing what this looks like:

image

image

- accessibility could be better; for example, screen readers should be informed about the popup when it opens

## Checklist


joelit:
Looks good and works somewhat as expected. 

With mouse, the popup becomes visible with a mouse over, and it becomes hidden with a mouse out. The popup can locked in place by clicking the button so that it remains visible after mouse out. I'd expect like to see the popup becoming hidden with another click on the button. However, it seem this cannot be done with pure CSS buttons.

With touch screen (my laptop), the current functionality is: click on the button and the popup becomes visible, click on the button again and nothging happens, but click away and the popup becomes hidden.

My suggestions:
 - change the button to input type="checkbox" that is toggleable while remaining pure CSS
 - have a JS event listener to the button that toggles the visibility on and off
 - disregard this notion all together, as I'm not sure it's worth the effort

osma:
I added the remaining two cases of SKOS XL popups as well as Cypress tests for them.
I also split off some of the tests from `concept.cy.js` into a separate file.

See the updated description in the OP. Now ready for review.

Accessibility could probably be improved further, ideas welcome!

src/view/xl-label.inc.twig
```diff
@@ -0,0 +1,9 @@
+<div class="tooltip-html">
+  <button class="btn p-0 m-0"><i class="fa-solid fa-circle-info"></i></button>
```
joelit, line 2: The button element should have something other than the class "fa-circle-info" to indicate to screen readers that this is an info button. For example, `<button class="btn p-0 m-0" aria-label="More info on this label">` or similar. Furthermore, it could maybe have something like `aria-controls="id-for-dl-element"`.

src/view/concept-card.inc.twig
```diff
@@ -36,6 +36,9 @@
         {%- endif -%}
         </div>
         <div class="col-sm-8" id="concept-label">
+        {% if concept.hasXlLabel %}
+          {{ include('xl-label.inc.twig', {xlLabel: concept.xlLabel}) }}
+        {% endif %}
         {% if concept.notation %}
           <h1 id="concept-preflabel"
```
joelit, line 43: The prefLabel could have `aria-describedby="id-of-the-tooltip-html"` attribute to ling the label h1-element to the tooltip info. It is possibly easier to wrap the xl-label.inc.twig template inside a `<div id="id-of-the-tooltip-html">` element to keep the assigning of unique identifiers in the control of the concept-card.inc.twig

tests/cypress/template/concept.cy.js
```diff
@@ -356,6 +236,18 @@ describe('Concept page', () => {
...
+
+    // the tooltip should originally be invisible
+    cy.get('.prop-skos_altLabel .tooltip-html-content').should('not.be.visible')
+
+    // focus on the button to trigger the tooltip
+    cy.get('.prop-skos_altLabel .tooltip-html button').focus()
+
+    // the tooltip should now be visible
+    cy.get('.prop-skos_altLabel .tooltip-html-content').should('be.visible')
+  })
```
joelit, line 250: I would add hiding the tooltip:
```
    // loose the focus on the button to hide the tooltip
    cy.get('.prop-skos_altLabel .tooltip-html button').blur()
    
    // the tooltip should now be invisible again
    cy.get('.prop-skos_altLabel .tooltip-html-content').should('not.be.visible')
```

joelit (changes requested):
Looks pretty good! I could think of some clear improvements to accessibility, as requested.
Also a small addition to the cypress tests (example included in the review comments).

resource/css/skosmos.css
```diff
@@ -951,3 +954,82 @@ body {
...
+
+.tooltip-html button {
+  line-height: 1.0;
+  padding: 3px;
+  margin: -3px;
+}
+
+.tooltip-html button:focus {
+  border: 2px solid var(--vocab-text);
+  border-radius: 0;
```
joelit, line 977: I would add a compensation for the border in the margin like so:

```suggestion
  border-radius: 0;
  margin: -4px;
```
So that the text doesn't jitter when the button is pushed (Chromium, not sure how it would be with other browsers).

joelit (approved):
LGTM! Everything looks fine in WAVE, too. I added a very minor fix to CSS.

PR #1720, opened 2024-12-12
Remove the search language cookie

### LINKED ISSUE #1692: Content language setting is not working consistently in Skosmos 3

joelit:

There was unnecessary state management between URL parameters and language cookies for search language. This PR aims to clarify the situation while fixing a minor bug. Supersedes the previous PR #1693 .

- Closes #1692 

- Remove the SKOSMOS_SEARCH_LANG cookie alltogether

None.

## Checklist

### DIFF
--- a/resource/js/vocab-search.js
+++ b/resource/js/vocab-search.js
@@ -143,6 +143,9 @@ const vocabSearch = Vue.createApp({
         if ('uri' in result) { // create relative Skosmos page URL from the search result URI
           result.pageUrl = window.SKOSMOS.vocab + '/' + window.SKOSMOS.lang + '/page?'
           const urlParams = new URLSearchParams({ uri: result.uri })
+          if (this.selectedLanguage !== window.SKOSMOS.lang) { // add content language parameter
+            urlParams.append('clang', this.selectedLanguage)
+          }
           result.pageUrl += urlParams.toString()
         }
         // render search result renderedTypes
--- a/resource/js/vocab-search.js
+++ b/resource/js/vocab-search.js
@@ -204,7 +204,7 @@ const vocabSearch = Vue.createApp({
     showAutoComplete () {
       this.showDropdown = true
       this.$forceUpdate()
-    },
+    }
   },
   template: `
     <div class="d-flex my-auto ms-auto">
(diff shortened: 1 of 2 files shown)


osma (changes requested):
The changes look fine and when testing, I didn't find any problems. However, there is a bug in the search component: it doesn't add the `clang` parameter to links, so the content language setting will be forgotten (and after these changes it won't be remembered by a cookie either). This can be tested the following way:

1. Go to YSO home page with UI language set to English
2. Change content language to Finnish (note: content shows up in Finnish)
3. Search for "koira"
4. Click on the first link "koira"
5. You will see the concept page "dog"; the content language setting was forgotten

This is not a regression as such, because current `main` shows the same behaviour. However, I think it would make sense to fix it in this PR as it's closely related to the search/content language state handling that this PR addresses.

I think the problem is this line: https://github.com/NatLibFi/Skosmos/blob/c43b099f1fa6afca7c6bbf980084c23097afaee1/resource/js/vocab-search.js#L145

Here, the `clang` parameter should be set as well, at least if `clang != lang`.

osma (approved):
LGTM, passing of clang now works as well

PR #1723, opened 2024-12-17
Add translations for tab-alpha.js and vocab-search.js (#1722)

### LINKED ISSUE #1722: Sub of #1712: Add translations for front-end components
In the previous stages, translation functionalities have been implemented in some Vue components. The same must be done for the other Vue components created so far.

- [ ] placeholder="Search..." \@resource/js/vocab-search.js

Skosmos users / developers

The key-value pairs for translations maintained with Symfony must also be available and then in use for the Vue components, so ensuring the functionality of the translations is the goal.

Skosmos is widely used around the world and its user base spans across various language regions. So, the language functionalities are important.

miguelvaara:
## Review guidelines

Visually observable changes in the browser are primarily related to parts generated by the following files:

- resource/js/tab-alpha.js
- resource/js/vocab-search.js

**Test the user interface:**

- Go: [our Skosmos root]/yso/fi/
- Type "pork" into the vocabulary search bar and observe the different concept types.
- Type "gdfgdgfgf" into the vocabulary search bar and observe the _No results_ text.
- Click on a letter in the alphabetical list and observe the _Loading more items_ text.

**Note:**

- The _msgs_ variable has been removed from the SKOSMOS object.
- The issue in the alphabetical list's Cypress test was caused by a timeout. For some reason in my development environment data transfer from the triple store to the frontend was so slow that an "amazingly" long wait time was required. The question remains as to how this "issue" should be properly addressed.

Improves frontend translation functionalityies.

#1722 

- Closes #1722

Added similar features as in the previous Vue-related translation implementations.

What if we explore the ways to implement the translation functionality using JavaScript's promise features. By doing so, translations could be loaded in a lazy manner, meaning that if translations need to be awaited, the translation key would briefly appear to the user (likely some milliseconds), but as soon as the translations are ready, the key would be replaced with the actual translation.

This is a very common way to implement similar functionalities. The current implementation works but it has a weakness: if additional dependencies will be needed later that also need to be awaited before the page loads, they must be added to the same functionality (risk of excessive complexity).

Therefore, using the _async -> await -> response -> .then -> .catch_ structure could be considered.

## Checklist


resource/js/vocab-search.js
```diff
@@ -1,292 +1,275 @@
...
+          return label + langSpec
         }
-        // use the translateType function to map translations for the type IRIs
-        result.renderedType = result.type.map(this.translateType).join(', ')
-        result.showNotation = this.showNotation
-      })
+        return null
+      },
+      translateType (type) {
+        return this.translateTypes(type)
```
UnniKohonen, line 122: The `$t` function could be called directly in `translateType`.

resource/js/vocab-search.js
```diff
@@ -1,292 +1,275 @@
...
+    mounted () {
+      this.languages = window.SKOSMOS.languageOrder
+      this.selectedLanguage = this.parseSearchLang()
+      this.searchCounter = 0 // used for matching the query and the response in case there are many responses
+      this.languageStrings = window.SKOSMOS.language_strings
       this.renderedResultsList = []
+      this.showNotation = window.SKOSMOS.showNotation
+    },
+    methods: {
+      translateTypes (type) {
```
UnniKohonen, line 34: This function can be removed if `translateType` calls `$t` directly

resource/js/tab-alpha.js
```diff
@@ -1,238 +1,253 @@
...
+          </ul>
+        </template>
+      </div>
+    `
+  })
+
+  tabAlphaApp.mount('#tab-alphabetical')
+}
+
+function waitForTranslationService () {
```
UnniKohonen, line 245: This function has the same name as the functions in vocab search and vocab counts. They should all be renamed.

resource/js/tab-alpha.js
```diff
@@ -1,238 +1,253 @@
...
+      this.loadingMessage = this.loadingMoreItemsMessage() // window.SKOSMOS.msgs[window.SKOSMOS.lang]['Loading more items'] ?? window.SKOSMOS.msgs.en['Loading more items']
     },
-    setListStyle () {
-      const pagination = this.$refs.tabAlpha.$refs.pagination
-      const sidebarTabs = document.getElementById('sidebar-tabs')
+    beforeUpdate () {
+      this.setListStyle()
+    },
+    methods: {
+      loadingMoreItemsMessage () {
```
UnniKohonen, line 39: A computed value should be used for fetching the translation instead of a function

resource/js/tab-alpha.js
```diff
@@ -1,238 +1,253 @@
...
-      const listElement = this.$refs.tabAlpha.$refs.list
-      if (listElement.scrollTop + listElement.clientHeight >= listElement.scrollHeight - 1) {
-        this.loadMoreConcepts()
+    mounted () {
+      // load alphabetical index if alphabetical tab is active when the page is first opened (otherwise only load the index when the tab is clicked)
+      if (document.querySelector('#alphabetical > a').classList.contains('active')) {
+        this.loadLetters()
       }
+
+      this.loadingMessage = this.loadingMoreItemsMessage() // window.SKOSMOS.msgs[window.SKOSMOS.lang]['Loading more items'] ?? window.SKOSMOS.msgs.en['Loading more items']
```
UnniKohonen, line 33: This can be removed if a computed value is used for the message

resource/js/tab-alpha.js
```diff
@@ -1,238 +1,253 @@
...
+    data () {
+      return {
+        indexLetters: [],
+        indexConcepts: [],
+        selectedConcept: '',
+        selectedLetter: '',
+        loadingLetters: false,
+        loadingConcepts: false,
+        loadingMoreConcepts: false,
+        loadingMessage: '',
```
UnniKohonen, line 15: `loadingMessage` can be removed from data if a computed value is used

UnniKohonen (changes requested):
There are a few minor chnages I'd make:
- A computed value should be used for the loading message in alphabecal tab component
- The `$t` function could be called directly in the `translateType` function and `translateTypes` could be removed in vocab search component

A bigger issue is that vocab counts, alphabetical tab and vocab search Vue components all define and call the function `waitForTranslationService` which means that only one of the components can be mounted. Each of the components should get its own function for waiting for the translations (like in the term counts component).

resource/js/vocab-search.js
miguelvaara: Removed

resource/js/tab-alpha.js
miguelvaara: Moved out of the global scope to local scope, renamed and in addition, refactored into arrow functions.

resource/js/tab-alpha.js
miguelvaara: Good point! Thanks :-). Fixed!

resource/js/tab-alpha.js
miguelvaara: Fixed

resource/js/tab-alpha.js
miguelvaara: There was something here that I did not fully grasp, but that' s okay. I understand that variables defined in `computed` can be used "directly," but I still think the use of a lifecycle hook would have been necessary. However, it seems to work without it. I did as you suggested, as it works. :-)

osma:
This PR seems to contain commits merged from the main branch (e.g. the PHPUnit 10 upgrade), which make the commit history and GitHub diff view a bit messy. Maybe it would help to switch the base branch briefly to something else and back to `main`, that would at least force GitHub to recompute the diff view, hopefully making it cleaner.

osma:
Now the merge conflicts have been resolved and the unnecessary 20 second waits in Cypress tests have been removed. :tada:

UnniKohonen (approved):
LGTM!

PR #1726, opened 2024-12-17
Issue1724 autocomplete result list

### LINKED ISSUE #1724: Autocomplete result list improvements
* Autocomplete dropdown list needs to be customizable in length, with a scroll bar
* The background should not be white, as it was considered confusing
* The right side border should be in line with the search button

joelit:

Reacting to user feedback

- Closes #1724 

* Search result list background changed to darker color
* Search result list width according to the search field width
* Add a scroll bar if the search result list or the languagage selection list is longer than 50% of view window height

* The layout changes need to be approved

## Checklist


osma (commented):
The changes so far look good, although I'm not sure if the darker background is such a great idea. It looks a bit confusing IMHO.

If I may make an additional request, I think the left margin/padding of search results should be the same as in the search field. Currently they don't align very well, as shown here with a red line:

image

resource/css/skosmos.css
```diff
@@ -204,6 +205,7 @@ body {
   }
 
   #search-wrapper {
+    display: flex;
```
UnniKohonen, line 208: This line seems to be overwritten by something else. If flex is needed for the element, `!important` should be added.

UnniKohonen (changes requested):
There is one line in the CSS that seems to be redundant, otherwise looks good to me!

resource/css/skosmos.css
joelit: Damn, that was the only thing that remained when I was trying out a different layout solution. It was needed then, but it seems it is not needed in the current layout.

PR #1730, opened 2024-12-18
Simplify search page URL construction

### LINKED ISSUE #1703: Search page fails when search is started from a concept page with a ?uri URL parameter

https://test.dev.finto.fi/yso/fi/page?uri=http%3A%2F%2Fwww.yso.fi%2Fonto%2Fyso%2Fp19378

## Description of the problem

1. Go to a concept page with a URL that includes the `uri` parameter, e.g. https://test.dev.finto.fi/yso/fi/page?uri=http%3A%2F%2Fwww.yso.fi%2Fonto%2Fyso%2Fp19378 (you can get there for example from the search autocomplete list)
2. Enter a search term, e.g. "koira", and press enter
3. You will be redirected to the URL `https://test.dev.finto.fi/yso/fi/page?uri=http%3A%2F%2Fwww.yso.fi%2Fonto%2Fyso/fi/search?clang=fi&q=koira` which is nonsensical and doesn't work

The expected outcome would be to show the search result page, e.g. http://localhost/Skosmos/yso/fi/search?clang=fi&q=koira

I think the search box constructs the URL for the search page in a way that doesn't work when the `uri` parameter is used.

Firefox

osma:

When performing a search using the search widget, construction of the URL fails in some cases (see #1703). This PR fixes the bug by simplifying the URL construction, avoiding the use of string replace operations that caused the bug.

- Closes #1703

* simplify search page URL construction, avoid string replace

I didn't make a Cypress test. I figured that since the bug is being removed by mostly removing code, there is no need for a new test. But you could argue otherwise.

## Checklist

### DIFF
--- a/resource/js/vocab-search.js
+++ b/resource/js/vocab-search.js
@@ -172,11 +172,9 @@ const vocabSearch = Vue.createApp({
     gotoSearchPage () {
       if (!this.searchTerm) return
 
-      const currentVocab = window.SKOSMOS.vocab + '/' + window.SKOSMOS.lang + '/'
-      const vocabHref = window.location.href.substring(0, window.location.href.lastIndexOf(window.SKOSMOS.vocab)) + currentVocab
       const searchUrlParams = new URLSearchParams({ clang: window.SKOSMOS.content_lang, q: this.searchTerm })
       if (this.selectedLanguage === 'all') searchUrlParams.set('anylang', 'true')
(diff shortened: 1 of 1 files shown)


joelit (approved):
This is brilliant! I should sometimes slow down and write it simpler - thanks! I agree that this is kind of irrelevant to test in cypress.

PR #1731, opened 2024-12-18
Upgrade to EasyRdf 1.14.0, fixing some mocked resources

### LINKED ISSUE #1700: Update easyrdf from 1.13.1

Sweetrdf/easyrdf has a newer version that is not fully compatible with our code. There's a need for making the update, seeing what gets broken in phpunit, fixing those parts and being paranoid about the code that is not well covered by the tests.

Fixing library dependencies early will be easier than fixing them late.

osma:

We should use recent versions of EasyRdf (see #1700). This PR upgrades EasyRdf from 1.13.1 to 1.14.0, fixing some mock resources in the test suite according to changes in that release.

This isn't the latest version of EasyRdf, but the next release 1.14.1 changes some Accept headers used in SPARQL CONSTRUCT queries. It appears that those changes are not compatible with the (rather old) Jena 4.8.0 version of Fuseki we are using, so going any further probably requires upgrading Fuseki as well. The most recent EasyRdf release currently is 1.15.0.

- Closes #1700 (maybe)

* bump the EasyRdf dependency to 1.14.0
* fix some mocked resources in tests

Is this enough for now, or should we aim for 1.14.1 or 1.15.0 (the current newest release)?

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
   "license": "MIT",
   "version": "3.0-dev",
   "require": {
-    "sweetrdf/easyrdf": "1.13.*",
+    "sweetrdf/easyrdf": "1.14.0",
     "symfony/polyfill-php80": "1.*",
     "symfony/polyfill-php81": "1.*",
     "twig/twig": "3.14.*",
--- a/tests/ConceptMappingPropertyValueTest.php
+++ b/tests/ConceptMappingPropertyValueTest.php
@@ -81,8 +81,8 @@ class ConceptMappingPropertyValueTest extends PHPUnit\Framework\TestCase
         $mocksource = $this->getMockBuilder('EasyRdf\Resource')->disableOriginalConstructor()->getMock();
         $mockres = $this->getMockBuilder('EasyRdf\Resource')->disableOriginalConstructor()->getMock();
         $labelmap = array(
-          array('en', 'english'),
-          array(null, 'default')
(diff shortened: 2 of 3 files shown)


osma:
Rebased on current `main` and force-pushed. Let's see if we can push this further instead of stopping at 1.14.0. We now have a newer Fuseki thanks to #1775 / PR #1778 (Fuseki 5.3.0) and PR #1788 (Fuseki 5.4.0).

osma:
Nope, the tests are still failing for EasyRdf 1.14.1 and above (also tested newest EasyRdf release 1.17.0). It seems there is still a compatibility issue between EasyRdf 1.14.1+ and Fuseki. This should probably be investigated further and reported to EasyRdf.

osma:
The Fuseki compatibility issues have been sorted out in EasyRdf 1.18.0, so I bumped to that version. Also rebased on current `main` and force-pushed.

Sadly, there is another EasyRdf/Fuseki compatibility issue; the Turtle parser fails to parse empty CONSTRUCT query responses. Reported as https://github.com/sweetrdf/easyrdf/issues/74

osma:
There was one more EasyRdf bug, in the RDF/XML parser. It is now fixed and 1.18.2 has been released with the fix.
This PR uses 1.18.2+ and everything seems to be working now.

Issue #1737, opened 2025-01-21
Sub of 1712 translations of Aria tags in tab-alpha


miguelvaara:
## Description of the enhancement
Add translations for ARIA tags.

## Who are the users that would benefit from the enhancement and how?
This benefits Skosmos end-users and through ease of work developers as well.

(Name specific organizations, roles and/or communities if possible, and/or point to relevant discussion elsewhere)

## What new functionalities would the enhancement make possible?
Automatic translations

## Why is the enhancement important?
Translation management must be as simple as possible.

(Why should this be prioritized, instead of spending energy on something else? Who would do the work to make this happen?)

PR #1741, opened 2025-01-23
Tab-alpha's Aria tag translations

miguelvaara:
Enhances accessibility

#1737 

- Closes #
#1737 

Adds dynamic translations for aria tags in different languages

1) Translations for the Sámi language are still pending, but a separate list of missing translations + issue will be created soon to add them later on.

2) **Every now and then**: If Cypress copy-paste errors occire, they can be ignored. The problem is known and does not mean that there is an actual problem on the page.

Something like this:
_1) Concept page, full vs. partial page loads
       concept notation can be copied to clipboard / partial:
      AssertionError: expected '33.10' to equal '33.2'
      + expected - actual
      -'33.10'
      +'33.2'_

3) The Symfony console's language tag extraction changes the order of the keys for some reason or another. However, the extractor must be used and unfortunately it causes somewhat unnecessary scrutiny in the file viewing.

**Review instructions**: Inspect the aria tags of the alphabetical list links using the browser's inspector tool.

## Checklist


resource/translations/messages.sv.json
```diff
@@ -42,6 +42,7 @@
     "Error: Term \"%term%\" not found in vocabulary!": "Fel: Termen \"%term\" finns inte i vokabul\u00e4ren!",
     "Feedback": "Respons",
     "Feedback has been sent!": "Respons har skickats!",
+    "Go to the concept page": "Go till begrepp sidan",
```
osma, line 45:
```suggestion
    "Go to the concept page": "Gå till begreppsidan",
```

resource/translations/messages.se.json
```diff
@@ -40,6 +40,7 @@
     "Error: Term \"%term%\" not found in vocabulary!": "Error: Tearbma \"%term%\" ii g\u00e1vdnon s\u00e1tner\u00e1jus!",
     "Feedback": "M\u00e1hcahat",
     "Feedback has been sent!": "M\u00e1hcahat lea s\u00e1ddejuvvon!",
+    "Go to the concept page": "KÄÄNNETTÄVÄ",
```
osma, line 43: Remove this - I think it's better to leave out translations that aren't known yet

resource/translations/messages.sv.json
miguelvaara: Fixex

tests/cypress/template/sidebar-alpha.cy.js
```diff
@@ -119,4 +119,13 @@ describe('Alphabetical index', () => {
     // check that the second mapping property has the right number of entries
     cy.get('.prop-mapping').eq(0).find('.prop-mapping-label').should('have.length', 3)
   })
+  // Check the correctness of Aria-labels (Sami language will be implemented later)"
+  it('Aria tags are correct for each language', () => {
+    cy.visit('/yso/en/')
+    cy.get('li.list-group-item.py-1.px-2 > a').should('have.attr', 'aria-label', 'Go to the concept page')
```
UnniKohonen, line 125: Maybe this selector (also lines 127 and 129) should be `#tab-alphabetical .list-group-item > a` or something along those lines in case the padding changes at some point.

miguelvaara: Good point, I cchanged it to match those selectors

osma (approved):
LGTM. There are a few new translations that are unrelated to the other contents of this PR, but they are useful to have, so I think they can stay.

PR #1744, opened 2025-01-23
Truncate long search result lines

### LINKED ISSUE #1528: Long search result lines should be truncated

There was a feature in Skosmos 2.x where long search result lines were truncated and toggleable for full view. The feature should be reimplemented in Skosmos 3.x.

### Screenshot of the previous implementation in Skosmos 2.x

Screenshot from 2023-09-27 12-28-17

joelit:

Closes #1528 

Implementation for truncating long search results

## Checklist

### DIFF
--- a/resource/js/truncate-search-results.js
+++ b/resource/js/truncate-search-results.js
@@ -11,13 +11,12 @@ function truncateSearchResults () {
       // if the element does not have a show all -link, add one
       const lastElement = result.lastElementChild
       if (!lastElement || lastElement.tagName !== 'A') {
-        const link = document.createElement('A');
-        link.textContent = renderShowAllText(result);
-        link.setAttribute("class", "p-0 ps-4 search-result-hide")
-        const backgroundColor = window.getComputedStyle(result).backgroundColor;
+        const link = document.createElement('A')
+        link.textContent = renderShowAllText(result)
+        link.setAttribute('class', 'p-0 ps-4 search-result-hide')
 
-        link.onclick = () => showAllResults(result);
-        result.appendChild(link);
+        link.onclick = () => showAllResults(result)
+        result.appendChild(link)
       }
     } else {
       removeShowAll(result)
--- a/resource/js/truncate-search-results.js
+++ b/resource/js/truncate-search-results.js
@@ -15,9 +15,18 @@ function truncateSearchResults () {
         link.style.position = 'absolute'
         link.style.right = '0'
         link.style.padding = '0'
+        link.style.paddingLeft = '1.5em'
         link.textContent = renderShowAllText(result)
         const backgroundColor = window.getComputedStyle(result).backgroundColor
-        link.style.backgroundColor = backgroundColor
+        const cssString = `background:
+                             linear-gradient(
+                               90deg, transparent,
+                               ${backgroundColor},
+                               ${backgroundColor});`
+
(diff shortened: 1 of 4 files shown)


osma (changes requested):
Looks good! A few comments:

# Visual style

There is no padding on the left side of the ellipsis (...) and sometimes this leads to visual artefacts that look a bit broken:

image

It would be better to have some left padding, as in Skosmos 2, so that the ellipsis doesn't visually merge with the text.

Even better would be some kind of transparent fade-out effect, as in this picture (drawn using Gimp):

image

# Pointer style

I noticed that the mouse pointer becomes a cursor (vertical bar) when hovering above the link, not a pointer (hand with finger) as I'd expect. I think this is some CSS rule that applies to the whole search result list, because the same happens when hovering over prefLabels in other languages (e.g. "dog parks" in above pictures). I think the correct thing to do here would be to show a regular arrow cursor over "dog parks" (it's not clickable after all), and a pointer cursor when hovering over the "click for more" link.

# Tooltip

I think it would be useful to show a tooltip (just a plain link title would do) that explains what happens if you click on the link. For example "expand to show all items" or something along those lines.

# Use of `.style` attribute

I noticed that the JS code sets element styles directly using the `.style` attribute. I think it would be better to set a specific class for the link elements and define the styles using CSS instead - that's easier to maintain in the future. In some cases it might be helpful to use Bootstrap-defined styling classes such as `p-0` for `padding: 0`.

joelit:
I implemented all suggestions except the tooltip text. Is the translation infrastructure ready for it in case of plain JS functions, or should I implement a placeholder text?

osma:
There are now some JS problems that cause CI tests to fail. SonarCloud also complains about an unused JS variable.

joelit:
> There are now some JS problems that cause CI tests to fail. SonarCloud also complains about an unused JS variable.

Ah, sorry. Fixed!

joelit:
The tooltip for "show more" is a good idea, but it really wants a ready and documented translation infrastructure for vanilla JS-components. I will note it in the relevant epic-issue.

resource/js/truncate-search-results.js
```diff
@@ -0,0 +1,50 @@
+function truncateSearchResults () {
+  const results = document.querySelectorAll('.search-result .list-group .list-group-item')
+  results.forEach((result) => {
+    result.style.whiteSpace = 'nowrap'
```
osma, line 4: Can these two `.style` settings be moved to the CSS file as well?

resource/js/truncate-search-results.js
```diff
@@ -0,0 +1,50 @@
...
+
+function removeShowAll (element) {
+  const lastElement = element.lastElementChild
+  if (lastElement && lastElement.tagName === 'A') {
+    element.removeChild(lastElement)
+  }
+}
+
+function showAllResults (element) {
+  element.style.whiteSpace = 'normal'
```
osma, line 35: Ditto, maybe move to CSS file?

osma (changes requested):
There are still some `.style` settings that could perhaps be moved to CSS, see comments.

The fade-out effect looks quite good but I think there's too much overlap between the ellipsis (...) and the fading text so the result looks a bit messy:

image

I think the text should fade out completely before the ellipsis, maybe with a few pixels of empty (white) space in between the two.

PR #1742, opened 2025-01-23
[WIP] Responsive layout for the concept page

osma:
## Reasons for creating this PR

The concept page renders badly in a narrow browser window - see #1739.
This PR contains fixes to make it more responsive. The main target is to make the concept page render sensibly in a half-width browser window, typically around 900px wide.

Here's how this looks currently in a 900px wide Firefox window:

image

## Link to relevant issue(s), if any

- Closes #1739

## Description of the changes in this PR

* avoid grey margins on the sides that take up extra space
* make property/value display on the concept page responsive (it will switch to another layout at the `lg` breakpoint i.e. below 992px)

## Known problems or uncertainties in this PR

n/a

## Checklist

- [x] phpUnit tests pass locally with my changes
- [x] I have added tests that show that the new code works, or tests are not relevant for this PR (e.g. only HTML/CSS changes)
- [x] The PR doesn't reduce accessibility of the front-end code (e.g. tab focus, scaling to different resolutions, use of `.sr-only` class, color contrast)
- [x] The PR doesn't introduce unintended code changes (e.g. empty lines or useless reindentation)

Issue #1739, opened 2025-01-21
Check responsiveness of the concept page with half-width browser window


joelit:
A reported requirement spec was that Skosmos layout should work with two browser windows side-by-side. Currently some of the UI elements get stacked horizontally in this case, which is not prefered behaviour. Skosmos layout elements should stack horizontally in very narrow layouts, e.g. for mobile screens.

Issue #1684, opened 2024-10-01
CSS Improvements for Vocabulary Front Page


joelit:
## Description of the enhancement

 - [ ] The footer should be white and it should line up with the main content
 - [x] The search bar should be followed by a wide margin
 - [ ] The search tips are missing
 - [x] The search bar background should be white
 - [ ] The Footer text should line up with the text of the main content
 - [x] The page margins should be smaller
 - [ ] The search bar margins should line up with the main content
 - [x] The Skosmos logo should be larger
 - [ ] The UI elements should line up in the bottom of the layout
 - [ ] The vocab statistic components should follow the visual spec
 - [x] The space between the columns should be wider

Issue #1753, opened 2025-01-29
Epic: Mobile Layout


miguelvaara:
## What new functionalities would the enhancement make possible?

- [ ] 1 Landing page
  - [ ] 1.1 Font sizes are too large
- [ ] 2 Vocab home page
  - [ ] 2.1 Vocab title and vocab search bar should on top of each other
  - [ ] 2.2. Vocab title font size is too large
  - [ ] 2.3 Vocab heading ("Vocabulary information") font size is too large and margins/padding is too large
  - [ ] 2.4 Property label and property text font size is slightly too large
  - [ ] 2.5 Border between properties should be closer to property labels
  - [ ] 2.6 Property text overflow should be handled correctly
  - [ ] 2.7 Vocab statistics need a new mobile layout
- [ ] 3 Concept page
  - [ ] 3.1 Vocab title and vocab search bar should on top of each other
  - [ ] 3.2. Vocab title font size is too large
  - [ ] 3.3 Breadcrumb font size is too large
  - [ ] 3.4 Concept label and copy button font is too large
  - [ ] 3.5 Property label and property text font size is slightly too large
  - [ ] 3.6 Border between properties should be closer to property labels
  - [ ] 3.7 "In other languages" property should be aligned better
  - [ ] 3.8 Mappings need a new mobile layout
- [ ] 4 About page
  - [ ] 4.1 Responsive margins/padding
- [ ] 5 Feedback page
  - [ ] 5.1 Responsive margins/padding
  - [ ] 5.2 Submit button and required fields explanation should be aligned correctly
- [ ] 6 Error page
  - [ ] 6.1 Responsive margins/padding
- [ ] 7 Search result pages
  - [ ] 7.1 Responsive margins/padding
  - [ ] 7.2 Sidebar should not be visible
- [ ] 8 Header & footer
  - [ ] 8.1 Hamburger menu for languages and about/feedback pages without icons
  - [ ] 8.2 Skosmos logo is too small
  - [ ] 8.3 Footer columns should be aligned correctly
  - [ ] 8.4 Global search should fill the whole screen
  - [ ] 8.5 Global search fields should be on top of each other
  - [ ] 8.6 Global search close button should be in the top right corner
- [ ] 9 Sidebar
   - [ ] 9.1 Should not be visible at all, instead sidebar view selection is on top of concept card
- [ ] 10

PR #1754, opened 2025-01-29
Responsive layout for the vocabulary home page + layout adjustments

osma:

The vocabulary home page renders badly in a narrow browser window. This was noted in #1739 which was fixed in PR #1742, but only for the concept page. This PR does a similar adjustment to the vocabulary home page.

Additionally, this adjusts the gutters, padding and margins for both the vocabulary home page and the concept page so that they better match each other and the visual specification.

Here's how the vocabulary home page looks in a narrowish (~900px) browser window:

image

- part of #1684 

- make the grey gutters between columns a bit larger, and the grey bar above main content a bit narrower, trying to match the visual spec (but this depends a lot on the browser; I mainly used Firefox)
- make property/value columns on the vocabulary home page responsive
- various small margin/padding/gutter adjustments, mainly for the vocabulary home page but also some for the concept page

I hope I didn't break layouts on other page types.

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -717,11 +717,18 @@ body {
   .main-content-section {
     background-color: var(--vocab-bg);
     margin-bottom: .2rem;
+    padding-left: 3.5rem;
+    padding-right: 3.5rem;
   }
 
   #concept-heading, #vocab-heading {
     margin-top: 2rem;
     padding-bottom: 2.5rem;
+  }
+
+  #vocab-heading, .vocab-statistics, #vocab-download-links {
+    margin-left: -.75rem;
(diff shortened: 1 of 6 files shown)


joelit:
LGTM! This fixes the layout on the 767...992px range. It would be a good idea to create specs for the mobile layout (sub 767px) layout to make sure the logic stays consistent. I take it that the idea is to use sm, md & lg breakpoints for the responsive layout?

osma:
@joelit Yes, the mobile layout needs further specification. I think most of it can be implemented using the breakpoints you mention. We have a separate epic for it now: #1753 (just a placeholder currently)

PR #1756, opened 2025-01-29
Reorganized the usage of the main element in vocabulary search page

joelit:

Vocabulary search page was mis-using the main element, causing there to be a duplicate structure:
image

 * Brought the structure of vocabulary search page more in line with vocabulary front page and concept page
 * Removed references to global search page for now

None

## Checklist


src/view/base-template.twig
```diff
@@ -93,7 +93,7 @@
     </div>
     {% endif %}
   </header>
-  <main id="main-container" class="{% if request.vocabid == '' or global_search %} frontpage{% elseif concept_uri != '' %} termpage{% elseif request.vocab != '' %} vocabpage{% if list_style %} {{ list_style }}{% endif %}{% endif %} py-5">
+  <main id="main-container" class="{% if request.vocabid == '' or global_search %}frontpage{% elseif concept_uri != '' %}termpage{% elseif request.vocab != '' %}{% if request.page == 'search'%}searchpage{% else %}vocabpage{% endif %}{% if list_style %} {{ list_style }}{% endif %}{% endif %} py-5">
```
osma, line 96: Oof. This could really use some refactoring... We already have a Twig variable `pageType` that I think would be a much better source of information for this, and perhaps it could be used directly as a CSS class name (adding some prefix). Oh well, maybe that's out of scope for this PR...

src/view/vocab-search.twig
```diff
@@ -4,7 +4,7 @@
 {% block url %}{{ BaseHref }}{{ request.langurl }}{% endblock %}
 
 {% block content %}
-<main class="searchpage py-5" id="main-container" >
+<div py-5">
```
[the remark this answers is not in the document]
joelit: OK, I'll check what is the needed padding for this, and amend it to the spacing of a nested element.

src/view/base-template.twig
joelit: That might be worth a shot, I'll have a look as generating the class name.

src/view/concept-card.inc.twig
```diff
@@ -1,4 +1,4 @@
-<div class="col-md-8" id="main-content">
+<div class="col-md-8 ps-3" id="main-content">
```
osma, line 1: Please adjust the gutter instead here: https://github.com/NatLibFi/Skosmos/blame/53aa787e6269859f5cb8739ea327ced0a2322453/resource/css/skosmos.css#L359

src/view/sidebar.inc.twig
```diff
@@ -1,4 +1,4 @@
-<div class="col-md-4 position-relative" id="sidebar-col">
+<div class="col-md-4 pe-3 position-relative" id="sidebar-col">
```
osma, line 1: ditto

src/view/vocab-info.inc.twig
```diff
@@ -1,7 +1,6 @@
-<div class="col-md-8" id="main-content">
+<div class="col-md-8 ps-3" id="main-content">
```
osma, line 1: ditto

src/view/base-template.twig
```diff
@@ -93,7 +93,7 @@
     </div>
     {% endif %}
   </header>
-  <main id="main-container" class="{% if request.vocabid == '' or global_search %} frontpage{% elseif concept_uri != '' %} termpage{% elseif request.vocab != '' %} vocabpage{% if list_style %} {{ list_style }}{% endif %}{% endif %}">
+  <main id="main-container" class="{{pageType}}{% if list_style %} {{ list_style }}{% endif %}">
```
osma, line 96: This gets rid of the old and IMHO misleading page type classes `frontpage`, `termpage` and `vocabpage` , which is great! :tada: 

However, there are still many occurrences of these classes in `skosmos.css` that should be changed or removed.

PR #1758, opened 2025-01-30
Global Search twig-template

### LINKED ISSUE #1745: Global Search twig-template
Part of the #1481 as the first implementation for the epic.

- [x] 5. The page title is displayed comparable to "'kissa' - Finto"
  - [x]  5.1 Information from the page title is picked up for SEO metadata
- [x] 6. AMOUNT results for TERM text with a horizontal separator on the top of the list (and translated)
- [x] 7. Search returns correct number of hits
- [x] 8. Verify the properties and their values in one search hit
  - [x] 8.1 prefLabel
- [x] 8.2 broaders
- [x] 8.3 narrowers
- [x] 8.4 reltatedMatches
- [x] 8.5 altLabels
- [x] 8.6 memberOf
- [x] 8.7 foreignLabels
- [x] 8.8 type
- [x] 8.9 URI
- [x] 9. prefLabels in foreign languages are displayed with their language tags in parentheses
- [x] 10. Icons for properties are displayed
- [x] 13. Long lists are truncated and end in a link that displays the value count for this property list per #1744

joelit:

- Closes #1745 

* This PR implements the twig template for global search
* In the absence of a search field, the result page can be tested by visiting `[Skosmos-home]/en/search?clang=en&q=kissa`

* The PR would greatly benefit from being rebased on PRs #1744  and #1756 after they have been merged

## Checklist

- [X] The PR doesn't reduce accessibility of the front-end code (e.g. tab focus, scaling to different resolutions, use of `.sr-only` class, color contrast)
- [X] The PR doesn't introduce unintended code changes (e.g. empty lines or useless reindentation)

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -88,7 +88,7 @@ body {
   #main-container.global-search .list-group .fa-solid, #main-container.global-search .list-group .fa-regular {
     color: var(--vocab-text);
     font-size: 1.15rem;
-    width: 20px;
+    width: 1.5rem;
     text-align: center;
   }
 
@@ -191,8 +191,8 @@ body {
     font-family: var(--font-family-heading);
   }
 
-  #search-multi-vocab {
-    width: 800px;
+  #global-search-wrapper {
+    max-width: 1200px;
   }
 
   #search-wrapper {
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -79,12 +79,12 @@ body {
     margin: 0 5px;
   }
 
-  #main-container.searchpage .fa-arrow-right, #main-container.global-search .fa-arrow-right {
+  #main-container.vocab-search .fa-arrow-right, #main-container.global-search .fa-arrow-right {
     color: var(--vocab-text);
     font-size: .85rem;
   }
 
-  #main-container.searchpage .list-group .fa-solid, #main-container.searchpage .list-group .fa-regular,
+  #main-container.vocab-search .list-group .fa-solid, #main-container.vocab-search .list-group .fa-regular,
   #main-container.global-search .list-group .fa-solid, #main-container.global-search .list-group .fa-regular {
     color: var(--vocab-text);
     font-size: 1.15rem;
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -79,12 +79,13 @@ body {
     margin: 0 5px;
   }
 
-  #main-container.vocab-search  .fa-arrow-right {
+  #main-container.searchpage .fa-arrow-right, #main-container.global-search .fa-arrow-right {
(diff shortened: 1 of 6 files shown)


src/view/global-search.twig
```diff
@@ -0,0 +1,21 @@
+{% set pageType = 'global-search' %}
+{% extends "base-template.twig" %}
+{% block title %}'{{ term }}' - {{ "Search results" | trans }} - {{ GlobalConfig.serviceName }}{% endblock %}
```
osma, line 3: According to the current [SEO spec for page titles and descriptions](https://github.com/NatLibFi/Skosmos/issues/1533#issuecomment-2063290242), the middle part "Search results" should be dropped. So the title should include just the search term and the service name.

tests/cypress/template/global-search.cy.js
```diff
@@ -1,3 +1,57 @@
 describe('Global search page', () => {
-  it('no-op test', () => {})
+  const term = 'bass';
+  it('Contains title and title metadata', () => {
+      cy.visit(`/en/search?clang=en&q=${term}`)
+
+      const expectedTitle = "'bass' - Search results - Skosmos being tested"
```
osma, line 6: This of course needs to be changed as well if you change the title in the template above, as suggested.

osma (changes requested):
I gave a few suggestions for minor code adjustments.

Looking at the search result page, I see several layout issues. The spacing between the icons and the text is sometimes nonexistent, and there are sometimes extra spaces before commas:

image

But this is not consistent as sometimes the result items are rendered just fine:

image

joelit:
OK, the inconsistency of the search results for search result property label spacing and property value spacing is pretty weird, and probably has further implications for the consistency of the twig template. I'll track this down to the root cause.

osma (changes requested):
A couple of issues still need fixing:

1. The spacing between icons and text on the search results page is still a bit erratic
2. SonarCloud complains about unnecessary escape characters

UnniKohonen (changes requested):
Otherwise looks good to me, but there seems to be one unused CSS rule.

Also, IMO, the fontsize of preflabel links is too small, compare this PR (1) to the mockup layout (2). The same issue is also present on the vocab search result page.

1 image

2 image

resource/css/skosmos.css
```diff
@@ -190,8 +191,8 @@ body {
     font-family: var(--font-family-heading);
   }
 
-  #search-multi-vocab {
-    width: 800px;
+  #global-search-wrapper {
```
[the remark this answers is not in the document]
joelit: Ah, #global-search-wrapper is part of the search field. I will move it to the PR that has the rest of the definitions for the global search field.

UnniKohonen (approved):
LGTM!

PR #1759, opened 2025-01-30
Fix scrollbar styling and visibility

### LINKED ISSUE #1738: Force scrollbar visibility for autocomplete drowdown

The web browsers get to decide how to display a scrollbar when using Bootstrap 5 overflow class. This should be controlled more, so that the browser would at least show the scroll bar in the cases where a scroll bar is intended.

Chrome on Mac is one case where the scrollbar is not shown unless the dropdown is actually scrolled.

UnniKohonen:

Scrollbars are not visible by default in browsers on Mac in Skosmos 3. This PR adds a permanent scrollbar to sidebar and searchbar in Webkit based browsers.

- Closes #1738

- Add scrollbar styling for searchbar in Webkit browsers
- Add separate scrollbar styling in Firefox for searchbar and sidebar

Styling seems to not be applied at all in Firefox on Windows 11. In Firefox on Mac, the scrollbar seems to only be visible when elements are scrolled. Though, fixing the issue on Mac does not seem to be currently possible.

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -292,6 +292,7 @@ body {
     }
   } 
 
+  /* Setting scrollbar style for Webkit based browsers (Chrome, Safari, Edge, etc) */
   #search-wrapper ul::-webkit-scrollbar {
     -webkit-appearance: none;
     -moz-appearance: none;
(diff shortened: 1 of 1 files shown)


joelit:
LGTM!

- I verified that the CSS settings work on both Chrome and Firefox
- I also verified that while it is possible to set the scrollbar thumb and track colours in Firefox, it is not possible to set the scrollbar thumb on focus colour
- I also verified that while it is possible to change the width of the scrollbar on Firefox, the options are thin, thinner, invisible - and we have the thickest thin possible.

joelit (approved):

PR #1772, opened 2025-03-12
implement event listener for notifying when translations are ready

### LINKED ISSUE #1771: Add callback notification facility to the JS translation service
Part of #1712 

The JS frontend implements a simple [translation service](https://github.com/NatLibFi/Skosmos/blob/main/resource/js/translation-service.js). It loads the UI translations from a JSON file and provides a `$t` function for translating messages to the UI language.

Currently, Vue and other JS frontend components need to implement a special looping function that waits for the translation service to become available before initializing the UI component. This is duplicated boilerplate, and a bit ugly because it relies on polling.

Instead, the translation service could provide a facility for registering callbacks that need to be notified after the translation service has been initialized and the `$t` function is available. UI components could each register their own callback function (that initializes the component) to be called when the translation service is ready, instead of polling and waiting.

It would simplify the JS frontend code and make the UI a bit more responsive (because we could drop the polling), so all Skosmos users but especially developers would benefit.

It would simplify JS component code and remove inefficient boilerplate code.

This is the right time to do it because we are implementing the translation functionality.

osma:

Smarter event handling for frontend components that need to be initialized only after the translation service is ready.

See #1771

- Closes #1771

- add onTranslationReady function for registering event handlers that should be called when translations are ready
- use onTranslationReady in Vue components instead of stupid polling boilerplate code

n/a

## Checklist

### DIFF
--- a/resource/js/tab-alpha.js
+++ b/resource/js/tab-alpha.js
@@ -1,4 +1,4 @@
-/* global Vue, $t */
+/* global Vue, $t, onTranslationReady */
 /* global partialPageLoad, getConceptURL, fetchWithAbort */
 
 function startAlphaApp () {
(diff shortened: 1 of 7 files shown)


osma:
Hmm, I think there's a race condition that still needs to be fixed.

UnniKohonen (approved):
LGTM!

osma:
I added a check for the case when callbacks are registered after the translation service is initialized, and made sure those are called immediately. I tested it by making a 1000ms delayed call to onTranslationReady and it seemed to work. AFAICT, this special case can't happen with the current onTranslationReady calls we have, but it might happen later, so it's better to be able to handle that already.

All done, merging.

Issue #1489, opened 2023-08-24
New search bar for Skosmos 3.0


joelit:
## Description of the enhancement

Skosmos 3.0 needs a new search bar component, possibly using Vue
 * The search bar should be toggled hidden, to look [like this](https://github.com/NatLibFi/Skosmos/blob/0f0a582f1afd1e9cbc1bf98cdbbb8c8090d5af8c/frontpage.html#L26)
 * When revealed, the search bar should look different depending on the page type:
   * On vocabulary page, concept page and vocabulary search page, the search bar should look [like this](https://github.com/NatLibFi/Skosmos/blob/0f0a582f1afd1e9cbc1bf98cdbbb8c8090d5af8c/vocabpage.html#L40)
   * On landing page and the global search page, the search bar should look [like this](https://github.com/NatLibFi/Skosmos/blob/0f0a582f1afd1e9cbc1bf98cdbbb8c8090d5af8c/searchpage-multi-vocab.html#L40)

PR #1779, opened 2025-03-18
Issue1489 global search bar

joelit:

- Resolves some parts of #1489 

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -204,6 +204,7 @@ body {
   #global-search-toggle {
     background-color: var(--headerbar-vocab-bg);
     border-style: none;
+    white-space: nowrap;
   }
 
   #global-search-toggle.collapsed::after {
@@ -228,13 +229,29 @@ body {
     display: inline-block;
   }
 
-  #search-multi-vocab-title {
-    color: var(--headerbar-text-2);
-    font-family: var(--font-family-heading);
+  #global-search-text {
+    min-width: 400px;
+    overflow: auto;
   }
 
-  #search-multi-vocab {
-    width: 800px;
+  #global-search-text h2 {
+    font-family: var(--font-family-heading);
+    font-weight: bold;
+    font-size: 2rem;
+    padding: 1rem;
+  }
+
+  #global-search-bar {
+    flex: 0 0 700px;
+    max-width: 700px;
+  }
+
+  #global-search-bar div {
+    display: inline-block;
+  }
+
+  #global-search-bar .dropdown {
+    display: inline-block;
   }
 
   #search-wrapper {
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -7,6 +7,7 @@
   --light-color: #f2f5f7;
   --alert-color: #a0270a;
   --white-color: #ffffff;
+  --header-bar-text: var(--white-color);
   --feedback-form-bg: var(--white-color);
   --feedback-form-text: var(--dark-color);
   --feedback-form-input: var(--light-color);
@@ -39,6 +40,12 @@
   --font-size: 14px;
   --font-family: 'Public Sans', sans-serif;
   --font-family-heading: 'EB Garamond', serif;
+}
+
+@font-face {
+  font-family: 'Font Awesome 5 Solid';
+  font-style: normal;
+  src: url('../fontawesome/webfonts/fa-solid-900.ttf') format('truetype');
 }
 
 body {
@@ -172,7 +179,7 @@ body {
   /* Headerbar
    *****************************************/
   #headerbar.landing, #headerbar.landing-one-vocab {
-    color: var(--headerbar-text-1);
+    color: var(--headerbar-text);
   }
 
   #headerbar.vocab-home, #headerbar.concept {
@@ -192,6 +199,33 @@ body {
     color: var(--headerbar-text-2);
     font-family: var(--font-family-heading);
     text-decoration: none;
+  }
+
+  #global-search-toggle {
+    background-color: var(--headerbar-vocab-bg);
+    border-style: none;
+  }
+
+  #global-search-toggle.collapsed::after {
+    background-image: none;
+    transform: none;
+    padding-left: 0.5rem;
+
+    content: "\f078"; /* fa-chevron-down */
+    color: var(--header-bar-text);
+    font-family: "Font Awesome 5 Solid";
+    display: inline-block;
+  }
+
+  #global-search-toggle::after {
+    background-image: none;
+    transform: none;
+    padding-left: 0.5rem;
+
+    content: "\f077"; /* fa-chevron-up */
+    color: var(--header-bar-text);
+    font-family: "Font Awesome 5 Solid";
+    display: inline-block;
   }
 
   #search-multi-vocab-title {
--- a/resource/js/global-search.js
+++ b/resource/js/global-search.js
@@ -1,0 +1,407 @@
+/* global Vue, $t, onTranslationReady */
+
+function startGlobalSearchApp () {
+  const globalSearch = Vue.createApp({
+    data () {
+      return {
+        languages: [],
+        selectedLanguage: null,
+        searchTerm: null,
+        searchCounter: null,
+        renderedResultsList: [],
+        languageStrings: null,
+        vocabStrings: null,
+        showDropdown: false,
+        showNotation: null
+      }
+    },
+    computed: {
+      searchPlaceholder () {
+        return $t('3. Kirjoita hakutermi')
+      },
+      noResults () {
+        return $t('No results')
+      },
+      selectSearchLanguageAriaMessage () {
+        return $t('Select search language')
+      },
+      textInputWithDropdownButtonAriaMessage () {
+        return $t('Text input with dropdown button')
+      },
+      searchAriaMessage () {
+        return $t('Search')
+      }
+    },
+    mounted () {
+      this.languages = window.SKOSMOS.languageOrder
+      this.selectedLanguage = this.parseSearchLang()
+      this.searchCounter = 0 // used for matching the query and the response in case there are many responses
+      this.languageStrings = ["Suomi", "Ruotsi", "Englanti"]
+      this.vocabStrings = ["YSO", "YKL", "KOKO"]
+      this.renderedResultsList = []
+      this.showNotation = true
+    },
+    methods: {
+      autoComplete () {
+        const delayMs = 300
+
+        // when new autocomplete is fired, empty the previous result
+        this.renderedResultsList = []
+
+        // cancel the timer for upcoming API call
+        clearTimeout(this._timerId)
+        this.hideAutoComplete()
+
+        // TODO: if the search term is in cache, use the cache
+
+        // delay call, but don't execute if the search term is not at least two characters
+        if (this.searchTerm.length > 1) {
+          this._timerId = setTimeout(() => { this.search() }, delayMs)
+        }
+      },
+      search () {
+        const mySearchCounter = this.searchCounter + 1 // make sure we can identify this search later in case of several ongoing searches
+        this.searchCounter = mySearchCounter
+        let skosmosSearchUrl = 'rest/v1/search?'
+        const skosmosSearchUrlParams = new URLSearchParams({ query: this.formatSearchTerm(), unique: true })
+        if (this.selectedLanguage !== 'all') skosmosSearchUrlParams.set('lang', this.selectedLanguage)
+        skosmosSearchUrl += skosmosSearchUrlParams.toString()
+
+        fetch(skosmosSearchUrl)
+          .then(data => data.json())
+          .then(data => {
+            if (mySearchCounter === this.searchCounter) {
+              this.renderedResultsList = data.results // update results (update cache if it is implemented)
+              this.renderResults() // render after the fetch has finished
+            }
+          })
+      },
+      formatSearchTerm () {
+        if (this.searchTerm.includes('*')) { return this.searchTerm }
+        return this.searchTerm + '*'
+      },
+      notationMatches (searchTerm, notation) {
+        if (notation && notation.toLowerCase().includes(searchTerm.toLowerCase())) {
+          return true
+        }
+        return false
+      },
+      parseSearchLang () {
+        // if content language can be found from uri params, use that and update it to SKOSMOS object and to search lang cookie
+        const urlParams = new URLSearchParams(window.location.search)
+        const paramLang = urlParams.get('clang')
+        const anyLang = urlParams.get('anylang')
+        if (anyLang) {
+          this.changeLang('all')
+          return 'all'
+        }
+        if (paramLang) {
+          this.changeLang(paramLang)
+          return paramLang
+        }
+        // otherwise pick content lang from SKOSMOS object (it should always exist)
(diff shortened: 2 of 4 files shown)


src/view/scripts.inc.twig
```diff
@@ -30,6 +30,13 @@ window.SKOSMOS = {
   "lang": "{{ request.lang }}",
   "vocab": "{{ request.vocabid }}",
   "pageType": "{{ pageType }}",
+  "vocab_list": {
+  {% for key, vocabs in request.vocabList %}
+    {% for key2, value in vocabs %}
+      "{{ value.resource.localname }}": "{{ value.config.shortname }}",
+    {% endfor %}
+  {% endfor %} },
+  "vocab_id": [ {% for vocab in request.vocabid %} "{{ vocab }}", {% endfor %}],
```
UnniKohonen: The comma after the last element is causing partial page load to fail.

src/view/scripts.inc.twig
```diff
@@ -30,6 +30,13 @@ window.SKOSMOS = {
   "lang": "{{ request.lang }}",
   "vocab": "{{ request.vocabid }}",
   "pageType": "{{ pageType }}",
+  "vocab_list": {
+  {% for key, vocabs in request.vocabList %}
+    {% for key2, value in vocabs %}
+      "{{ value.resource.localname }}": "{{ value.config.shortname }}",
+    {% endfor %}
+  {% endfor %} },
```
UnniKohonen: The comma after the last property is causing partial page load to fail.

resource/css/skosmos.css
```diff
@@ -192,45 +199,89 @@ body {
...
+  }
+
+  #global-search-toggle::after {
+    background-image: none;
+    transform: none;
+    padding-left: 0.5rem;
+
+    content: "\f077"; /* fa-chevron-up */
+    color: var(--header-bar-text);
+    font-family: "Font Awesome 5 Solid", sans-serif;
```
UnniKohonen: This should be changed to `font: var(--fa-font-solid);`. If so, font definition on line 45 can be removed.

resource/css/skosmos.css
```diff
@@ -192,45 +199,89 @@ body {
...
+  }
+
+  #global-search-toggle.collapsed::after {
+    background-image: none;
+    transform: none;
+    padding-left: 0.5rem;
+
+    content: "\f078"; /* fa-chevron-down */
+    color: var(--header-bar-text);
+    font-family: "Font Awesome 5 Solid", sans-serif;
```
UnniKohonen: This should be changed to `font: var(--fa-font-solid);`. If so, font definition on line 45 can be removed.

resource/css/skosmos.css
```diff
@@ -41,6 +42,12 @@
   --font-family-heading: 'EB Garamond', serif;
 }
 
+@font-face {
```
UnniKohonen: If lines 215 and 226 are changed, this font definition can be removed

resource/css/skosmos.css
```diff
@@ -266,6 +317,7 @@ body {
   #search-wrapper .dropdown-menu {
     transform: translateY(-1px);
     background-color: var(--search-dropdown-bg);
+    --bs-dropdown-link-active-bg: var(--search-button-bg);
```
UnniKohonen: I feel like bootstrap variables shouldn't be overridden (not sure though, maybe it's OK?) I would create a separate CSS rule that changes the background color in the right context.

resource/js/global-search.js
```diff
@@ -0,0 +1,427 @@
...
+        if (!this.searchTerm) return
+
+        const searchUrlParams = this.formatSearchUrlParams()
+        const searchUrl = window.SKOSMOS.baseHref + window.SKOSMOS.lang + '/search?' + searchUrlParams.toString()
+
+        window.location.href = searchUrl
+      },
+      changeLang (clang) {
+        this.selectedLanguage = clang
+        window.SKOSMOS.content_lang = clang
```
UnniKohonen: I feel like the SKOSMOS object should be read-only. It doesn't seem like this is affecting the rest of the component though, so it could probably be removed.

UnniKohonen (changes requested):
General layout/usability notes:
- All vocabularies should be the first option in the vocab dropdown menu
- There should maybe be some padding/margins inside the vocab dropdown menu
- The button closing the search bar is not implemented
- Checkboxes should have custom styling
- Checking/unchecking vocabs is annoying because the dropdown moves
- There is some lag in the animation for me when opening/closing the menu

resource/css/skosmos.css
```diff
@@ -192,49 +196,188 @@ body {
...
+    padding-right: 2rem;
+  }
+
+  #search-wrapper .vocab-dropdown-btn span {
+    display: block;
+    overflow: hidden;
+    z-index: 0;
+  }
+
+  #search-wrapper .vocab-dropdown-btn span::after {
```
UnniKohonen: This does not seem to do anything

resource/js/global-search.js
```diff
@@ -0,0 +1,428 @@
...
+    },
+    computed: {
+      vocabSelectorPlaceholder () {
+        return $t('1. Valitse sanasto')
+      },
+      langSelectorPlaceholder () {
+        return $t('2. Valitse kieli')
+      },
+      searchPlaceholder () {
+        return $t('3. Kirjoita hakutermi')
```
UnniKohonen: Search filed placeholder does not have a translation, maybe it should be added to Lokalise as part of this PR?

resource/js/global-search.js
```diff
@@ -0,0 +1,428 @@
...
+      textInputWithDropdownButtonAriaMessage () {
+        return $t('Text input with dropdown button')
+      },
+      searchAriaMessage () {
+        return $t('Search')
+      },
+      getSelectedVocabs () {
+        return this.selectedVocabs.map(key => ({ key, value: this.vocabStrings[key] }))
+      },
+      searchLanguage () {
```
UnniKohonen: `searchLanguage` computed value is unnecessary

resource/js/global-search.js
```diff
@@ -0,0 +1,428 @@
...
+            if (mySearchCounter === this.searchCounter) {
+              this.renderedResultsList = data.results // update results (update cache if it is implemented)
+              this.renderResults() // render after the fetch has finished
+            }
+          })
+      },
+      formatSearchUrlParams () {
+        const params = new URLSearchParams({ query: this.formatSearchTerm(), unique: true })
+        if (this.searchLanguage !== 'all') {
+          params.set('clang', this.searchLanguage)
```
UnniKohonen: Should use `this.selectedLanguage` instead of `this.searchLanguage`

resource/js/global-search.js
```diff
@@ -0,0 +1,428 @@
...
+      },
+      searchLanguage () {
+        return this.selectedLanguage
+      },
+      selectedVocabsString () {
+        return this.getSelectedVocabs.map(voc => voc.value).join(' ')
+      }
+    },
+    mounted () {
+      this.languages = window.SKOSMOS.languageOrder
```
UnniKohonen, line 49: Many of the assignments in mounted are not needed, the values can be set in data directly. At least `searchCounter`, `renderedResultsList` and `showNotation` can be removed. It's probably OK to remove `languages`, `languageStrings` and `vocabStrings` too.

resource/js/global-search.js
```diff
@@ -0,0 +1,428 @@
...
+            lang: window.SKOSMOS.lang
+          })
+        }
+        this.showAutoComplete()
+      },
+      hideAutoComplete () {
+        this.showDropdown = false
+        this.$forceUpdate()
+      },
+      gotoSearchPage () {
```
UnniKohonen, line 228: Is the query parameter supposed to have the wildcard character when opening the search page? In Skosmos 2, it is not included on the search result page

UnniKohonen (changes requested):
The new vocab list looks great! I forgot to mention this in the last review but the vocab list is missing the option for all vocabularies, is it supposed to be included in Skosmos 3.0/this PR?

Some small nitpicks about the layout:
- In the layout guide, the unchecked checkboxes have a light background and a dark border (but the guide is for Finto and not Skomos so it's probably OK)
- The magnifying glass button is not a square

resource/js/global-search.js
joelit: I am leaving it to the next increment, as it's not strictly necessary here yet

resource/js/global-search.js
joelit: It is needed for the API, but you are correct that Skosmos 2 does not show the wildcard in the URL. I do think it should be shown, though, as that's what's sent to the backend.

joelit:
SonarQube is complaining about the repetitive nature of the two dropdown menus in the search bar, wishing for a generic parameterized implementation that renders the different dropdown menus. In this case where we should ever have just the two menus for vocab and language, I am not sure getting rid of the repetitive code would improve the readability or the maintainability of the code.

resource/js/global-search.js
osma: I don't think we have a requirement that the search result page should have the same URL parameters as the call to the REST API method. So I agree with @UnniKohonen - I think the wildcard should not be part of the URL for the search result page.

resource/js/global-search.js
```diff
@@ -0,0 +1,431 @@
...
+        fetch(skosmosSearchUrl)
+          .then(data => data.json())
+          .then(data => {
+            if (mySearchCounter === this.searchCounter) {
+              this.renderedResultsList = data.results // update results (update cache if it is implemented)
+              this.renderResults() // render after the fetch has finished
+            }
+          })
+      },
+      formatSearchUrlParams () {
```
osma, line 99: This method is used for two different purposes:
- generating parameters for the search result page URL
- generating parameters for the REST API `search` method

This leads to unfortunate consequences: some search result page URL parameters have been renamed (`q` -> `query`, `vocab` -> `vocabs`, and it is also given unnecessary parameters `lang` and `unique`).

I suggest splitting this into two methods: one for the search result page and another for the REST API.

src/controller/WebController.php
```diff
@@ -339,7 +339,7 @@ public function invokeGlobalSearch($request)
 
         $parameters = new ConceptSearchParameters($request, $this->model->getConfig());
 
-        $vocabs = $request->getQueryParam('vocabs'); # optional
+        $vocabs = $request->getQueryParam('vocab'); # optional
```
osma: I suggest reverting this (and related changes to Cypress tests). Let's keep the old URL parameters for the search result page.

src/controller/WebController.php
```diff
@@ -387,7 +387,7 @@ public function invokeGlobalSearch($request)
                 'rest' => $parameters->getOffset() > 0,
                 'global_search' => true,
                 'search_failed' => $errored,
-                'term' => $request->getQueryParamRaw('q'),
+                'term' => $request->getQueryParamRaw('query'),
```
osma: Ditto, same as above, revert this.

src/view/base-template.twig
```diff
@@ -35,47 +35,56 @@
...
+        {% endif %}
+        {% endfor %}
+        {% endif %}
+      </ul>
+    </div>
+    {% if request.vocabid == '' and request.page != 'about' and request.page != 'feedback' %}
+    <div class="collapse container-fluid mt-2 pe-0 py-3" id="global-search-bar">
+      <div class="row d-flex align-items-top">
+        <div class="col" id="global-search-text">
+          <h2>Yhteishaku sanastoista</h2>
```
osma: Ditto, would be good to add a `trans` filter here so we don't forget:

```suggestion
          <h2>{{ "Global search"|trans }}</h2>
```

src/view/base-template.twig
```diff
@@ -35,47 +35,56 @@
...
-            </li>
-            {% endif %}
-            {% endfor %}
-            {% endif %}
-          </ul>
+    <div class="container-fluid bg-dark d-flex my-auto py-3 px-4 text-bg-dark">
+      {% if request.vocabid == '' and request.page != 'about' and request.page != 'feedback' %}
+        <div>
+          <button class="fs-6 py-3 text-light collapsed" id="global-search-toggle" data-bs-toggle="collapse" data-bs-target="#global-search-bar" aria-expanded="false" aria-controls="global-search-bar">
+              Yhteishaku sanastoista
```
osma: Would be good to add a `trans` filter here so we don't forget:

```suggestion
              {{ "Global search"|trans }}
```

joelit:
The requested changes were applied. I'll make sure the related epic contains the updated specifications, or the request for discussion thereof. I ended up splitting the language updating mechanism from the language parsing mechanism for ease of maintenance.

resource/css/skosmos.css
```diff
@@ -162,7 +166,7 @@ body {
   /* Headerbar
    *****************************************/
   #headerbar.landing, #headerbar.landing-one-vocab {
-    color: var(--dark-color);
+    color: var(--headerbar-text);
```
osma: Does not match variable defined above.

```suggestion
    color: var(--header-bar-text);
```

resource/css/skosmos.css
```diff
@@ -192,49 +196,180 @@ body {
     font-family: var(--font-family-heading);
   }
 
-  #search-multi-vocab-title {
-    color: var(--vocab-text);
+  #global-search-toggle {
+    transition: none;
+    background-color: var(--headerbar-vocab-bg);
```
osma, line 202: No such variable has been defined (but I see that it had been used before in skosmos.css line 173)
Just set the variable at the top?

resource/js/global-search.js
```diff
@@ -0,0 +1,436 @@
...
+        renderedResultsList: [],
+        languageStrings: null,
+        vocabStrings: null,
+        showDropdown: false,
+        showNotation: null
+      }
+    },
+    computed: {
+      vocabSelectorPlaceholder () {
+        return $t('1. Valitse sanasto')
```
osma:
```suggestion
        return $t('1. Choose vocabulary')
```

resource/js/global-search.js
```diff
@@ -0,0 +1,436 @@
...
+      },
+      langSelectorPlaceholder () {
+        return $t('2. Valitse kieli')
```
osma:
```suggestion
        return $t('2. Choose language')
```

resource/js/global-search.js
```diff
@@ -0,0 +1,436 @@
...
+      },
+      searchPlaceholder () {
+        return $t('3. Kirjoita hakutermi')
```
osma:
```suggestion
        return $t('3. Enter search term')
```

resource/js/global-search.js
```diff
@@ -0,0 +1,436 @@
...
+      },
+      formatVocabParam () {
+        const vocabs = this.getSelectedVocabs
+        return vocabs.map(voc => voc.key).join(' ')
+      },
+      notationMatches (searchTerm, notation) {
+        return notation?.toLowerCase()?.includes(searchTerm.toLowerCase()) === true
+      },
+      getSearchLang () {
+        const urlParams = new URLSearchParams(window.location.search)
```
osma, line 126: Not sure I understand why this state information needs to be taken from the URL, when there are other places as well (e.g. component state, DOM, SKOSMOS object). But we can talk about this later

Issue #1480, opened 2023-08-22
Epic: Landing Page


joelit:
## Description of the enhancement

* Multiple vocabulary layout [static example](https://github.com/NatLibFi/Skosmos/blob/skosmos-3-html/frontpage.html)
* Single vocabulaty layout [static example](https://github.com/NatLibFi/Skosmos/blob/skosmos-3-html/frontpage-one-vocab.html)

- [x] Multiple vocabulary layout implemented
- [ ] Single vocabulary layout implemented
- [ ] Customizable service logo ("Skosmos" by default)
- [x] Main heading is "Vocabularies" / "Sanastot"
- [x] Vocabularies grouped by category; category headings follow visual spec (size, color, padding/margin)
- [x] Vocabularies shown; vocabulary names follow visual spec (size, color, underline offset, underline color, padding/margin)
- [x] Equal length columns (aligned at both top and bottom)
- [x] Footer implemented and follows visual spec
- [x] Keyboard navigation works
- [x] Responsive layout
- [ ] The background should have a line motif
- [x] The background colour should follow the spec
- [x] The space between the columns follows the spec

Issue #1565, opened 2023-11-24
Refactor: define vocab_url Twig filter for constructing links to vocabularies


osma:
## Description of the enhancement

Our Skosmos 3 Twig templates contain code like this for constructing the URL to the home page:
https://github.com/NatLibFi/Skosmos/blob/43684969b45a3d92cc6cfd9948338260f56fffec/src/view/base-template.twig#L34

(the same kind of snippet appears at least two more times further down in the same file)

This is a very cumbersome and fragile way of constructing the URL, which usully amounts to something like `/en` or `/fi` or `/fi?clang=en` . For other kinds of URLs, we already have defined a custom filter `link_url` for Twig. I think a similar filter, perhaps called `vocab_url` should be defined and used here, which would move the URL construction logic away from the Twig template and into the PHP backend code, which is better suited for this kind of conditional URL string manipulation.

## Who are the users that would benefit from the enhancement and how?

Developers trying to keep sane

## What new functionalities would the enhancement make possible?

Make the Twig template cleaner and easier to understand

## Why is the enhancement important?

Refactoring early is good.

PR #1780, opened 2025-03-18
Visual layout improvements especially for landing page

### LINKED ISSUE #1320: Avoid specifying font sizes in px
This accessibility problem was pointed out by @kinow in [this discussion](https://github.com/NatLibFi/Skosmos/pull/1319#discussion_r864732515). Users with poor eyesight often set their browser to use a larger than normal font size and web sites should respect that. (There are other mechanisms for this as well, for example page zoom and minimum font size, which should be working already; this issue is specific to the default font size setting.)

Any Skosmos page.

1. Set the browser font size to larger than default (e.g. Chromium: Settings -> Appearance -> Font Size -> Huge).
2. Open any Skosmos page, e.g. https://dev.finto.fi

Expected that the fonts should be much larger than normal. But they are the same as usual, ignoring the Font Size setting.

The problem is that font sizes have been declared with explicit pixel sizes such as `14px` and `16px` in CSS. Instead, relative values using `em` or `rem` could be used, which would allow the text size to scale according to the browser setting.

Compare with https://ai.finto.fi which uses a similar layout as Finto/Skosmos, but uses different font size declarations. Here I've chosen huge font size in Chromium. It's not super pretty, but at least the text is very large as it should be:

image

Chromium (but same on Firefox)

osma:

The landing page didn't follow the visual spec very well and there were issues with font sizes, with some users seeing absurdly large text. This PR attempts to fix the font size inconsistency issue by changing all font-size declarations to use `rem` units. There are many visual improvements to the landing page layout as well as some adjustments that also affect other page types.

- part of #1480 (mostly)
- closes #1320 

- always specify font sizes in `rem` (except for the `:root` element, where `em` is better)
- adjust landing page layout to better follow visual spec
- make landing page layout responsive & columns equal height
- adjust footer layout to better follow visual spec, at least on landing page
- set landing page default headline to "Vocabularies" (can be overridden in config.ttl)
- many small adjustments to font sizes, colors etc., also affecting other page types

- I hope I didn't break anything too badly when changing CSS rules

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -62,15 +62,15 @@ body {
 /* Font awesome
  *****************************************/
   #main-container.landing #info-box .fa-arrow-right {
-    color: var(--info-box-arrow);
+    color: var(--secondary-dark-color);
   }
 
   #main-container.landing-one-vocab #info-box .fa-arrow-right {
-    color: var(--info-box-one-arrow);
+    color: var(--secondary-dark-color);
   }
 
   #vocab-box .fa-arrow-right {
-    color: var(--vocab-box-one-arrow);
+    color: var(--secondary-dark-color);
   }
 
   #main-container.vocab-home .fa-arrow-right, #main-container.concept .fa-arrow-right {
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -39,13 +39,12 @@
   --font-size: 1em; /* same as the browser default font size, often 16px */
   --font-family: 'Public Sans', sans-serif;
   --font-family-heading: 'EB Garamond', serif;
-
-  font-family: var(--font-family);
-  font-size: var(--font-size);
 }
 
 body {
   color: var(--dark-color);
+  font-family: var(--font-family);
+  font-size: var(--font-size);
 }
 
 .bg-dark {
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -161,7 +161,7 @@ body {
   /* Headerbar
    *****************************************/
   #headerbar.landing, #headerbar.landing-one-vocab {
-    color: var(--headerbar-text-1);
+    color: var(--dark-color);
   }
 
   #headerbar.vocab-home, #headerbar.concept {
@@ -178,13 +178,13 @@ body {
   }
 
   #vocab-title a {
-    color: var(--headerbar-text-2);
+    color: var(--vocab-text);
     font-family: var(--font-family-heading);
     text-decoration: none;
   }
 
   #search-multi-vocab-title {
-    color: var(--headerbar-text-2);
+    color: var(--vocab-text);
     font-family: var(--font-family-heading);
   }
 
@@ -193,7 +193,7 @@ body {
   }
 
   #search-wrapper {
-    color: var(--headerbar-text-2);
+    color: var(--vocab-text);
     float: right;
     height: 3.5em;
 }
@@ -204,7 +204,7 @@ body {
 
   #search-wrapper button.dropdown-toggle {
     background-color: var(--search-bg);
-    color: var(--headerbar-text-2);
+    color: var(--vocab-text);
     border: 1px solid var(--search-border);
     border-radius: 0;
     padding-left: 1rem;
@@ -223,7 +223,7 @@ body {
   #search-wrapper #language-list {
     inset: -3px auto auto 0px !important;
     background-color: var(--search-bg);
-    color: var(--headerbar-text-2);
+    color: var(--vocab-text);
     border: 1px solid var(--search-border);
     border-radius: 0;
   }
@@ -239,7 +239,7 @@ body {
   }
 
   #search-wrapper .form-check-input:checked {
-    background-color: var(--headerbar-text-2);
+    background-color: var(--vocab-text);
   }
 
   #search-wrapper input {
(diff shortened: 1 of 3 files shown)


src/view/landing.twig
```diff
@@ -5,52 +5,43 @@
 {% block url %}{{ BaseHref }}{{ request.langurl }}{% endblock %}
 
 {% block content %}
-<div class="col-md-7">
+<div class="col-lg-7 mb-custom mb-lg-0">
```
UnniKohonen, line 8: `mb-custom` doesn't seem to have any effect here

src/view/landing.twig
```diff
@@ -5,52 +5,43 @@
...
+        <p class="fs-5">Skosmos is being developed at the National Library of Finland.</p>
       </div>
-    </div>
-    <div class="col-md-12 flex-grow-1">
-      <div class="bg-medium px-5 py-5" id="info-box">
-
-        <a class="fs-4 text-dark text-decoration-none" href="#">Skosmos is open source and on GitHub -></a>
 
+      <div class="bg-medium px-5 py-5 flex-grow-1 d-flex align-items-center" id="info-box">
+        <a class="fs-5 text-dark text-decoration-none" href="#">Skosmos is open source and on GitHub -></a>
```
UnniKohonen, line 43: A font awesome arrow icon could be added here.

src/view/landing.twig
```diff
@@ -5,52 +5,43 @@
...
       {% endif %}
       {% for vocabClassName,vocabArray in request.vocabList %}
-      <div class="vocab-category border-top pb-5">
-        <h3 class="fs-3 pt-2">{{ vocabClassName }}</h3>
+      <div class="vocab-category border-top pb-4">
+        <h3 class="fs-4 pt-2">{{ vocabClassName }}</h3>
         <ul class="list-group">
         {% for vocab in vocabArray %}
-          <li class="list-group-item border-0 ps-0"><a class="fs-4 fw-bold" href="{{ vocab.id }}/{{ request.lang }}/{% if request.contentLang != request.lang and request.contentLang != '' and request.contentLang in vocab.config.languages %}?clang={{ request.contentLang }}{% endif %}">{{ vocab.title }}</a></li>
+          <li class="list-group-item ps-0 py-1"><a class="fs-5 fw-bold" href="{{ vocab.id }}/{{ request.lang }}/{% if request.contentLang != request.lang and request.contentLang != '' and request.contentLang in vocab.config.languages %}?clang={{ request.contentLang }}{% endif %}">{{ vocab.title }}</a></li>
```
UnniKohonen, line 23: Maybe this should be split into multiple lines

src/view/landing.twig
```diff
@@ -5,52 +5,43 @@
...
 </div>
-<div class="col-md-5">
-  <div class="gy-5 row h-100 flex-column">
-    <div class="col-md-12">
-      <div class="bg-dark px-5 py-5 text-light" id="welcome-box">
-
-        <h2 class="fw-bold fs-2">Welcome to the Skosmos browser demo</h2>
-        <p class="fs-4">
+<div class="col-lg-5 d-flex flex-column">
+      <div class="bg-dark px-5 py-5 mb-custom text-light" id="welcome-box">
```
UnniKohonen, line 32: Since the `mb-custom` class is only used here, I'd maybe define the margin directly for `welcome-box` in CSS

resource/css/skosmos.css
```diff
@@ -36,27 +36,16 @@
...
-  --font-size: 14px;
+  --font-size: 1em; /* same as the browser default font size, often 16px */
   --font-family: 'Public Sans', sans-serif;
   --font-family-heading: 'EB Garamond', serif;
-}
 
-body {
-  color: var(--dark-color);
   font-family: var(--font-family);
   font-size: var(--font-size);
```
UnniKohonen, line 44: `font-size` and `font-family` should be in `body`

UnniKohonen (commented):
This looks much better than it did before but there are some things I noticed that should still be fixed:
- The main sans-serif font is not set correctly
- To me the Skosmos logo looks a little big on the landing in comparison to the new sizing
- Vocab titles on vocab home page are no longer centered vertically

I also commented some minor code chages

src/view/landing.twig
osma: It does if you use a narrow screen (less than `lg`) - then it will ensure some space between the contents of the first column and that of the second column.

src/view/landing.twig
osma: Yes, or even refactored completely to take out the logic from the Twig template...

src/view/landing.twig
osma: Not true (see above comment), it is used in two places for the responsive layout. But maybe it should be renamed to something that better describes its purpose...

src/view/landing.twig
osma: I would like to refactor this into a separate `vocab_url` Twig filter (similar to `link_url`), but I think it's better to do it separately from this PR which is already getting quite large in scope.

I edited the existing issue #1565 which explains how to do this.

osma:
Thanks for the comments @UnniKohonen! I think I've now implemented all the suggestions, except the horrible way of constructing links to vocabularies, which I think is a separate problem a bit out of scope for this PR, but we have issue #1565 to remind us about it.

PR #1785, opened 2025-04-17
WIP: optimize change list / New tab SPARQL query

osma:
This is work in progress. It's not yet functional.

I'm trying to optimize the "New" / "New and Deprecated" tab that shows the recently added (or deprecated) concepts. It is very slow, especially in the case of a large vocabulary such as KANTO/finaf.

The optimized SPARQL query in this PR introduces a cutoff date, by default 1 year in the past. Only concepts whose timestamp (created or modified, depending on circumstances) is after this cutoff date are included in the query results. This drastically reduces the number of concepts to consider and so speeds up the query by a factor of approximately 10 (yso) to 15 (finaf).

n/a

- optimize the SPARQL query that queries the new and deprecated concepts, introducing a cutoff date
- try to fix the unit tests that broke due to the above

Unfortunately, this change causes problems for the unit tests. The test data is much older than 1 year, so a lot of tests will fail because the time stamps are older than the default cutoff date. I tried to fix this by making the cutoff date into a parameter, but this makes the code more complicated, and I didn't yet get it working. I wonder if another mechanism - for example a configuration setting that sets the cutoff date - would work better for this.

This PR is against skosmos-2, because the functionality would be useful in current (Skosmos 2.x) based Finto. However, the same change should be made to Skosmos 3 (main) as well.

## Checklist

### DIFF
--- a/controller/WebController.php
+++ b/controller/WebController.php
@@ -595,14 +595,15 @@ class WebController extends Controller
      * @param string $prop the name of the property eg. 'dc:modified'.
      * @param int $offset starting index offset
      * @param int $limit maximum number of concepts to return
+     * @param int $cutoffYears amount of years to consider past current date
      * @return Array list of concepts
      */
-    public function getChangeList($request, $prop, $offset=0, $limit=200)
+    public function getChangeList($request, $prop, $offset=0, $limit=200, $cutoffYears=1)
     {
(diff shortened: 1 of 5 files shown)


osma:
Superseded by #1954 which is for Skosmos 3. Closing this one.

PR #1787, opened 2025-05-13
Narrower container for landing & other global page types

osma:

The landing page is too wide, it doesn't match the layout spec.

This PR adjusts the landing page and other global page types so that the container is a bit narrower on typical desktop browser window sizes. In practice, it reverts the landing page to the default Bootstrap 5 widths for `.container` while retaining the expanded 1460px width for the vocabulary-specific page types (vocab-home, concept, vocab-search).

- Part of #1480 

* set max-width: 1460px only for vocabulary-specific page types instead of all of them

Not sure if these widths are ideal, but at least it should be an improvement.

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -359,7 +359,7 @@ body {
 
   /* Main container
    *****************************************/
-  .container {
+  .vocab-home .container, .concept .container, .vocab-search .container {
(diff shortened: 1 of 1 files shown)


joelit (approved):
LGTM

Issue #1751, opened 2025-01-29
Plugin Support


miguelvaara:

joelit:
This Epic is done in the scope of Skosmos 3.0! Further plugin development should be managed in separate issues. :)

PR #1790, opened 2025-05-21
New plugin infrastructure for Skosmos 3.0

### LINKED ISSUE #1789: New plugin infrastructure for Skosmos 3.0

Plugin infrastructure for Skosmos 3.0 needs to be designed and implemented.

Adding plugins to Skosmos 3.0 using vanilla JS or Vue.

## Task list

- [ ] Update PluginRegister to follow new folder structure
- [ ] Load plugin JS files
- [ ] Load plugin CSS files
- [ ] Make plugin callbacks on page load
- [ ] Add new elements to DOM for plugins

UnniKohonen:

Skosmos 3 does not currently have support for plugins. This PR adds new infrastructure for plugins.

- Closes #1789
- Part of #1751

- Remove support for plugin templates
- Fix paths to plugin files in plugin registry
- Add plugins and plugin callbacks to scripts.inc file
- Add a function to call plugin callbacks on page load
- Add plugin slot elements to DOM
- Add test plugins
- Add cypress tests for plugins

## Checklist

### DIFF
--- a/resource/js/make-plugin-callbacks.js
+++ b/resource/js/make-plugin-callbacks.js
@@ -18,10 +18,12 @@ const makeCallbacks = () => {
   }
 }
 
+// Make callbacks on page load
 document.addEventListener('DOMContentLoaded', () => {
   makeCallbacks()
 })
 
+// Make callbacks on partial page load
 document.addEventListener('loadConceptPage', () => {
   makeCallbacks()
 })
--- a/resource/js/make-plugin-callbacks.js
+++ b/resource/js/make-plugin-callbacks.js
@@ -1,0 +1,27 @@
+const makeCallbacks = () => {
+  const newPage = window.SKOSMOS.page || 'page'
+  const newUri = window.SKOSMOS.uri
+  const newPrefs = window.SKOSMOS.prefLabels
+  const ldJsonScript = document.querySelector('script[type="application/ld+json"]')
+  const embeddedJsonLd  = ldJsonScript ? JSON.parse(ldJsonScript.innerHTML) : {}
+
+  const params = {'uri': newUri, 'prefLabels': newPrefs, 'page': newPage, 'json-ld': embeddedJsonLd}
+
+  if (window.pluginCallbacks) {
+    for (let i in window.pluginCallbacks) {
+      const fname = window.pluginCallbacks[i]
+      const callback = window[fname]
+      if (typeof callback === 'function') {
+        callback(params)
+      }
+    }
+  }
+}
+
+document.addEventListener('DOMContentLoaded', () => {
+  makeCallbacks()
+})
+
+document.addEventListener('loadConceptPage', () => {
+  makeCallbacks()
+})
--- a/resource/js/make-plugin-callbacks.js
+++ b/resource/js/make-plugin-callbacks.js
@@ -5,7 +5,7 @@ const makeCallbacks = () => {
   const ldJsonScript = document.querySelector('script[type="application/ld+json"]')
   const embeddedJsonLd = ldJsonScript ? JSON.parse(ldJsonScript.innerHTML) : {}
 
-  const params = { uri: newUri, prefLabels: newPrefs, page: newPage, jsonLd: embeddedJsonLd }
+  const params = { uri: newUri, prefLabels: newPrefs, pageType: newPage, jsonLd: embeddedJsonLd }
 
   if (window.pluginCallbacks) {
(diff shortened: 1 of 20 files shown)


tests/cypress/template/plugins.cy.js
osma, line 1: These tests are excellent, but I think they could be expanded to cover partial page load events. The tests should verify that the plugins display the right information (that may depend on the current page and current concept) even after partial page loads. This naturally requires some changes to the plugins themselves.

src/view/scripts.inc.twig
```diff
@@ -70,17 +71,18 @@ window.SKOSMOS = {
...
 <script src="node_modules/bootstrap/dist/js/bootstrap.bundle.js"></script>
 
 <!-- Vue.js -->
 <script src="node_modules/vue/dist/vue.global.js"></script>
 
+<!-- Plugin JS sources -->
+{% for files in request.plugins.pluginsJS %}
+  {% for file in files %}<script src="{{ file }}"></script>{% endfor %}
+{% endfor %}
+{% if request.plugins.callbacks %}<script>var pluginCallbacks = [{% for function in request.plugins.callbacks %}{% if not loop.first %}, {% endif %}"{{ function }}"{% endfor %}];</script>{% endif %}
```
osma, line 84: This declares a new global JS variable `pluginCallbacks`. Would it make sense to move it within the SKOSMOS object instead, to avoid polluting the global namespace?

resource/js/make-plugin-callbacks.js
```diff
@@ -0,0 +1,29 @@
+const makeCallbacks = () => {
+  const newPage = window.SKOSMOS.pageType
+  const newUri = window.SKOSMOS.uri
+  const newPrefs = window.SKOSMOS.prefLabels
+  const ldJsonScript = document.querySelector('script[type="application/ld+json"]')
+  const embeddedJsonLd = ldJsonScript ? JSON.parse(ldJsonScript.innerHTML) : {}
+
+  const params = { uri: newUri, prefLabels: newPrefs, page: newPage, jsonLd: embeddedJsonLd }
```
osma, line 8: The `page` parameter name is a bit misleading. Now that we are changing the plugin API anyway, could we rename it to `pageType`?

resource/js/make-plugin-callbacks.js
```diff
@@ -0,0 +1,29 @@
+const makeCallbacks = () => {
+  const newPage = window.SKOSMOS.pageType
+  const newUri = window.SKOSMOS.uri
+  const newPrefs = window.SKOSMOS.prefLabels
+  const ldJsonScript = document.querySelector('script[type="application/ld+json"]')
```
osma, line 5: This works for now, but I think it would be better to give the script element containing the JSON-LD an `id` attribute and use `getElementById` here.

osma:
I see that SonarCloud has two complaints about use of `var`:

> var VANILLA_JS_PLUGIN = VANILLA_JS_PLUGIN || {}
> Unexpected var, use let or const instead.

(and the same for the Vue plugin)

Can these be fixed as well? Sorry I didn't notice this in my earlier review.

resource/js/partial-page-load.js
```diff
@@ -45,8 +45,8 @@ const updateTitle = (conceptHTML) => {
 }
 
 const updateJsonLD = (conceptHTML) => {
-  const JsonLD = document.querySelector('script[type="application/ld+json"]')
-  const newJsonLD = conceptHTML.querySelector('script[type="application/ld+json"]')
+  const JsonLD = document.querySelector('#json-ld-data')
```
osma, line 48:
```suggestion
  const JsonLD = document.getElementById('json-ld-data')
```

resource/js/partial-page-load.js
```diff
@@ -45,8 +45,8 @@ const updateTitle = (conceptHTML) => {
...
+  const newJsonLD = conceptHTML.querySelector('#json-ld-data')
```
osma, line 49:
```suggestion
  const newJsonLD = conceptHTML.getElementById('json-ld-data')
```

osma (approved):
Maybe fix the SonarCloud warnings and consider my suggestions for use of getElementById. Then this is good for merging :tada:

resource/js/partial-page-load.js
UnniKohonen: `getElementById` is a method of the `document` object so it can't be used with the `conceptHTML` object. This is why I used the `querySelector` method on both lines.

PR #1791, opened 2025-05-21
Set max width for topbar and headerbar

osma:

The layout of the top header looks funny on very wide screens, because the contents (logo, navigation links etc.) are placed near the edges of the page.

This PR changes the header elements to use Bootstrap Grid classes (row, col) and then sets a max-width of 1460px for the contents. Also some margins and paddings are adjusted. The end result is that the navigation elements at the top are kept in the middle of the page, not near the edges.

- part of #1480

* use Bootstrap Grid classes for header
* add max-width 1460px for the header elements
* adjust margins and paddings

I'm still not 100% happy with the alignment of the left and right edges. Especially on some zoom levels / page widths the edges don't always align nicely.

One problem here is that the Skosmos logo used within vocabularies (negative version, for use on a dark background) includes a lot of empty space around the logo itself within the svg file, making it hard to align. I tried two methods of cropping the svg to its content, but when I replaced the svg file with a cropped version, the logo just failed to display at all. I can't figure out why, but in any case, here are the two cropped logos (they look a bit funny here if you have a white background) in case they are useful:
 
skosmos-NEGA-RGB-cropped
skosmos-NEGA-RGB-cropped-2

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -145,6 +145,7 @@ body {
     background-repeat: no-repeat;
     width: 168px;
     height: 60px;
+    display: inline-block;
   }
 
   #skosmos-logo-top h2 {
@@ -166,6 +167,11 @@ body {
 
   #headerbar.vocab-home, #headerbar.concept {
     background-color: var(--headerbar-vocab-bg);
+  }
+
+  header .container-fluid .row {
+    max-width: 1460px;
+    margin: 0 auto;
   }
 
   #skosmos-logo {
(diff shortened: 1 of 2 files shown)


joelit (changes requested):
Looks good, and it is definitely an improvement! I would align the logo to the main container using the bootstrap width breakpoints at 1400, 1200, 992, etc. so that the logo would get similar margins to the main container. Now the left-side margin of the logo is aligned to the topbar text.

However, this is down to individual taste. We could just iterate and ask for user feedback and resume tweaking the CSS on a later occasion - but then the issue needs to be documented.

The svg-logo started working by giving read permissions to it.

joelit (approved):

PR #1800, opened 2025-08-20
Avoid useless SPARQL queries

### LINKED ISSUE #1799: Useless SPARQL queries for concept labels

http://localhost/Skosmos/finaf/fi/search?clang=fi&q=aaron

http://localhost/Skosmos/finaf/fi/page?uri=http%3A%2F%2Furn.fi%2FURN%3ANBN%3Afi%3Aau%3Afinaf%3A000224268

## Description of the problem

Skosmos performs a lot of SPARQL queries, many of which seem to be useless or redundant. This happens in particular on the search results page, but also to some extent on the concept page. There are many SPARQL queries that query for the label of a particular URI (via the GenericSparql->queryLabel() method). In some cases the same query is performed repeatedly while processing the same request.

FWIW, I first tried to address this in PRs #1796 (and #1798 for Skosmos 2) which adds a SPARQL query result cache, but after some more research, I think those are just papering over the real problem - most of the queries should be avoided in the first place, and a cache is probably not useful after fixing that.

Here are known cases of useless queries:

1. ConceptPropertyValue.__construct() calls $this->getLabel() to determine if a concept is external to the current vocabulary. But getLabel() often falls back to calling getExternalLabel(), resulting in a SPARQL query. That should not be necessary for finding out whether the concept is external, so it could be avoided.

2. Concept.getProperties() ends up performing many getLabel() / queryLabel() calls, especially when displaying concepts like "Aaron (Raamatun henkilö)" (see URLs above) which contains a lot of links to other URI namespaces (mts, viaf, yso, lexvo...). These are probably necessary when displaying the **concept page**, but not when displaying the concept on the **search results page**, which won't display most of the linked information to the user.

There could be other useless queries as well, this is just scratching the surface.

Firefox or curl, doesn't matter

osma:

Skosmos performs a number of useless SPARQL queries, see #1799.

This PR reduces the number of unnecessary SPARQL queries, thereby increasing performance.

**Benchmark results**

(Note: These have been updated following the merge of #1796, which eliminated duplicate SPARQL queries using a cache - before the query cache was added, the differences were even more dramatic.)

On my local Skosmos installation, a global search query for "aaron"

* before (main): took 3.3 seconds and resulted in 22 SPARQL queries
* after (this PR): took 0.7 seconds and resulted in 2 SPARQL queries

Showing the concept page for "Aaron (Raamatun henkilö)"

* before (main): took 2.7 seconds and resulted in 22 SPARQL queries
* after (this PR): took 2.1 seconds and resulted in 16 SPARQL queries

- Closes #1799

* make it possible to limit the properties that Concept.getProperties() returns, avoiding the processing of properties that aren't shown on the search result page (and the SPARQL queries that would be triggered by that processing)
* introduce an `$allowExternal` parameter for many getLabel() methods that, when set to false, avoids performing a SPARQL query in situations where it isn't really necessary (in a few cases, this replaces a similar preexisting `$queryExVocabs` parameter that was used inconsistently)
* refactor Concept.getProperties() and Concept.getMappingProperties() a little bit, extracting helper methods that perform part of the work

Unit test coverage is not 100% for the modified code. In particular, I haven't written unit tests for the parts of Concept.php that were refactored. However, this is not a regression - we didn't have such tests earlier.

It's possible that some special cases related to how information is displayed on search results or concept pages have been broken by the changes in the PR, but it's pretty difficult to tell. I've tried to test for tricky cases but I may well have missed something.

## Checklist

### DIFF
--- a/tests/ConceptMappingPropertyValueTest.php
+++ b/tests/ConceptMappingPropertyValueTest.php
@@ -113,6 +113,7 @@ class ConceptMappingPropertyValueTest extends PHPUnit\Framework\TestCase
         $mapping = new ConceptMappingPropertyValue($this->model, $this->vocab, $mockres, $mocksource, 'skos:exactMatch');
         $this->assertEquals('english lit', $mapping->getLabel('en'));
         $this->assertEquals('default lit', $mapping->getLabel());
+        $this->assertEquals('default lit', $mapping->getLabel()); // from labelcache
     }
 
     /**


joelit (approved):
LGTM! In certain cases this really cuts down on the number of generated queries.

PR #1802, opened 2025-08-21
Heading structure fixes

### LINKED ISSUE #1801: Heading level structure issues in Skosmos 3

http://localhost/Skosmos/

## Description of the problem

I reviewed the heading level structure of all currently implemented (main branch) Skosmos 3 page types. I mostly used the WAVE plugin in Firefox, which reports issues related to heading levels (but it won't complain if there are more than one h1 elements, because it's not a strict rule to avoid it).

The general requirements that make sense for Skosmos are:

1. There should be one, and only one, H1 element on a page
2. Heading levels should not be skipped (e.g. `h1, h2, h3` is OK, but `h1, h3` or `h1, h2, h4` is not)

Here is an overview:

* landing: no issues (uses h1, h2, h3)
* vocab-home: **two** h1 elements instead of one
* error: **no headings at all**
* global search results: **no headings at all**
* vocab search results: **no h1**, but there are h2 and h4 headings, **h3 is skipped**
* concept: no issues (uses h1, h2, h3)
* about: no issues (uses h1, h2)
* feedback: no issues (uses h1, h2)

Things to fix:

- [x] vocab-home: make sure there is only one h1 element
- [x] error: add h1 element
- [x] global search results: add h1 element
- [x] vocab search results: add h1 element
- [x] vocab search results: make sure heading levels are not skipped

Firefox

osma:

Fix many heading level issues in Skosmos 3, see #1801.

- Closes #1801

* add vocab header (with the vocab name and search fields) to vocab search result page
* make sure vocab-search and global-search pages have a h1 header: the "XX results for 'YYY'" message is now a h1 element
* adjust heading levels on vocab-search page to avoid skipping levels: the "Search options" element is now h3 instead of h4
* always use h2 element for the vocab header; previously it was h1 for the vocab-home page and h2 on other pages, but that causes problems with partial page loads, so keeping it always h2 is simpler (the "Vocabulary information" heading is now the only h1 element on vocab-home pages)
* add h1 heading "Error" on the error page
* unrelated: the error page was crashing on URLs like yso/en/page/xxxxx where xxxx is not a valid localname; this PR fixes the crash, although the resulting error page looks a bit funny (see below)
* change h4 headings to h2 on about page in Swedish and English to avoid skipping levels (in Finnish it was already h2)

I'm not 100% sure that it's a great idea to make the "XX results for 'YYY'" message on search results pages into a h1 level heading. But there is no other obvious heading on the page, unless we want to add a prominent heading like "Search results" to the page.

I fixed a PHP crash on the error page within a vocabulary (e.g. http://localhost/Skosmos/yso/en/page/xxx ), but now the sidebar looks a bit strange, probably due to some CSS issues. Also, there should probably be a vocab header on this page:

kuva

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -905,6 +905,12 @@ body {
   #search-results {
     background-color: var(--vocab-bg);
   }
+
+  #search-results h1 {
+    font-size: 1.3rem;
+    font-family: var(--font-family);
+  }
+
   .search-result {
     border-top: 1px solid var(--vocab-table-border);
     margin-top: 2.5rem;
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -907,7 +907,7 @@ body {
   }
 
   #search-results h1 {
-    font-size: 1.3rem;
+    font-size: 1.1rem;
     font-family: var(--font-family);
   }
 
--- a/src/model/Vocabulary.php
+++ b/src/model/Vocabulary.php
@@ -449,7 +449,7 @@ class Vocabulary extends DataObject implements Modifiable
      * @param string $clang content language
      * @return Concept
      */
-    public function getConceptInfo(string $uri, string $clang): Concept
+    public function getConceptInfo(string $uri, string $clang): ?Concept
     {
         $sparql = $this->getSparql();
         $conceptInfo = null;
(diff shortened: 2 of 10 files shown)


miguelvaara:
The review is scheduled for the day before the last day of the ongoing psrint, right after the daily

miguelvaara (commented):
Even though I cannot completely forget my earlier impressionss of Skosmos, I try to picture how the site and its pages would come across if someone described them to me just by using the heading levels (what kind of site it is and what the pages include).

A specialist at Celia once told me that a visually impaired person might not necessarily use a site in the "correct" way I described, but that over time assistive tools have shaped very personal habits and ways of navigating.

To avoid this mismatch I stick to my “eyes closed” experience while trusting that the chosen way of presenting the heading levels is valid and relevant. I focus on checking if the haedings work as they should.

**Issue 1:**
On http://localhost/skosmos/sv/about
there is a straight jump from level 2 to level 4.

PR #1803, opened 2025-08-22
Provide pre-built Docker images on Quay.io

### LINKED ISSUE #1763: Pre-built Docker images for public use
We're currently looking to deploy Skosmos on a Kubernetes cluster, and I was wondering if you could provide pre-built Docker images for public use.

While the source code is available and can be built locally, having a pre-built Docker image would simplify the deployment process, especially in automated environments or for those looking to quickly integrate Skosmos without needing to build the images themselves.

miguelvaara:
Sharing Docker images simplifies deployment processes, especially in automation as described in the issue: https://github.com/NatLibFi/Skosmos/issues/1763.

- Closes #1763

When the feature branch for this PR is merged into `main`, every push, pull request and tagged release on `main` will publish latest Docker image to Quay.io. Instructions for using the Docker images will be added to the documentation ([temporarly here](https://github.com/NatLibFi/Skosmos/pull/1803#issuecomment-3223686512)).  

It would have been possible to create a separate issue for improving GitHub Actions logging, but since logging is needed throughout working on this PR, the improvements are included here together with the related changes.

- The "new" YAML code style may raise some “feelings” ;-), but in my view using two spaces for indentation before list markers is clearer and more elegant. This format is also accepted by [YAML Lint](https://www.yamllint.com/).

- Uploading Docker images to Quay.io is implemented in GitHub Actions as two workflows: pushes and pull requests trigger one workflow while pushing a tag triggers another (one may appear as “skipped” in the Actions UI but in practice they are separate workflows). All dependency causing tests must pass successfully before the Quay.io-related jobs are executed.

- Progress was slowed down by timeouts in the `build-test` jobs. The limit has been increased to 60 minutes. Personally, I would like this change to be merged into `main`, but of course the matter is open for discussion.

- The code coverage tests, I did not consider them meaningful in the context of this issue.

## Instruction for the reviewer:

Make a change to the `testfile.txt` file (located in the document root) and commit it. Also add a tag (`git tag -a vN.N.N" -m "Test release version N.N.N"`) and push it (`git push origin issue1763-pre-built-docker-images-for-public-use --tags`). Monitor the GitHub Actions runs and finally verify the result in `quay.io/natlibfi/skosmos`. Try pulling the new image following the [instructions](https://github.com/NatLibFi/Skosmos/pull/1803#issuecomment-3223686512).

## Checklist

- [x] Created a repository on Quay.io
- [x] Defined permissions
- [x] Configured GitHub Actions for both Docker images
- [x] Tested and verified functionality
- [x] Documentation complete
- [x] Reviewed / approved
- [x] Merge

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,7 +110,7 @@ jobs:
     needs: [ build-test, cypress-test ]
     runs-on: ubuntu-24.04
     timeout-minutes: 15
-    if: github.event_name == 'push' && github.ref_name == 'issue1763-pre-built-docker-images-for-public-use'
+    if: github.event_name == 'push' && github.ref_name == 'main'
 
     steps:
       - name: Check out repo
@@ -129,8 +129,7 @@ jobs:
         with:
           images: quay.io/natlibfi/skosmos
           tags: |
-            type=raw,value=ci-latesttest
-            type=raw,value=ci-latesttest-${{ github.sha }}
+            type=raw,value=latest
 
       - name: Build and push to Quay.io
         uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83  # v6.18.0
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
   "description": "Thesaurus and controlled vocabulary browser using SKOS and SPARQL",
   "type": "project",
   "license": "MIT",
-  "version": "0.1.80",
+  "version": "3.0-beta.1-dev",
   "require": {
     "sweetrdf/easyrdf": "1.13.*",
     "symfony/polyfill-php80": "1.*",
--- a/tests/ModelTest.php
+++ b/tests/ModelTest.php
@@ -31,22 +31,22 @@ class ModelTest extends PHPUnit\Framework\TestCase
         $this->assertNotNull($model->getConfig());
     }
 
-   /**
-    * @covers Model::getVersion
-    */
-   public function testGetVersion()
-   {
-       $version = $this->model->getVersion();
-       // make sure that the returned version (which comes from composer.json)
-       // matches the version from git tags
-       $git_tag = rtrim(shell_exec('git describe --tags --always'));
-       $this->assertStringStartsWith(
-           "v" . $version,
-           $git_tag,
-           "Composer version '$version' doesn't match git tag '$git_tag'.\n" .
-     "Please run 'composer update' to update the Composer version."
-       );
-   }
+    /**
+     * @covers Model::getVersion
+     */
+    public function testGetVersion()
+    {
+        $version = $this->model->getVersion();
+        // make sure that the returned version (which comes from composer.json)
+        // matches the version from git tags
+        $git_tag = rtrim(shell_exec('git describe --tags --always'));
+        $this->assertStringStartsWith(
+            "v" . $version,
+            $git_tag,
+            "Composer version '$version' doesn't match git tag '$git_tag'.\n" .
+      "Please run 'composer update' to update the Composer version."
+        );
+    }
 
     /**
      * @covers Model::getVocabularyList


miguelvaara:
The instruction text will be added during the merge to the page [Install Skosmos with Fuseki in Docker](https://github.com/NatLibFi/Skosmos/wiki/Install-Skosmos-with-Fuseki-in-Docker).

## Using Pre-built Docker Images

You can view the tags associated with Skosmos releases in its [Quay.io repository](https://quay.io/repository/natlibfi/skosmos). For more details about releases and their related tags, see our [Releases page](https://github.com/NatLibFi/Skosmos/releases).

### Tags

Our releases follow semantic versioning and we apply it on Quay.io as follows:

- The `latest` tag always points to the most recent release.  
- When a patch version changes, a corresponding minor version tag is also created.  
- For each `nn.nn.nn` release tag, both a date-tagged and a hash-tagged version exist.  

### Pulling from Quay.io

You can either pull the Docker image separately with ```docker pull quay.io/natlibfi/skosmos:[tag]``` and then run a container from it as you prefer or you can both pull and run it in a single command
```docker run -d --name skosmos-test -p 8080:80 quay.io/natlibfi/skosmos:[tag]```.

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -2,6 +2,8 @@ name: CI
 
 on:
   push:
+    tags:
+      - '**'
```
osma, line 6: In my understanding, this means that the whole workflow is only triggered when a tag (any tag) is pushed. This would disable CI tests for most pushes. I don't think that's what we want?

.github/workflows/ci.yml
```diff
@@ -13,89 +15,194 @@ jobs:
...
+
+      - name: Check PHP code style
+        run: ./vendor/bin/php-cs-fixer fix src --diff --dry-run
+
+      - name: Check JavaScript code style
+        run: cd resource/js; npx standard *.js
+
+      - name: Run PHPUnit tests
+        run: |
+          while sleep 60; do echo "The tests are still running .."; done &
```
osma: Is this useful? Seems a bit hacky

.github/workflows/ci.yml
```diff
@@ -13,89 +15,194 @@ jobs:
...
+      - name: Run Cypress tests
+        run: npx cypress run
+
+  publish-docker-latest:
+    name: publish latest Docker image
+    needs: [ build-test, cypress-test ]
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
+    if: >-
+      (github.event_name == 'push' && github.ref_name == 'issue1763-pre-built-docker-images-for-public-use') ||
```
osma: This condition would be false after merging to main, so it should be dropped

.github/workflows/ci.yml
```diff
@@ -13,89 +15,194 @@ jobs:
...
+      (github.event_name == 'pull_request' && github.head_ref == 'issue1763-pre-built-docker-images-for-public-use')
```
osma: Ditto, would be false on main

.github/workflows/ci.yml
```diff
@@ -13,89 +15,194 @@ jobs:
...
+      - name: Build and push to Quay.io
+        uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83  # v6.18.0
+        with:
+          context: .
+          file: dockerfiles/Dockerfile.ubuntu
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+
+      - name: Debug event context
```
osma: Do we need this or is this a leftover from debugging?

.github/workflows/ci.yml
```diff
@@ -13,89 +15,194 @@ jobs:
...
+        run: |
+          echo "event_name=${{ github.event_name }}"
+          echo "ref=${{ github.ref }}"
+          echo "ref_name=${{ github.ref_name }}"
+
+      - name: Show computed tags
```
osma: Do we need this or is this a leftover from debugging?

.github/workflows/ci.yml
```diff
@@ -13,89 +15,194 @@ jobs:
...
+      - name: Build and push to Quay.io
+        uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83  # v6.18.0
+        with:
+          context: .
+          file: dockerfiles/Dockerfile.ubuntu
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+
+      - name: Debug event context
```
osma: Do we need this or is this a leftover from debugging?

.github/workflows/ci.yml
```diff
@@ -13,89 +15,194 @@ jobs:
...
+        run: |
+          echo "event_name=${{ github.event_name }}"
+          echo "ref=${{ github.ref }}"
+          echo "ref_name=${{ github.ref_name }}"
+
+      - name: Show computed tags
```
osma: Do we need this or is this a leftover from debugging?

composer.json
```diff
@@ -3,7 +3,7 @@
   "description": "Thesaurus and controlled vocabulary browser using SKOS and SPARQL",
   "type": "project",
   "license": "MIT",
-  "version": "3.0-beta.1-dev",
+  "version": "0.1.80",
```
osma: I don't think we want to change the version here

testfile.txt
osma, line 1: Drop this from the PR

tests/ModelTest.php
osma, line 1: Whitespace changes should be reverted unless there is a good reason for them

.github/workflows/ci.yml
```diff
@@ -2,6 +2,8 @@ name: CI
 
 on:
   push:
+    tags:
```
osma, line 5:
```suggestion
    branches:
      - '**'
    tags:
```

osma (changes requested):

osma:
Good now, let's hope it works in main! I will merge this now

.github/workflows/ci.yml
miguelvaara: The log generation code was left out

.github/workflows/ci.yml
miguelvaara: THe comment was relevant and Osma's suggestion and it was implemented:
  push:
    tags:
      - '**'
    branches:
      - '**'

PR #1805, opened 2025-08-27
Refactor: cleaner URL construction using global_url filter

### LINKED ISSUE #1565: Refactor: define vocab_url Twig filter for constructing links to vocabularies

osma:

Generating URLs in Twig templates is ugly. This PR defines a new `global_url` Twig extension function/filter and uses it to construct URLs so that the logic is implemented in PHP code.

- Closes #1565

- define `global_url` filter
- use the filter in base-template.twig instead of complicated Twig expressions

None

## Checklist

### DIFF
--- a/src/view/landing.twig
+++ b/src/view/landing.twig
@@ -20,7 +20,7 @@
         <h3 class="fs-4 pt-2">{{ vocabClassName }}</h3>
         <ul class="list-group">
         {% for vocab in vocabArray %}
-          <li class="list-group-item ps-0 py-1"><a class="fs-5 fw-bold" href="{{ vocab.id }}/{{ request.lang }}/{% if request.contentLang != request.lang and request.contentLang != '' and request.contentLang in vocab.config.languages %}?clang={{ request.contentLang }}{% endif %}">{{ vocab.title }}</a></li>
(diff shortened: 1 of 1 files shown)


UnniKohonen (approved):
All the links seem to be generated correctly.

I guess the filter could be called `global_url` since it is used to create non vocabulary specific links. (And `link_url` could be called `vocab_url` but maybe that's confusing).

Also, is it possible to use the filter in `landing.twig` on line 23 to generate the vocab list links?

Issue #1747, opened 2025-01-28
Epic: Sidebar


UnniKohonen:
## Description of the enhancement

Implementing sidebar structure and configuration for Skosmos 3. New configuration is described in #1465 and #1493

## What new functionalities would the enhancement make possible?

- [x] 1 Layout for vocab/concept page
- [ ] 2 Layout for search results page
- [x] 3 New configuration
- [x] 4 Cypress tests

PR #1806, opened 2025-08-28
Sidebar tab styling on narrow screens

UnniKohonen:

Sidebar tabs end up on two separate lines on narrow screen sizes. This PR adds an initial fix for this.

- Part of #1747

- Make sidebar tab padding more narrow
- Add stylings for sidebars with 4 tabs:
  - Reduce padding more on narrow screens
  - Make last tab overflow with a fade out on narrow screens
  - Use A-Z on alphabetical tab

Addresses requirement 7.2 in #1747

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -520,7 +520,6 @@ body {
     align-items: center;
     justify-content: center;
     transition: none;
-    padding: 0.5rem;
   }
 
   #main-container.vocab-home #sidebar .nav .active, #main-container.concept #sidebar .nav .active {
@@ -614,7 +613,7 @@ body {
 
   /* In smaller screen sizes, the last sidebar tab overflows to the right */
   @media only screen and (min-width: 768px) and (max-width: 1070px) {
-    #sidebar-tabs.wide .nav {
+    #sidebar-tabs.wide {
       flex-wrap: nowrap;
     }
 
@@ -654,13 +653,6 @@ body {
   @media only screen and (max-width: 900px) {
     #sidebar .nav-link {
       padding: 0.5rem 0.1rem !important;
-    }
-  }
-
-
-  @media only screen and (max-width: 820px) {
-    #sidebar .nav-link {
-      padding: 0.5rem 1px !important;
     }
   }
 
--- a/src/view/sidebar.inc.twig
+++ b/src/view/sidebar.inc.twig
@@ -3,9 +3,9 @@
   {% block sidebar %}
     <div class="sidebar-buttons">
       <h2 class="visually-hidden">{{'Sidebar listing: list and traverse vocabulary contents by a criterion' | trans}}</h2>
-      <ul class="nav nav-tabs-no-style nav-justified{% if vocab.config.sidebarViews | length == 4 and vocab.config.showDeprecatedChanges %} wide{% endif %}" id="sidebar-tabs" role="tablist">
+      <ul class="nav nav-tabs-no-style nav-justified{% if vocab.config.sidebarViews | length >= 4 and vocab.config.showDeprecatedChanges %} wide{% endif %}" id="sidebar-tabs" role="tablist">
         {% for view in vocab.config.sidebarViews %}
-          {% if view == 'alphabetical' %} {% set view_trans_text = (vocab.config.sidebarViews | length == 4) ? 'A-Z' : 'Alpha-nav' %}
+          {% if view == 'alphabetical' %} {% set view_trans_text = (vocab.config.sidebarViews | length >= 4) ? 'A-Z' : 'Alpha-nav' %}
           {% elseif view == 'hierarchy' %} {% set view_trans_text = 'Hier-nav' %}
           {% elseif view == 'groups' %} {% set view_trans_text = 'Group-nav' %}
           {% else %} {% set view_trans_text = vocab.config.showDeprecatedChanges ? 'Changes-and-deprecations-nav' : 'Changes-nav' %}{% endif %}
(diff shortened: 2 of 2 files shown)


resource/css/skosmos.css
```diff
@@ -607,6 +612,58 @@ body {
     color: var(--vocab-text) !important;
   }
 
+  /* In smaller screen sizes, the last sidebar tab overflows to the right */
+  @media only screen and (min-width: 768px) and (max-width: 1070px) {
+    #sidebar-tabs.wide .nav {
```
osma: Does this actually match any HTML elements? It seems to me that the `.nav` class is set for the same element (`#sidebar-tabs`), so this doesn't match its children.

In any case, I think this could be replaced with `class="flex-nowrap"` which is a Bootstrap utility class.

resource/css/skosmos.css
```diff
@@ -607,6 +612,58 @@ body {
...
+      pointer-events: none;
+      background: linear-gradient(to right, transparent, var(--sidebar-tab-inactive-bg));
+    }
+
+    #sidebar-tabs.wide .nav-item:last-of-type .nav-link.active::after {
+      background: linear-gradient(to right, transparent, var(--sidebar-tab-active-bg));
+    }
+  }
+
+  @media only screen and (max-width: 900px) {
```
osma, line 653: Could these be replaced with Bootstrap's responsive padding utility classes such as `px-sm-1 px-md-2 py-2` ...?

resource/css/skosmos.css
```diff
@@ -607,6 +612,58 @@ body {
...
+    #sidebar .nav-link {
+      padding: 0.5rem 0.1rem !important;
+    }
+  }
+
+
+  @media only screen and (max-width: 820px) {
```
osma: Ditto, use responsive padding utility classes?

resource/css/skosmos.css
```diff
@@ -607,6 +612,58 @@ body {
...
+    #sidebar-tabs.wide .nav-link {
+      padding: 0.5rem 0.25rem;
+    }
+
+    #sidebar-tabs.wide .nav-item:last-of-type {
+      overflow: hidden;
+    }
+
+    #sidebar-tabs.wide .nav-item:last-of-type .nav-link {
+      justify-content: start !important;
```
osma, line 633: Could this be replaced with Bootstrap's `text-start` utility class?

resource/css/skosmos.css
```diff
@@ -607,6 +612,58 @@ body {
...
+    #sidebar-tabs.wide .nav {
+      flex-wrap: nowrap;
+    }
+
+    #sidebar-tabs.wide #alphabetical {
+      max-width: 40px;
+    }
+
+    #sidebar-tabs.wide .nav-link {
+      padding: 0.5rem 0.25rem;
```
osma, line 625: Replace with Bootstrap padding classes?

resource/css/skosmos.css
```diff
@@ -607,6 +612,58 @@ body {
...
+    }
+
+    #sidebar-tabs.wide .nav-item:last-of-type {
+      overflow: hidden;
```
osma, line 629: Replace with `overflow-hidden` class? Or is it too difficult to determine the last tab in Twig templates?

src/view/sidebar.inc.twig
```diff
@@ -3,9 +3,9 @@
   {% block sidebar %}
     <div class="sidebar-buttons">
       <h2 class="visually-hidden">{{'Sidebar listing: list and traverse vocabulary contents by a criterion' | trans}}</h2>
-      <ul class="nav nav-tabs-no-style nav-justified" id="sidebar-tabs" role="tablist">
+      <ul class="nav nav-tabs-no-style nav-justified{% if vocab.config.sidebarViews | length == 4 and vocab.config.showDeprecatedChanges %} wide{% endif %}" id="sidebar-tabs" role="tablist">
```
osma: Could we change the condition to `>= 4` just in case it's later possible to add more than 4 tabs?

src/view/sidebar.inc.twig
```diff
@@ -3,9 +3,9 @@
...
         {% for view in vocab.config.sidebarViews %}
-          {% if view == 'alphabetical' %} {% set view_trans_text = 'Alpha-nav' %}
+          {% if view == 'alphabetical' %} {% set view_trans_text = (vocab.config.sidebarViews | length == 4) ? 'A-Z' : 'Alpha-nav' %}
```
osma: Ditto, `>= 4` would be more future-proof

resource/css/skosmos.css
```diff
@@ -509,13 +509,18 @@ body {
...
+  }
+
   #main-container.vocab-home #sidebar .nav-link, #main-container.concept #sidebar .nav-link {
     color: var(--sidebar-tab-inactive-text) !important;
     height: 100%;
     display: flex;
     align-items: center;
     justify-content: center;
     transition: none;
+    padding: 0.5rem;
```
osma: Use Bootstrap utility class `p-2` or something along those lines?

osma (changes requested):
The end result looks good, but I'm a bit worried about the amount of new CSS this introduces, and the `!important` rules that could cause problems for users.

I think there could be opportunities to use Boostrap utility classes (e.g. [Spacing](https://getbootstrap.com/docs/5.3/utilities/spacing/), [Text](https://getbootstrap.com/docs/5.3/utilities/text/), [Overflow](https://getbootstrap.com/docs/5.3/utilities/overflow/)) in the HTML instead of CSS rules. I've suggested a few in the comments. Of course there might be some problem with these that aren't immediately obvious to me :) But please at least check if there are ways like this to simplify the CSS.

resource/css/skosmos.css
UnniKohonen: I fixed the rule but it has to be done in CSS as it is only applied on narrow screens

resource/css/skosmos.css
UnniKohonen: The bootstrap breakpoints aren't really granular enough (md is >768px and lg >992px screen width)

resource/css/skosmos.css
UnniKohonen: I will remove this because it is not really needed

resource/css/skosmos.css
UnniKohonen: The rule is dependent on the screen width which the twig template doesn't know

resource/css/skosmos.css
UnniKohonen: Same as above

resource/css/skosmos.css
UnniKohonen: Same as above

Issue #1813, opened 2025-10-01
Match the translation tags in the translation files with those in the Skosmos code and push the changes to Lokalise


miguelvaara:
## Description of the enhancement
The translation files contain unused key–value pairs that need to be removed and the change should finally be pushed to Lokalise.

- find the most efficient way to detect unused tags
- identify / remove them
- push to Lokalise
- document the process

## Who are the users that would benefit from the enhancement and how?
Skosmos users worldwide, as well as translators working on the project.

(Name specific organizations, roles and/or communities if possible, and/or point to relevant discussion elsewhere)

## What new functionalities would the enhancement make possible?
A less cluttered and more manageable translation set.

## Why is the enhancement important?
It is important to support both Skosmos users/deployerss and the translators. By keeping translations clean and relevant the usability and maintainability of the system improves.

(Why should this be prioritized, instead of spending energy on something else? Who would do the work to make this happen?)

PR #1817, opened 2025-10-28
Custom content slots for landing page, about page and footer

### LINKED ISSUE #1815: Landing page: Customizable content boxes
Skosmos 2 included a mechanism to customize the landing page by adding/editing Twig templates called `left.inc` and `right.inc` (as well as `footer.inc` that applied to all pages). These files, if they existed, were simply included into the HTML structure.

For Skosmos 3, we need something similar, but ideally a bit more flexible. For the landing page specifically, the [Skosmos 3.0 Architecture](https://github.com/NatLibFi/Skosmos/wiki/Skosmos-3.0-Architecture) document shows a structural diagram, drawn in the early days of Skosmos 3 development, with boxes called `landing-top-slot`, `landing-start-slot`, `landing-end-slot` and `landing-bottom-slot`. IIRC the idea was that these boxes could be populated with additional content defined by the administrator.

The task is to define these boxes in our DOM structure and to set up a mechanism for including content into them from customizable template files.

In our current landing.twig template, we have some placeholder content ("Welcome to the Skosmos browser demo", "Skosmos is open source and on GitHub") in two boxes in the right sidebar on the landing page. These should be removed from the landing.twig template and replaced with customizable content in separate files.

# Proposed implementation

1. Create a new top level directory `slots` with subdirectories `landing-top`, `landing-start`, `landing-end`, `landing-bottom`
2. Add DOM elements (DIVs most likely) with IDs `landing-top-slot`, `landing-start-slot`, `landing-end-slot` and `landing-bottom-slot` to the Twig templates
3. Populate the above mentioned DIVs by including any Twig template files from `slots/*/*.twig`, in alphabetical order
4. Move the current placeholder content boxes into separate files `slots/landing-end/0-welcome.twig` and `slots/landing-end/1-skosmos-github.twig`

osma:

Implement customizable content slots on the landing and about pages as well as the footer that is displayed on all pages. See #1815.

This PR also removes our own custom content such as the "Welcome to the Skosmos browser demo" and "Skosmos is open source and on GitHub" boxes, the footer, and the about page content. We don't want these to be included with the Skosmos codebase since they are not useful for other instances. They should be stored separately, just like the configuration files.

- Closes #1815

1. Create a new top level directory `custom-templates` with subdirectories `landing-top`, `landing-start`, `landing-end`, `landing-bottom`, `about`, `footer`, `html-head`
2. Add DIV elements with IDs `landing-top-slot`, `landing-start-slot`, `landing-end-slot` and `landing-bottom-slot` to the `landing.twig` template and similar adjustments to `about.twig` and `base-template.twig`
3. With help from WebController that sets a new global variable `customTemplates`, populate the above mentioned DIVs by including any Twig template files from `custom-templates/*/*.twig`, in alphabetical order; in some cases (landing-end-slot and about-slot), show a placeholder message with instructions if no templates have been defined.
4. Remove our own custom content as well as CSS rules that were related to them. I had to reorganize a few CSS rules due to this.
5. Add a PHPUnit test that verifies that the custom templates are found
6. Add Cypress tests that verify that the placeholder message is shown (on the landing page) and custom content is displayed (on the about page).

Here is an example of the placeholder message shown on the landing page:

image

* needs wiki documentation
* the new translated messages (the two sentences of the placeholder message) are not yet pushed to Lokalise nor translated to other languages

## Checklist


osma:
TODO:

- [x] investigate why the skosmos-github box is rendered differently (no full height)
- [x] define additional slots at least for about and footer (move the current footer content to a custom template)
- [x] move vocab-list inside landing-start-slot DIV

miguelvaara (commented):
**Some general comments:**
The note _Adding an empty Twig file will remove this message_ is clear and straightforward instruction for removing a message that the developer might consider unnecessary. It could hardly be any easier to remove — well done!

With this approach we are getting closer to a publishing system — which is a good thing within reasonable limits. I thing we are still within those limits, since the easy customizability of Skosmos’s structure and appearance is surely welcome from the developers’ perspective.  And after all, developers still have the possibility to modify Skosmos as they want.

src/controller/WebController.php
```diff
@@ -70,6 +70,33 @@ public function __construct($model)
...
+    }
+
+    /**
+     * Find any custom templates from the given directory and return them as an array.
+     * @param string $dir path of custom templates directory
+     * @return array array of custom template filenames, keyed by slot
+     */
+    public function findCustomTemplates($dir)
+    {
+        $customTemplateSubDirs = glob($dir . '/*', GLOB_ONLYDIR);
```
miguelvaara, line 86: Would it be a good idea to check whether the directory exists before referring to it (checking or some throwing of the error)? I am not sure myself 🤔. In practice, no directories are actually required but at some level it feels necessary to handle that kind of situation. Just wondering about it as in similar situations there is often a check like that. It is probably easier to debug (like a wrong directory name - typo) if the errors just pop up.

osma: The `glob` function simply returns an empty array if the directory doesn't exist, so I don't think it makes sense to add an additional check. However, I added a PHPUnit test that verifies that findCustomTemplates doesn't break if the directory doesn't exist. (The only reason the $dir is passed as parameter here is to make it easier to test this method using unit tests. It would be difficult to test it if the path was hardcoded here.)

osma:
Wrote up some wiki documentation: https://github.com/NatLibFi/Skosmos/wiki/Custom-templates

I also resolved the merge conflict for `messages.en.json` caused by #1813 .

miguelvaara (approved):
We will probably need to discuss the length of translation message keys at some point later on. The recommendation is that in the default file the key and the message are the same, but that can sometimes make the code rather “interesting” ;-) to read. So the question is -- do we want to do it by the book or make it look good?

Push to Lokalise after the merge?

Everything looks great!

Post scriptum:

A small curiosity, nothing that needs action:

“Du kan ta bort detta meddelande genom att lägga till en tom twig-fil.”

Might “det här” (instead of “detta”) sound a bit too Swedish from a Finland-Swedish perspective? Maybe - maybe not? No need to get stuck on this, I just found myself wondering about it.

osma:
Translations have already been pushed to Lokalise (including the long message). We can talk about that but I think it's out of scope for this PR. Likewise the Swedish translation: we can change that in Lokalise later if we need to; I think many such adjustments will still be needed down the line.

Merging this as it is now, thank you for the review!

PR #1819, opened 2025-10-29
Fix flaky vocab-home Cypress test under GitHub Actions CI

osma:

One of the vocab-home Cypress tests started failing very often (more than 50%) under GitHub Actions CI. It is a test that verifies that a partial page load is performed after clicking on the hierarchy. Part of the test checks that a mapping to Wikidata is shown properly. This fails, because the expected Wikidata label "fish" is not being shown.

I couldn't replicate the problem locally so I assume this has something to do with traffic between GitHub Actions and Wikidata. Maybe Wikidata blocks some queries?

This PR adjusts the test slightly so that either a label or Wikidata URI is an acceptable value. Now it should work under GitHub Actions as well.

n/a

* robustness fix: accept either wikidata label or URI in vocab-home cypress test

It was a bit mysterious so not 100% sure this fixes the problem for good.

## Checklist

### DIFF
--- a/tests/cypress/template/vocab-home.cy.js
+++ b/tests/cypress/template/vocab-home.cy.js
@@ -65,7 +65,7 @@ describe('Vocabulary home page', () => {
     // check the second mapping property name
     cy.get('.prop-mapping h2', {'timeout': 15000}).eq(0).contains('Exactly matching concepts')
     // check the second mapping property values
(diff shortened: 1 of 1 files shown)


osma:
The Cypress tests have now passed 6 out of 6 times in CI. Merging this.

PR #1820, opened 2025-10-30
Remove unnecessary translation keys and messages

### LINKED ISSUE #1813: Match the translation tags in the translation files with those in the Skosmos code and push the changes to Lokalise

miguelvaara:
We do not want to remove key–message pairs that are likely to be needed for translation again, as that would cause unnecessary and frustrating rework and slow down the completion of translations.

The situation is quite good (around 10 removed), but after the release of Skosmos 3, the list needs to be reviewed again, since some of the key–message pairs used in Skosmos 2 will probably end up unused.

#1813 
- Closes #1813 

### When a key–message pair in `messages.en.json` should be removed

A key–message pair will be removed **only if**:
- it is not used in Skosmos 3  
- and it was not used in Skosmos 2 either (the probability of inheritance to Skosmos 3 is rather high) 
- the key–message pair is not generic enough to being propably used later 

### Searching for and evaluating usage cases

Open the attached file `messages.en-checklist.json`, where:

- `N` → the pair will be **removed**  
- `?` → **uncertain**, difficult to decide  

In the Skosmos root directory, run:

```
grep -nr --include='*.js' --include='*.php' --include='*.twig' "[search text]" [skosmos 2 folder] | grep trans
grep -nr --include='*.js' --include='*.php' --include='*.twig' "[search text]" . | grep trans
```
#### List all found translation keys and their states

**Note:**  
The following command also lists as **unused** those key-message pairs in the language file whose key and message are identical (even if the message itself is used)

```
php bin/console debug:translation en
```

#### Go through all translation keys found in PHP, JS, and Twig code

```
bin/console translation:extract en --dump-messages
```
When the translation file is finalized and merged into main, it can be pushed to Lokalise.

---

-

## Checklist

[messages.en-checklist.json](https://github.com/user-attachments/files/23235497/messages.en-checklist.json)


resource/translations/messages.en.json
```diff
@@ -95,7 +93,6 @@
     "foaf:page": "Page",
     "foreign prefLabel help": "Terms for the concept in other languages.",
     "foreign prefLabels": "In other languages",
-    "from all": "from all",
```
osma, line 98: I think we might need this in the global search widget, even though the current PR #1799 doesn't seem to use it currently.

resource/translations/messages.en.json
```diff
@@ -160,7 +156,6 @@
     "skosext:DeprecatedConcept": "Deprecated concept",
     "skosext:partOf": "Is part of",
     "skosext:partOf_help": "Whole which the concept is a part of.",
-    "skosmos:marcSourceCode": "MARC source code",
```
osma, line 163: This is used in e.g. YSO concept scheme metadata as an RDF property. I think it should not be dropped. Instead it should be added to extra-msgids.twig.

resource/translations/messages.en.json
miguelvaara: Done

resource/translations/messages.en.json
miguelvaara: Done

osma (approved):
LGTM!

resource/translations/messages.en.json
miguelvaara: done

resource/translations/messages.en.json
miguelvaara: done

PR #1823, opened 2025-11-05
Render hierarchy with concept schemes

UnniKohonen:

Hierarchies are not rendered in vocabs with multiple concept schemes. This PR fixes rendering these hierarchies

- Part of #1521

- Add `showConceptSchemesInHierarchy` to `SKOSMOS` object
- Render concept schemes as top concepts in vocabs with `showConceptSchemesInHierarchy` set to true
  - Use `/{vocid}/` method when fetching top concepts on vocab home page
  - Use `/{vocid}/topConcepts/` when loading children of a concept scheme after button click in hierarchy tree
  - Use `/{vocid}/` when fetching top concepts on concept page 
  - Use `/{vocid}/topConcepts/` when opening concept scheme page to fetch its children
- Fix `getConceptURL` method to work when uri space is same as current uri

Addresses requirement 6b in #1521

## Checklist

### DIFF
--- a/resource/js/tab-hierarchy.js
+++ b/resource/js/tab-hierarchy.js
@@ -277,12 +277,12 @@ function startHierarchyApp () {
         }
       },
       createConceptNode (concept, isOpen = false, children = []) {
-        return { 
+        return {
           uri: concept.uri,
           label: concept.label || concept.prefLabel,
           hasChildren: concept.hasChildren,
-          children: children,
-          isOpen: isOpen,
+          children,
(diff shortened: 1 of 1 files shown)


osma (approved):
Looks good. I tested this on several vocabularies with different kinds of hierarchical structures and couldn't find any obvious problems.

I noticed that there are some `console.log` debug messages in the JS code left. But there are some also in other JS components. I think we should aim to clean them up at some point, but that can be done a bit later...

PR #1825, opened 2025-11-06
Truncate long property value lists on concept page

### LINKED ISSUE #1795: truncate property value lists that are longer than set number of items with "show all X values" button
This is a repeat of #833 (PR #969) but for Skosmos 3.

For example these concept pages have long property lists that should be truncated by default, with a "show all X values" link to expand:

- YSO: [clothes](https://test.dev.finto.fi/yso/en/page/p3594) (26 skos:narrower values)
- YKL/PLC: [38.55 Erityisopetus. Erityiskasvatus. Erityiskoulut](https://test.dev.finto.fi/ykl/fi/page/38.55) (25 keywords and 25 terms in other languages)

osma:

Overwhelmingly long lists of property values on the concept page should be truncated, see #1795. This functionality exists in Skosmos 2, and this PR implements the same feature for Skosmos 3.

I looked at the jQuery-based implementation of this functionality in Skosmos 2 and adapted the same logic for Skosmos 3, but using vanilla JS instead. There is also a Cypress test that verifies that it works both in the initial page load and partial page load cases.

Example of the result (on the page http://localhost/Skosmos/ykl/fi/page/38.55 ):

kuva

- Closes #1795
- part of #1484 

- add vanilla JS code that truncates long lists of property values (by adding CSS classes to the elements), triggered on initial page load & partial page load
- add CSS rules that hide/reveal property values as appropriate
- add Cypress tests to verify the above

- The truncation limit is hardcoded at 15, as in Skosmos 2. It would take more work to make it configurable.
- This PR only truncates regular property values, e.g. `skos:narrower`. It does not truncate the "terms in other languages" section, unlike Skosmos 2, where also these were truncated. In Skosmos 3, the DOM structure of the "in other languages" section is very different than regular properties, so the same code doesn't work for both.

## Checklist

### DIFF
--- a/tests/cypress/template/concept-full-vs-partial.cy.js
+++ b/tests/cypress/template/concept-full-vs-partial.cy.js
@@ -87,6 +87,24 @@ describe('Concept page, full vs. partial page loads', () => {
       // If that happens, make sure the browser window has focus and re-run the test.
       cy.window().its('navigator.clipboard').invoke('readText').then((result) => {}).should('equal', '33.2');
     })
+    it('narrower concepts: long lists are truncated / ' + pageLoadType, () => {
+      // Go to "periods of time" concept page which has 21 narrower concepts
+      if (pageLoadType == "full") {
+        cy.visit('yso/en/page/p4035') // go directly to "periods of time"
+      } else {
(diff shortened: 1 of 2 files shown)


UnniKohonen (approved):
LGTM!

PR #1830, opened 2025-11-24
Upgrade PHP-CS-Fixer, add support for PHP 8.4

osma:

Keep Skosmos up to date by supporting more recent PHP releases.

n/a

* Upgrade PHP-CS-Fixer to 3.90 (at least 3.80 is needed for PHP 8.4 support)
* Update CI configuration to run PHPUnit tests under PHP 8.4 as well
* Update list of supported PHP versions in README

none

## Checklist


osma:
The tag is indeed a problem, but it has already (around the time of the 3.0-beta.2 release) been recreated in GitHub so the problem should only appear in local git clones.

Merging this PR.

Issue #1752, opened 2025-01-29
Epic: CSS/Style (General)


miguelvaara:
## Who are the users that would benefit from the enhancement and how?

(Name specific organizations, roles and/or communities if possible, and/or point to relevant discussion elsewhere)

## Why is the enhancement important?

(Why should this be prioritized, instead of spending energy on something else? Who would do the work to make this happen?)

PR #1831, opened 2025-11-25
Refactor Skosmos 3 CSS

### LINKED ISSUE #1597: CSS audit needed
With Skosmos main CSS file approaching the thousand line mark, we need to go through the individual definitions to spot inconsistencies

It would be easiest to bring the style sheet in line during the front update but before we start to finalize small details. If we are to define a colour theme picker, the CSS refactoring should be done before it.

## What tools we should consider?
 * https://purifycss.online/
 * https://uncss-online.com/
 * https://developer.chrome.com/docs/devtools/coverage
 
 ## What approaches we can use
 * https://github.com/NatLibFi/Skosmos/issues/1320
 * This seemed relevant, as we have a base template, additional templates, modules and themes: https://smacss.com/book/categorizing/

We _can_ proceed with the old CSS file and add to it when needed. It would just become even harder to maintain than currently, so we maybe should pay CSS some attention as a whole sooner rather than later.

UnniKohonen:

Current Skosmos 3 CSS needs to refactored. The code is badly organized and there is a lot of unused and unnecessary code. This PR attempts to fix these issues.

- Closes #1597
- Part of #1752

- Fix errors in code
- Remove unused rules
- Remove duplicate rules
- Remove unnecessary definitions
- Reorganize and group code according to page type and page section
- Fix HTML for about, feedback and error pages

- HTML code is not majorly altered in this PR and therefore no consistent CSS framework is used (for example [SMACSS](https://smacss.com/))
- Color schemes are not addressed by this PR, those will have to done separately
- There are a few unused rules left in the CSS that either have incorrect selectors (vocab statistics) or relate to features that will be added later (striped background images)
- Color definitions are not consistently applied, this will have to fixed when implementing color schemes
- There are probably many unnecessary individual definitions remaining in the code

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -431,6 +431,7 @@ header .container-fluid .row {
 #search-wrapper ul::-webkit-scrollbar {
   -webkit-appearance: none;
   -moz-appearance: none;
+  appearance: none;
 }
 
 #search-wrapper ul::-webkit-scrollbar:vertical {
@@ -838,6 +839,7 @@ header .container-fluid .row {
 #sidebar .sidebar-list::-webkit-scrollbar {
   -webkit-appearance: none;
   -moz-appearance: none;
+  appearance: none;
 }
 
 #sidebar .sidebar-list::-webkit-scrollbar:vertical {


miguelvaara (commented):
Based on the changes made, the following have been reviewed in particular:

1
- Sidebar
- Search
- about
- feedback

2
The changes were compared visually with the layouts created by Hahmo (colors and sizes were not measured), and everything appears to be in order.

3
Frontend tests run – all OK

4
Accessibility tests run – all OK

5
The points that raised questions were the following:

6
The variables `--sidebar-scrollbar-thumb` and `--vocab-bg` do not have fallback colors defined, which could cause issues in older browsers that do not support CSS variables. We agreed together that we will not provide support for such old browsers. In practice all common browsers released after spring 2017 support this.

7
-.tooltip-html-content:after {
+.tooltip-html-content::after {
= Okay 

8
A more detailed reviewing would not maybe needed so based on a reasonably thorough review the work looks good - excellent! 👍  If you decide to fix the appearance issue (on the other comments) - great but if not, that is probably fine as well. In any case, I think this can be merged.

resource/css/skosmos.css
```diff
@@ -65,1250 +69,964 @@ body {
...
+} 
 
-  #search-wrapper .autocomplete-result span.result {
-    color: var(--vocab-link);
-  }
+/* Setting scrollbar style for Webkit based browsers (Chrome, Safari, Edge, etc) */
+#search-wrapper ul::-webkit-scrollbar {
+  -webkit-appearance: none;
+  -moz-appearance: none;
+}
```
miguelvaara, line 435: The proposed change does not have a major impact on the current implementation, but the _appearance: none_ approach is commonly used. I consider it as "your choice" and I think you can go for a merge

appearance: none; removes the native browser styling from form elements and UI controls.

[https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/appearance](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/appearance
)

```
#search-wrapper ul::-webkit-scrollbar {
  -webkit-appearance: none;
  -moz-appearance: none;
}

->

#search-wrapper ul::-webkit-scrollbar {
  appearance: none;
  -webkit-appearance: none;
  -moz-appearance: none;
}
```

resource/css/skosmos.css
```diff
@@ -65,1250 +69,964 @@ body {
...
+} 
 
-  nav#concept-breadcrumbs li, nav#concept-breadcrumbs li.collapse.show {
-    display: inline;
-  }
+/* Setting scrollbar style for Webkit based browsers (Chrome, Safari, Edge, etc) */
+#sidebar .sidebar-list::-webkit-scrollbar {
+  -webkit-appearance: none;
+  -moz-appearance: none;
+}
```
miguelvaara, line 843: The proposed change does not have a major impact on the current implementation, but the _appearance: none_ approach is commonly used. I consider it as "your choice" and I think you can go for a merge

appearance: none; removes the native browser styling from form elements and UI controls.

[https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/appearance](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/appearance
)

```
/* Setting scrollbar style for Webkit based browsers (Chrome, Safari, Edge, etc) */
#sidebar .sidebar-list::-webkit-scrollbar {
  -webkit-appearance: none;
  -moz-appearance: none;
}

->

/* Setting scrollbar style for Webkit based browsers (Chrome, Safari, Edge, etc) */
#sidebar .sidebar-list::-webkit-scrollbar {
  appearance: none;
  -webkit-appearance: none;
  -moz-appearance: none;
}
```

PR #1835, opened 2025-11-26
Encode concept URIs in API calls with URLSearchParams (fix hash namespace issues)

### LINKED ISSUE #1828: Mapping endpoint call fails with 500

A concept page, for example /myvocab/en/page/001?clang=en-GB

## Description of the problem

Open the browser to one concept in the vocabulary. The box underneath the main concept display box shows a loading icon and never loads anything. The underlying concept URI in the SKOS data is https://my.skosmos.domain/myvocab#001.

Checking the request sent, I have this:

```plaintext
https://my.skosmos.domain/rest/v1/myvocab/mappings?uri=https://my.skosmos.domain/myvocab
```

The URI parameter looks incomplete, and does not contain the term ID. Further checking the log:

```plaintext
PHP Fatal error:  Uncaught TypeError: Vocabulary::getConceptInfo(): Argument #2 ($clang) must be of type string, null given, called in .../src/controller/RestController.php on line 703 and defined in .../src/model/Vocabulary.php:452
Stack trace:
#0 .../src/controller/RestController.php(703): Vocabulary->getConceptInfo('https://my.skosmos...', NULL)
#1 .../src/rest.php(63): RestController->mappings(Object(Request))
#2 {main}
  thrown in .../src/model/Vocabulary.php on line 452
```

Actual domain and file paths have been redacted. I think the issue is that my SKOS vocabulary uses hashtag style namespace, and when sending the request for getting mapping, skosmos did not escape the hashtag correctly, causing the browser to truncate everything after the hashtag, which caused the backend to not receive parameters like clang, or the term ID itself.

Chromium: 140.0.7339.186

osma:

It was reported in #1828 that concept mappings are not shown on concept pages where the concept URI contains a `#` character. This turned out to be an URL encoding problem that affects many API calls and also breaks e.g. the hierarchy display.

This PR changes the way REST API URLs are constructed to ensure that all arguments are properly URL encoded. String concatenation is replaced with [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) which handles encoding automatically.

The PR also includes a small test vocabulary which uses a hash-style namespace, and a Cypress test that verifies that mappings are loaded for these concepts (and indirectly also that the hierarchy is working).

- Closes #1828

- change all URL construction expressions in JS front-end code to use URLSearchParams
- add `hash` test vocabulary
- add Cypress test that verifies that the mappings and hiearchy are working for the hash vocabulary concepts

I tried to dig through all our JS front-end code looking for problematic ways to construct URLs and replaced them all with URLSearchParams. I don't think I've missed any, but it's possible. I'm a bit unsure about `get-concept-url.js` ; it's quite ugly in its current form and doesn't use URLSearchParams, but it seems to me that it is still encoding URL parameters properly using `encodeURIComponent` calls, so I didn't touch it.

## Checklist

### DIFF
--- a/resource/js/concept-mappings.js
+++ b/resource/js/concept-mappings.js
@@ -29,7 +29,7 @@ const conceptMappingsApp = Vue.createApp({
         lang: window.SKOSMOS.lang,
         clang: window.SKOSMOS.content_lang
       })
-      const url = `rest/v1/${window.SKOSMOS.vocab}/mappings?${params.toString()}`
+      const url = `rest/v1/${window.SKOSMOS.vocab}/mappings?${params}`
 
       fetchWithAbort(url, 'concept')
         .then(data => {


osma:
@miguelvaara it is used there (one line below the one you linked)

miguelvaara:
> @miguelvaara it is used there (one line below the one you linked)

You right - my bad - sorry!

osma:
I dropped the extra `toString()` from one of the expressions.

Note that since the search URL construction functionality is implemented in a different style (string concatenation, not string interpolation expressions), `toString()` is still needed there (I think) and I didn't touch that code in the PR as it already handled URL encoding properly.

tests/cypress/template/concept-full-vs-partial.cy.js
```diff
@@ -144,6 +144,30 @@ describe('Concept page, full vs. partial page loads', () => {
...
+
+      // check that we have some mappings
+      cy.get('#concept-mappings').should('not.be.empty')
+      // check that spinner does not exist after load
+      cy.get('#concept-mappings i.fa-spinner', {'timeout': 15000}).should('not.exist')
+
+      // check the first mapping property name
+      cy.get('.prop-mapping h2').eq(0).contains('Exactly matching concepts')
+      // check the first mapping property values
+      cy.get('.prop-mapping').eq(0).find('.prop-mapping-label').eq(0).contains('number sign')
```
miguelvaara, line 164:
```
      cy.get('.prop-mapping').eq(0).find('.prop-mapping-label').eq(0).contains('number sign')
```
I cannot get it to pass. It fails with the following error because it expects the string ‘number sign’.

containsnumber sign / AssertionError

Timed out retrying after 4000ms: Expected to find content: 'number sign' within the element: <div.col-5.prop-mapping-label> but never did.

tests/cypress/template/concept-full-vs-partial.cy.js:164:71

cy.get('.prop-mapping').eq(0).find('.prop-mapping-label').eq(0).contains('number sign')

osma: `skosmos:loadExternalResources` defaults to true, so I don't think that can be the problem.

PR #1840, opened 2025-11-27
Update library dependencies

### LINKED ISSUE #1837: Update library dependencies

To reduce and prevent technical debt, we should check our libraries regularly and aim to keep them as up to date as possible -- taking security into account but avoiding unnecessary and heavy "update parties".

All those who directly or indirectly maintain or use Skosmos.

The improvement does not add new functionality, but makes it possible to use Skosmos more securely and with fewer errors.

It ensures that the existing features work properly and remain secure.

miguelvaara:
To keep Skosmos up to date, secure and fully working

- Closes #1837 

All the tests pass except for the version number check in composer.json, which does not matter at this stage. I merged the changes that likely will come with your PR #1731

How does this approach look to you?

I am not sure if the merge with the PR #1731 was the best move to do ...

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,8 +6,6 @@
   "version": "3.0-dev",
   "require": {
     "sweetrdf/easyrdf": "~1.18.2",
-    "symfony/polyfill-php80": "1.*",
-    "symfony/polyfill-php81": "1.*",
     "twig/twig": "3.22.*",
     "willdurand/negotiation": "3.1.*",
     "punic/punic": "3.8.*",
--- a/package.json
+++ b/package.json
@@ -16,9 +16,9 @@
     "vue": "^3.5.25"
   },
   "devDependencies": {
-    "axe-core": "^4.10.*",
-    "cypress": "^13.17.*",
-    "cypress-axe": "^1.5.*",
+    "axe-core": "^4.11.*",
+    "cypress": "^15.7.*",
+    "cypress-axe": "^1.7.*",
     "standard": "^17.1.*"
   }
 }


osma:
Quick comments:

1. I think we can drop `symfony/polyfill-php80` and `symfony/polyfill-php81`, they are only useful for PHP versions below 8.0 / 8.1, and we are already requiring PHP 8.1 as a minimum.
2. What about [axe-core](https://www.npmjs.com/package/axe-core) and [cypress-axe](https://www.npmjs.com/package/cypress-axe), it looks like newer versions are available?
3. [Cypress](https://www.npmjs.com/package/cypress) also has much newer version available (15.7.0), should we try upgrading to that as well or is that too big a leap? (Could also be done in a separate PR if it's difficult)

miguelvaara:
> Quick comments:
> 
> 1. I think we can drop `symfony/polyfill-php80` and `symfony/polyfill-php81...

**Instructions for the design review:**

Because Cypress was updated by two major versions, Node.js had to be updated to version 20. You can enable it temporarily in your shell like this:

`source ~/.nvm/nvm.sh && nvm use 20 && node --version && npm --version`

If everything works and you want to make Node 20 the default permanently:

`source ~/.nvm/nvm.sh && nvm alias default 20 && nvm current`

And if you change your mind and want to go back to the previous version, for example version 18, you obviously can do it like this:

`source ~/.nvm/nvm.sh && nvm use 18 && node --version && npm --version`

Cypress shouts at the following:
image

PR #1843, opened 2025-11-27
Truncate list of terms in other languages

### LINKED ISSUE #1838: Truncate list of terms in other languages with "show all X values" button

Issue #1795 / PR #1825 implemented truncation of long lists of property values on the concept page. However, truncation of the list of terms in other languages was not implemented in that PR:

> This PR only truncates regular property values, e.g. skos:narrower. It does not truncate the "terms in other languages" section, unlike Skosmos 2, where also these were truncated. In Skosmos 3, the DOM structure of the "in other languages" section is very different than regular properties, so the same code doesn't work for both.

The list of terms should also be truncated, for example YKL/PLC: [38.55 Erityisopetus. Erityiskasvatus. Erityiskoulut](https://test.dev.finto.fi/ykl/fi/page/38.55) (25 keywords and 25 terms in other languages)

Users of e.g. YKL/PLC in Finnish libraries. YKL uses lots of altLabels.

Better UX for concept pages.

UX is important.

osma:

See #1838. This implements truncation for the list of terms in other languages on the concept page:

Here is how the result looks like on the YKL/PLC page [32.1 Political systems](http://localhost/Skosmos/ykl/en/page/32.1):
kuva

- Closes #1838
- part of #1484

* extend the code in `truncate-property-values.js` to also handle lists of terms in other languages, which has a somewhat different DOM structure than regular properties
* add a test vocabulary `altlabel` which contains a concept that has very many English and Finnish language altlabels
* add a Cypress test that tests the truncation functionality with the help of the `altlabel` vocabulary

1. The limit on the number of items shown is hardcoded at 15, as before. This makes sense for regular properties, but for these terms in other languages, it could perhaps be a bit smaller than that.
2. One of the Cypress tests in `global-search-bar.cy.js` broke due to the addition of the new `altlabel` vocabulary. For some reason the `testUnknownPropertyOrder` vocabulary disappeared from the vocabulary selection menu and was replaced by `altlabel`. I don't quite understand what happened here; why does the menu only show two vocabularies and not e.g. three? I modified the Cypress test accordingly, but there could be a bug hiding somewhere behind this strange behavior of the vocabulary selection menu.

## Checklist

### DIFF
--- a/tests/cypress/template/global-search-bar.cy.js
+++ b/tests/cypress/template/global-search-bar.cy.js
@@ -14,14 +14,14 @@ describe('Global search bar', () => {
 
 
     cy.get('#vocab-list li').eq(0).find('input[type="checkbox"]').check({ force: true })
-    cy.get('#vocab-selector .vocab-dropdown-btn').should('contain.text', 'testUnknownPropertyOrder')
+    cy.get('#vocab-selector .vocab-dropdown-btn').should('contain.text', 'altlabel')
 
     cy.get('#vocab-list li').eq(1).find('input[type="checkbox"]').check({ force: true })
-    cy.get('#vocab-selector .vocab-dropdown-btn').should('contain.text', 'testUnknownPropertyOrder')
+    cy.get('#vocab-selector .vocab-dropdown-btn').should('contain.text', 'altlabel')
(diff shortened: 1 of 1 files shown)


joelit (approved):
LGTM! IMO duplicated lines in cypress tests are fine, and SonarQube should be ignored for those.

PR #1844, opened 2025-11-27
Translated language name lists in global & vocab search bar

joelit:

Previously, we have used a mock-up translation of language lists in the search bar. This implements a punic-generated list of language names for the search bar dropdowns

- The feature is described in the search bar epics #1489 and #1514

* When in global search results page, or landing page, use the list of every available language in the dataset, translated by punic
* When in vocablulary-related page (vocab-home, concept, or vocab-search-result), use the list of languages configured for the vocabulary, translated by punic
* Remove the mock-up translations from skosmos-object
* Move some of the logic up to the controller level
* Format the list with 'all languages' option
* Update cypress tests accordingly

* Support for adding additional languages needs to be explored (e.g. for Northern Sami)

## Checklist

### DIFF
--- a/resource/js/global-search.js
+++ b/resource/js/global-search.js
@@ -51,7 +51,7 @@ function startGlobalSearchApp () {
     mounted () {
       this.languages = window.SKOSMOS.languageOrder
       this.selectedLanguage = this.updateSearchLang()
-      this.languageStrings = this.formatLangStrings()
+      this.languageStrings = this.formatLanguages()
       this.vocabStrings = window.SKOSMOS.vocab_list
     },
     watch: {
@@ -99,10 +99,10 @@ function startGlobalSearchApp () {
             }
           })
       },
-      formatLangStrings () {
-        const langStrings = window.SKOSMOS.contentLangStrings
+      formatLanguages () {
+        const languages = window.SKOSMOS.contentLanguages
         const anyLanguageEntry = { all: this.anyLanguage }
-        return { ...langStrings, ...anyLanguageEntry }
+        return { ...languages, ...anyLanguageEntry }
       },
       formatSearchUrlParams () {
         const params = new URLSearchParams({ q: this.searchTerm })
--- a/resource/js/vocab-search.js
+++ b/resource/js/vocab-search.js
@@ -36,7 +36,7 @@ function startVocabSearchApp () {
     mounted () {
       this.selectedLanguage = this.parseSearchLang()
       this.searchCounter = 0 // used for matching the query and the response in case there are many responses
-      this.languageStrings = this.formatLangStrings()
+      this.languageStrings = this.formatLanguages()
       this.renderedResultsList = []
       this.showNotation = window.SKOSMOS.showNotation
     },
@@ -75,10 +75,10 @@ function startVocabSearchApp () {
             }
           })
       },
-      formatLangStrings () {
-        const langStrings = window.SKOSMOS.contentLangStrings
+      formatLanguages () {
+        const languages = window.SKOSMOS.contentLanguages
         const anyLanguagesEntry = { all: this.anyLanguages }
-        return { ...langStrings, ...anyLanguagesEntry }
+        return { ...languages, ...anyLanguagesEntry }
       },
       formatSearchTerm () {
         if (this.searchTerm.includes('*')) { return this.searchTerm }
--- a/src/view/scripts.inc.twig
+++ b/src/view/scripts.inc.twig
@@ -39,7 +39,7 @@ window.SKOSMOS = {
   },
   {%- endif -%}
   "baseHref": "{{ BaseHref }}",
-  "contentLangStrings": { {% for code, name in content_languages %}"{{ code }}": "{{ name }}"{% if not loop.last %}, {% endif %}{% endfor %} }
+  "contentLanguages": { {% for code, name in content_languages %}"{{ code }}": "{{ name }}"{% if not loop.last %}, {% endif %}{% endfor %} }
 }
 </script>
 


src/controller/WebController.php
```diff
@@ -170,16 +170,17 @@ public function invokeLandingPage($request)
         // set template variables
         $categoryLabel = $this->model->getClassificationLabel($request->getLang());
         $sortedVocabs = $this->model->getVocabularyList(false, true);
-        $langList = $this->model->getLanguages($request->getLang());
+        $searchLangList = $this->model->getLanguages($request->getLang());
         $listStyle = $this->listStyle();
+        //$languageLabels = $this->getLanguageLabels($request->getLang());
```
osma: Remove commented out code

osma (changes requested):
Looks good, mostly things are working, but I have a few suggestions for changes:

1. The "All languages" option should be last, after the languages, not first, just like in Skosmos 2.
2. The "All languages" option could be called "Any language" in English, just like in Skosmos 2.
3. I suggest renaming the variables `$searchLangList` and `$vocabSearchLangList` (php), `search_lang_list` (twig) and `searchLangStrings` (SKOSMOS object). These are indeed used for the search language dropdown (along with the "all"/"any" option), but I think it would be better to call these "content languages" so they align with e.g. the `content_lang` and `clang` variable names. For example `searchLangStrings` could be called `contentLanguages`.
4. The SKOSMOS object now has both `languageOrder` and `searchLangStrings` fields, for slightly different reasons, but both containing a list of languages. Would it be possible to get rid of `languageOrder` and instead order `searchLangStrings`/`contentLanguages` in the right order?
5. This isn't a new bug in this PR (found by @UnniKohonen ), but I think it needs to be fixed: in the vocabulary search bar, selecting the "All languages" (or "Any language") option causes SKOSMOS.content_lang to be set to "all", which breaks the sidebar components that expect a language tag and not this special value: https://github.com/NatLibFi/Skosmos/blob/072d67f48611316ae2f0fc8989c605beb7b5143a/resource/js/vocab-search.js#L198

src/controller/WebController.php
```diff
@@ -158,6 +158,35 @@ private function listStyle()
...
+    /**
+    * Renders the list of supported languages from vocabulary config in order.
+    * The ordering is done according to the language order parameter in vocabulary config if such exists
+    * @param Vocabulary $vocab the vocabulary object
+    * @return array with language codes as keys and language labels as values
+    */
+    private function parseVocabularyLanguageOrder($vocab)
+    {
+        $vocabContentLangList = array_flip($vocab->getConfig()->getLanguages());
+        $languageOrder = $vocab->getConfig()->getLanguageOrder(null);
```
osma: Why use `null` here as an argument? It should be the current content language

src/controller/WebController.php
```diff
@@ -158,6 +158,35 @@ private function listStyle()
...
+
+        if (is_null($languageOrder)) {
```
osma: Can `$languageOrder` ever be null? It seems to me that `getLanguageOrder` always returns an array.

src/controller/WebController.php
```diff
@@ -158,6 +158,35 @@ private function listStyle()
...
+
+        $tmpList = [];
+
+        foreach ($languageOrder as $vocLang) {
+            if (isset($vocabContentLangList[$vocLang])) {
+                $tmpList[$vocLang] = $vocabContentLangList[$vocLang];
+                unset($vocabContentLangList[$vocLang]);
+            }
+        }
+        $sortedContentLangList = $tmpList + $vocabContentLangList;
```
osma: Could just return this directly without assigning to a variable.

src/controller/WebController.php
joelit: If an argument is passed for the getLanguageOrder, this language is put in the front of the list of returned language codes. If null is passed, the language order is returned as it's declared in the configuration. So, if the language parameter is passed as a parameter in this piece of code, we will get a situation where in the UI the current language is always shown as the first option in the language dropdown. I'm sure that's not a feature we want. I thought the expected behaviour is to return the languages in the order the skosmos:languageOrder is configured in config.ttl.

src/controller/WebController.php
joelit: Ah good. I was thinking of a situation where the languageOrder is not configured in config.ttl.

src/controller/WebController.php
joelit: It could. I added the assigning of a variable name for clarity of what's happening. But obscuring the code for brevity is always a trade-off.

src/controller/WebController.php
osma: I see. Practically it makes sense to get the desired end result, but feels like a hack. The method is declared as `getLanguageOrder($clang)` with a docstring `@param string $clang`. I suggest changing it so it explicitly allows `null` values (maybe $clang could default to null, so it could be just omitted in this call) and the method should not return the `null` value within the array as it currently does.

src/controller/WebController.php
osma: OK, I leave it up to you. (SonarCloud also complains about this)

tests/WebControllerTest.php
```diff
@@ -277,4 +277,13 @@ public function testFindCustomTemplatesDirectoryNotFound() {
         $this->assertEmpty($result);
     }
 
+    public function testParseVocabularyLanguageOrder() {
+        $request = new Request($this->model);
+        $vocab = $this->model->getVocabulary('multilang');
+
+        $languageOrder = $this->webController->parseVocabularyLanguageOrder($vocab);
+        $expected = array( 'en' => 'anglais', 'fi' => 'finnois', 'sv' => 'suédois', 'de' => 'allemand' );
+        $this->assertEquals($expected, $languageOrder);
```
osma: This test doesn't actually verify the order of the array. `assertEquals` will only check that the array has the same keys and values, but not their order.

I suggest changing to `assertSame`, which also checks the order. You will also have adjust the expected order so that the test will pass with the given vocabulary configuration `skosmos:fallbackLanguages ( "fi" "en" "sv" "de" )`:

```php
        $expected = array( 'fi' => 'finnois', 'en' => 'anglais', 'sv' => 'suédois', 'de' => 'allemand' );
        $this->assertSame($expected, $languageOrder);
```

osma (approved):
There's a problem with the new PHPUnit test. I suggested how to fix it in a comment.

Otherwise looks good, though I dislike the variable/method names `*LangStrings` (within the SKOSMOS object and in Vue code); I don't think the value being a string is a relevant aspect and I would rather rename them to `*Languages`, like this:

- `formatLangStrings` -> `formatLanguages`
- `langStrings` -> `languages`
- `contentLangStrings` ->`contentLanguages`

But I've already nitpicked a lot about variable names a lot in the previous round so I leave this up to you to decide. :wink: 

After considering these, this PR can be merged. It's a big improvement for the search bars!

PR #1850, opened 2025-12-03
Update topbar navigation (inc. language links) during partial page load

### LINKED ISSUE #1848: Changing UI language after partial page load opens wrong page

https://test.dev.finto.fi/yso/en/

## Description of the problem

1. Go to YSO vocab home page: https://test.dev.finto.fi/yso/en/
2. Click on any concept in the alphabetical list, e.g. "aardvark" (this will perform a partial page load)
3. Change the UI language to e.g. Finnish by clicking on the language name

This should stay on the same concept page but display it in Finnish. Instead it will take you back to the vocabulary home page (but in Finnish).

A similar bug happens when you start from one concept page (e.g. "aardvark") and switch to another concept using the hierarchy sidebar. If you then change the UI language, you will be back at "aardvark", not the concept you had switched to.

The problem seems to be that the language change links are not updated during partial page loads.

Firefox

osma:

Fix a bug where changing the UI language after a partial page load could load the wrong page, see #1848.

This PR adds one more step to the partial page load: updating the topbar navigation links by copying them from the loaded concept page into the current page DOM. I also added a Cypress E2E test to verify that it works.

- Closes #1848

- add `updateTopbarNav` function to partial-page-load.js
- add Cypress test to verify

None

## Checklist

### DIFF
--- a/resource/js/partial-page-load.js
+++ b/resource/js/partial-page-load.js
@@ -23,6 +23,13 @@ const fetchWithAbort = (function () {
   }
 })()
 
+const updateTopbarNav = (conceptHTML) => {
+  // update topbar navigation with language links
+  const conceptTopbarNav = conceptHTML.querySelector('#topbar-nav')
+  const topbarNav = document.querySelector('#topbar-nav')
+  topbarNav.innerHTML = conceptTopbarNav.innerHTML
+}
+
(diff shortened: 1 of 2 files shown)


UnniKohonen (approved):
LGTM!

PR #1854, opened 2025-12-04
Search bar fixes

joelit:

Some unimplemented features were spotted in the global search bar. This addessess some of those.

- Related to #1489 

* Do not pass content language (as null) to the search if it's not selected. Instead, use the UI language. 
* Fix global search autocomplete result links

## Checklist

### DIFF
--- a/resource/js/global-search.js
+++ b/resource/js/global-search.js
@@ -148,11 +148,6 @@ function startGlobalSearchApp () {
         }
         return null
       },
-      updateSearchLang () {
-        const lang = this.getSearchLang()
-        if (lang) { this.changeLang(lang) }
-        return lang
-      },
       renderMatchingPart (searchTerm, label, lang = null) {
         if (label) {
           let langSpec = ''


resource/js/global-search.js
```diff
@@ -205,11 +207,12 @@ function startGlobalSearchApp () {
               result.hit = this.renderMatchingPart(renderedSearchTerm, result.prefLabel, result.lang)
             }
           }
+
           if ('uri' in result) { // create relative Skosmos page URL from the search result URI
-            result.pageUrl = result.uri
-            const urlParams = this.formatSearchUrlParams()
-            if (this.selectedLanguage !== window.SKOSMOS.lang) { // add content language parameter
-              urlParams.append('clang', this.selectedLanguage)
+            result.pageUrl = window.SKOSMOS.baseHref + result.vocab + '/' + window.SKOSMOS.lang + '/page/?'
```
osma, line 207: I think this should rather be done using the `getConceptUrl` method defined in `get-concept-url.js`. This is used by the sidebar components to generate concept page links. It's better to have this logic in just one place, and the function will also use a short form URL (/page/p12345) when possible instead of always using the `?uri=` form.

The same applies to the vocab-search component.

joelit: Good call - I made an issue (#1856 ) of the use of getConceptUrl in forming links for the concepts. I found out it leaves out the `/Skosmos/` installation folder when forming concept links and it doesn't handle the content language too well, i.e. `http://localhost/fi/page/?uri=http%3A%2F%2Fwww.yso.fi%2Fonto%2Fyso%2Fp19378&clang=` so it needs to be polished a bit, and tested accross all the different use cases, adding tests etc.

Then it should be used here too.

resource/js/global-search.js
```diff
@@ -50,7 +50,7 @@ function startGlobalSearchApp () {
     },
     mounted () {
       this.languages = window.SKOSMOS.languageOrder
-      this.selectedLanguage = this.updateSearchLang()
+      this.selectedLanguage = this.getSearchLang()
```
osma, line 53: This seems to have been the only use of updateSearchLang() so the method could now be deleted.

PR #1855, opened 2025-12-04
Add translations for search component UI texts

osma:

The search component (global & vocabulary) were missing some UI translations. This PR adds them. It also changes the text "Global search" to "Search across vocabularies", which is hopefully more user-friendly.

- part of #1514 
  - addresses requirement "2.5 The translation component provides the translated UI messages (cypress tests for this)" though no cypress tests...
- part of #1489 
  - addresses requirement "2.3 All UI text is in English (in the form of translation message IDs) (Message IDs TBD)" 
  - addresses requirement "2.5 The translation component provides the translated UI messages (cypress tests for this)"

* change "Global search" message to "Search across vocabularies" in base-template.twig (twice)
* define translations for all the search component messages in Lokalise & pull them into this PR
* Skosmos 2 was using the message `Search language: any` which got translated into `Any language` in English. I simplified this so that the base message is `Any language`.
* adjust Cypress tests for the global search bar to work with the new translations

Didn't add Cypress tests for the vocab search bar. We already have some of those (e.g, the global search bar) and UI translation is such a basic mechanism that it doesn't make sense to test every single use of it.

## Checklist

### DIFF
--- a/tests/cypress/template/global-search-bar.cy.js
+++ b/tests/cypress/template/global-search-bar.cy.js
@@ -9,7 +9,7 @@ describe('Global search bar', () => {
   })
 
   it('dropdown menu header text is updated according to the selected vocabularies', () => {
-    cy.get('#vocab-selector .vocab-dropdown-btn').should('contain.text', '1. Choose vocabulary')
+    cy.get('#vocab-selector .vocab-dropdown-btn').should('contain.text', '1. Valitse sanasto')
     // select "altlabel"
     cy.get('#vocab-list').contains('label', 'altlabel').find('input[type="checkbox"]').check({ force: true })
     cy.get('#vocab-selector .vocab-dropdown-btn').should('contain.text', 'altlabel')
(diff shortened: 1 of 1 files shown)


miguelvaara (approved):
Everything looks really good, both the tests and syntax checks pass as well.

Hopefully we can get the Sámi terms into the global search soon, just like they are already partially implemented in the vocabulary search (the hint text in the search field). That would improve equality between users of different languages and also make the UI more consistent.

It is also a bit unfortunate thatnow we lost translations for a few important languages because of the "Search language: any" change, but on the other hand, it encourages us to motivate the translator community to contribute more.

Of course these comments are not about the implementation itself but the changes you made work really well.

I think it si ready to merge.

osma:
> It is also a bit unfortunate thatnow we lost translations for a few important languages because of the "Search language: any" change, but on the other hand, it encourages us to motivate the translator community to contribute more.

I migrated the translations for this term to the new message ID in Lokalise, so they were not lost.

PR #1863, opened 2025-12-26
Support ordered rdf list for concept property

### LINKED ISSUE #1864: Support ordered rdf list for concept property

I am working on setting up a vocabulary that needs to allow for an ordered list of related concepts but the skosmos UI does only show the blank node of the first element in the list instead of showing the as it does with the unordered list.

All users of rdf:List properties

Define narrower concepts in an ordered manner.

It's pretty much done and ready to pull from the PR: https://github.com/NatLibFi/Skosmos/pull/1863

jdfranel:

I am working on setting up a vocabulary that needs to allow for an ordered list of related concepts but the skosmos UI does only show the blank node of the first element in the list instead of showing the as it does with the unordered list.

Closes #1864

To support this ordered list I did the following changes:

- Updated the ConceptPropertyValue model to parse and navigate through the ordered list (from rdf:first unti rdf:nil)
- Added a config property `skosmos:rdfListItemsLimit` to limit the maximum number of element in the list (default to 1000
- Updated the *view/concept-card.inc.twig* to handle the ordered list and also added some styles in the *skosmos.css*
- Added tests for this feature. I created a test vocabulary *test-rdf-list.ttl* file, updated the *testconfig.ttl*, added tests in the *ConceptPropertyValueTest.php*

Also, while testing my changes I had tu edit the tests so they use the env `SKOSMOS_SPARQL_ENDPOINT` instead of the static `'http://localhost:9030/skosmos/sparql'` in *tests/ModelTest.php* and *tests/VocabularyTest.php*

## Checklist

### DIFF
--- a/tests/ConceptPropertyValueTest.php
+++ b/tests/ConceptPropertyValueTest.php
@@ -253,8 +253,8 @@ class ConceptPropertyValueTest extends PHPUnit\Framework\TestCase
      */
     public function testRdfListUnordered()
     {
-        // Should have 12 individual values (not a list)
-        $values = $this->getAssertedConceptRdfListPropertyValues('http://www.skosmos.skos/test-rdf-list/languages-unordered', 12); // NOSONAR - RDF URI, not a web request
+        // Should have 17 individual values (not a list)
+        $values = $this->getAssertedConceptRdfListPropertyValues('http://www.skosmos.skos/test-rdf-list/languages-unordered', 17); // NOSONAR - RDF URI, not a web request
         
         // None of the values should be RDF lists
         foreach ($values as $value) {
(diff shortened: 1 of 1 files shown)


jdfranel:
Hi @osma ,
Thank you for reviewing my PR and for your feedback.

**Alignment with existing truncation features:**

I've updated the visual truncation script to handle RDF lists in the same way as other multi-valued properties. Lists are now shown with a "show all" link when they exceed the default display limit of 15 items, maintaining consistency with the existing UX patterns.

**Regarding rdfListItemsLimit:**

This configuration parameter serves as a backstop to prevent performance issues with excessively long or potentially malformed (looping) RDF lists. Since RDF list processing requires recursive SPARQL queries to traverse the list structure, very long lists could significantly impact server response time. The default value of 1000 should handle most legitimate use cases while protecting against more extreme cases. I considered using a default of 0 (unlimited), but opted for the conservative approach to avoid unexpected performance degradation on production systems. This is a different concern from the UI truncation - it's about query performance rather than user experience.

**Cypress tests:**

I've added Cypress tests in tests/cypress/template/rdf-list.cy.js that verify:

* Non-truncated RDF lists are displayed correctly in the HTML DOM
* Truncated RDF lists show the appropriate "show all" UI element
* The "show all" functionality works as expected for RDF lists

Let me know if there is other changes you want me to make.

PR #1905, opened 2026-01-20
Update favicon (new Skosmos 3 look)

osma:

I noticed that Skosmos 3.0 was still using the old favicon (boxy S-shaped icon) that does not match the Skosmos 3 style and colors. This PR updates the icon to the new design (by Hahmo).

Here are the old and new favicons side by side in Firefox:

image

As before, the default favicon may be overridden by adding a `custom-favicon.ico` file in the root directory that will then be used instead of the default.

n/a

Set new favicon

None

## Checklist


joelit (approved):
LGTM!

PR #1906, opened 2026-01-20
Fix unstable concept page tests (mapping headers and copy functionality)

### LINKED ISSUE #1904: Stabilize the unstable behavior of mappings tests related to the concept page
yso/fi/page/p[something]

## Description of the problem
Steps to produce the problem, and the expected outcome after it has been fixed

Across different test runs, the testing of mapping features related to the concept page sometimes reports errors, even though neither the test data nor its presentation actuallly changes.

Tests should be run both from the command line an in the Cypress web app using both full test set and individual tests and test output may need to be temporarily increased.

miguelvaara:
Inconsistently behaving tests make development harder and reduce confidence in the test results which can eventually impact the quality of the code and the application.

- Closes #1904 

Mostly added timeouts in appropriate places to ensure elements are fully loaded before being asserted and improved robustness and explicitness (e.g. .invoke())

## Checklist

### DIFF
--- a/tests/cypress/template/concept-full-vs-partial.cy.js
+++ b/tests/cypress/template/concept-full-vs-partial.cy.js
@@ -200,21 +200,13 @@ describe('Concept page, full vs. partial page loads', () => {
       // check the first mapping property name
       cy.get('.prop-mapping h2').eq(0).invoke('text').should('contain', 'Exactly matching concepts')
       
-      // Debug: Log the entire mappings section HTML to see what's actually rendered
-      cy.get('#concept-mappings').then(($mappings) => {
-        cy.log(`Mappings HTML: ${$mappings.html().substring(0, 500)}`)
-      })
-      
       // Check if mapping labels exist at all
       cy.get('.prop-mapping .prop-mapping-label', {'timeout': 20000}).should('exist').then(($labels) => {
-        cy.log(`Found ${$labels.length} mapping label(s)`)
         const text = $labels.first().text()
-        cy.log(`First mapping label text: "${text}"`)
         expect(text).to.contain('number sign')
       })
       
       // check the first mapping property values
-      cy.get('.prop-mapping').eq(0).find('.prop-mapping-label').eq(0).invoke('text').should('contain', 'number sign')
       cy.get('.prop-mapping').eq(0).find('.prop-mapping-label').eq(0).find('a').invoke('text').should('equal', 'number sign')
       cy.get('.prop-mapping').eq(0).find('.prop-mapping-label').eq(0).find('a').should('have.attr', 'href', 'http://www.wikidata.org/entity/Q175743')
       cy.get('.prop-mapping').eq(0).find('.prop-mapping-vocab').eq(0).contains('www.wikidata.org')


miguelvaara:
# Notes and clarifications for the reviewer

Mapping-related tests were failing rather often with AssertionError. 

Why? The object tested strings but a null was given (the issue was identified after temporarily adding logging). 

The cause: race conditions in partial page loads where DOM elements were accessed before fully loaded.

## Changes Made

### concept-full-vs-partial.cy.js

- Notation clipboard test
Lines 79-81
Ensures partial page load completes before copying notation. Previously copied wrong value (33.10 instead of 33.2).

- Mapping stabilization wait
Lines 150, 153-155
Prevents race condition where mappings reload after initial spinner check.

- Explicit mapping label wait
Lines 160-161
Ensures mapping labels exist before accessing their content.

- Changed .contains() to .invoke('text').should('contain')
Lines 158, 164, 166, 171 an 173
More explicit assertions, consistent pattern and hopefylly easier debugging.

- Hash namespace test
Lines 192-193 sand 199-209
The same pattern as labyrinths test - explicit label wait with 20 s timeout and explicit text checking.

**Summa summarum:**
- All 4 unstable mapping tests now pass consistently
- Notation clipboard test fixed (was expecting 33.2, got 33.10)
- Minimal changes, only essential stabilization added
- No functionality changes (only timing/reliability improvements)

### concept.cy.js

- Property override test - explicit text check and comment
Lines 139-141 (added line 140, modified 141)
The same more explicit text checking pattern. Added timeout lining up with other similiar parts. Improved commenting.

**Summa Summarum:**

Replaces .contains('text') with .invoke('text').should('contain', 'text') for explicit text checking and adds stabilization waits after spinner checks. Adds explicit waits for mapping labels (20 s timeout) and adds heading verification for partial page loads before accessing content.

- All 4 unstable mapping tests now pass consistently
- Notation clipboard test fixed (was expecting 33.2, got 33.10)
- Property override test made consistent with the same pattern
- Essential stabilization and consistency improvements

**Post Scriptum:**
.invoke('text') always asserts on the resolved final string, which makes it a better choice in tests where the expected output is not 100 percent certain. It is also less sensitive to DOM timing quirks and makes debugging easier.

tests/cypress/template/concept-full-vs-partial.cy.js
```diff
@@ -76,6 +76,9 @@ describe('Concept page, full vs. partial page loads', () => {
         cy.visit('/test-notation-sort/en/page/?uri=http%3A%2F%2Fwww.skosmos.skos%2Ftest%2Fta0114') // go to "Buri" concept page
         // click on the link to "Eel" to trigger partial page load
         cy.get('#tab-hierarchy').contains('a', 'Eel').click()
+        // Wait for partial page load to complete
+        cy.get('#concept-heading h1', {timeout: 10000}).should('contain', 'Eel')
+        cy.wait(500) // stabilization wait
```
osma: I think it's bad practice to wait for 500ms unconditionally. It will always slow down the test by 0.5s regardless of whether the wait was necessary or not. And even 500ms may not always be enough.

If there is no reason for this wait, it should be dropped. If it is needed (to ensure that the notation code element has already been updated), then it would be better to replace it with a cy.get that checks the notation code directly, just like the above rows checks the concept heading.

tests/cypress/template/concept-full-vs-partial.cy.js
```diff
@@ -144,22 +147,30 @@ describe('Concept page, full vs. partial page loads', () => {
       // check that we have some mappings
       cy.get('#concept-mappings').should('not.be.empty')
       // check that spinner does not exist after load
+      // NOTE: we need to increase the timeout as the mappings can take a long time to load
       cy.get('#concept-mappings i.fa-spinner', {'timeout': 15000}).should('not.exist')
+      
+      // Wait to ensure mappings don't start reloading after initial check
```
osma: I don't understand this. Why would mappings start reloading?
Again, waiting 1000ms unconditionally is bad practice and slows down the test suite.

osma (changes requested):
Thanks for the PR, it's good to get rid of the flaky tests.

I gave a couple of comments related to cy.wait() operations. I hope that we wouldn't have to use cy.wait() because it always slows down the tests and generally doesn't address the root cause.

I'm not a big fan of the "NOTE: we need to increase the timeout" comments, because it should be fairly obvious that when a large timeout is set, there is a reason for that and that reason is that something is slow. But if you think this makes the tests clearer, I'm OK with that.

tests/cypress/template/concept-full-vs-partial.cy.js
miguelvaara: Removed the second spinner check, assuming that the spinner will not reappear. The additional check may be unnecessary, although there has previously been some instability during loading, which was the original reason for the test.

tests/cypress/template/concept-full-vs-partial.cy.js
miguelvaara: Replaced the fixed delay with a wait for the actual appearance of the #copy-notation button.

miguelvaara: Fixed

tests/cypress/template/concept-full-vs-partial.cy.js
miguelvaara: Fixed

PR #1912, opened 2026-01-27
Resolve Wikidata URIs using QLever SPARQL endpoint instead of WDQS

### LINKED ISSUE #1911: Wikidata mapping queries are not working

e.g. https://test.dev.finto.fi/yso/en/page/p14629 (yso: aardvark)

## Description of the problem

The Wikidata label for the corresponding concept is not showing (it shows up as `wd:Q46212` when it should be "aardvark"):

Image

This also currently causes the Cypress test "contains mappings for hash namespace" to fail. This is an outstanding issue in PR #1904 which attempts to fix breaking or flaky Cypress tests.

I traced why this happens. Long story short: WDQS (the Wikidata SPARQL endpoint) doesn't like the SPARQL queries performed by EasyRdf. It responds with an error

> 403, Please respect our robots policy https://wikitech.wikimedia.org/wiki/Robot_policy (1e30f7b)

I understand that it is prudent to restrict traffic by bots, especially in these times of rampant AI bots. But WDQS is a SPARQL API service - it is intended for machine use. And Skosmos tries to follow the policy...

Anyway, by playing with `curl` I found out that WDQS currently doesn't like HTTP/1.1 requests (`--http1.1` option in curl) and will always(?) respond like that when queried over HTTP/1.1. It works fine with HTTP/2.0. But the built-in HTTP client in EasyRdf only supports HTTP/1.1. (Maybe the optional Laminas HTTP client would work better?)

However, WDQS is also slow and unreliable. Recently, QLever has started providing a [Wikidata SPARQL endpoint](https://qlever.dev/wikidata). We could consider switching to that. I'm going to prepare a PR on this soon.

Firefox

osma:

Wikidata URIs were not being resolved properly, see #1911. This PR works around the issue by switching to the alternative QLever Wikidata SPARQL endpoint which is much faster and (hopefully) more reliable than the Blazegraph-based official WDQS.

- Closes #1911

* switch to QLever SPARQL endpoint for resolving Wikidata entities
* rename some classes (WDQS -> Wikidata) to avoid confusion, as we are no longer using WDQS

I hope it really helped solve the problem?

## Checklist

### DIFF
--- a/src/model/resolver/Resolver.php
+++ b/src/model/resolver/Resolver.php
@@ -28,7 +28,7 @@ class Resolver
         if (preg_match('|http://id.loc.gov/[^/]+/[^/]+/.+|', $uri)) {
             $res = new LOCResource($this->model, $uri);
         } elseif ($this->startsWith('http://www.wikidata.org/entity/', $uri)) {
-            $res = new WDQSResource($this->model, $uri);
+            $res = new WikidataResource($this->model, $uri);
         } else {
             $res = new LinkedDataResource($this->model, $uri);
         }
--- a/src/model/resolver/WikidataResource.php
+++ b/src/model/resolver/WikidataResource.php
@@ -1,8 +1,9 @@
 <?php
 
-class WDQSResource extends RemoteResource
+class WikidataResource extends RemoteResource
 {
-    public const WDQS_ENDPOINT = "https://query.wikidata.org/sparql";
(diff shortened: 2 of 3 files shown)


osma:
WDQS started working with HTTP/1.1 again, so I closed issue #1911, but I'll leave this PR open in case we want to switch to QLever anyway.

PR #1914, opened 2026-01-29
Include version number and website URL in HTTP User-Agent header

osma:

Previously, Skosmos only used the User-Agent header `Skosmos` without any further information. This does not follow relevant recommendations and conventions, and may cause difficulties in analyzing and controlling web traffic.

This PR adds the version number and https://skosmos.org URL to the User-Agent header used in SPARQL, Linked Data, LOC and WDQS requests. A PHPUnit test verifies that it is generated properly.

The "Skosmos" user-agent header was implemented in #1691. Back then we decided not to include the version number:

> I decided not to include the Skosmos version number in the user-agent string based on discussion with other Skosmos developers.

- follow-up to #1691

- add Model.getUserAgent() method that generates the User-Agent header value
- use that method to construct user-agent value in GenericSparql, LinkedDataResolver, LOCResolver and WDQSResolver
- add PHPUnit test to verify that the header value is properly generated

I didn't include a test that would verify that the user-agent setting is actually used in HTTP requests. That would probably be possible with Cypress (by intercepting requests) but IMHO it's a lot of work for little benefit.

## Checklist

### DIFF
--- a/tests/ModelTest.php
+++ b/tests/ModelTest.php
@@ -56,7 +56,7 @@ class ModelTest extends PHPUnit\Framework\TestCase
         $userAgent = $this->model->getUserAgent();
         $version = $this->model->getVersion();
         $expected = "Skosmos/$version (https://skosmos.org/)";
-        $this->assertEquals($userAgent, $expected);
+        $this->assertEquals($expected, $userAgent);
     }
 
     /**


miguelvaara (approved):
Everything seems to be working well. Thanks @osma! I think you can go ahead and merge.

PR #1916, opened 2026-01-29
dockerfile improvements

jdfranel:

- The Docker image contains many files that are not needed to run Skosmos in production (mostly development files). This takes unnecessary space and might result in secret leakage.
- Node 18 reached end of life on April 30, 2025, creating a security risk.

N/A

This PR improves the Docker build process by reducing the production image size and updating outdated dependencies.

**Key changes:**

1. **Updated `.dockerignore`** to exclude development and unnecessary files from the Docker build context
2. **Modified Dockerfile to use selective file copying** instead of copying the entire workspace
3. **Added a development stage** in `Dockerfile.ubuntu` that includes files needed for testing (used by `tests/docker-compose.yml`)
4. **Upgraded Node.js from version 18 to 24** in both the Docker image and GitHub workflows

**Files excluded from production image:**
- Development configuration: `.codeclimate.yml`, `.jshintrc`, `phpunit.xml`, `.sonarcloud.properties`
- Documentation: `README.md`, `CONTRIBUTING.md`, `LICENSE`, `SECURITY.md`, `PULL_REQUEST_TEMPLATE.md`
- Development tools: `ansible/`, `tests/`, `.github/`, `dockerfiles/`, `composer.lock`, `composer.phar`, `phpdoc.sh`
- Configuration files: `docker-compose.yml`, `Vagrantfile`, `cypress.config.js`, `.gitignore`, `.dockerignore`, `ansible.cfg`

These changes reduce the attack surface and image size while maintaining full functionality in production.

Removed the `git` package from the Docker image since there is no `.git` directory in the production build. This results in a `sh: 1: git: not found` error appearing in logs on page requests. Options to resolve:
- Re-add the `git` package to the image, or
- Handle the error gracefully to prevent it from appearing in logs

Input needed on the preferred approach.

## Checklist


osma:
Thanks for the PR. There are a lot of good improvements here.

Regarding the `git` issue: I think this happens because some of the Controller code currently relies on git commands to determine if the code has changed: https://github.com/NatLibFi/Skosmos/blob/4d120d23a002df7684e8d16fc82237fd6ef9c34d/src/controller/Controller.php#L150-L185

As long as this is the case, the git command and the `.git` tree are needed - or at least you will see error messages if you remove them.

IIRC, this check for the latest git commit is a mechanism to invalidate caches when the code changes for example via `git pull`. It is not very useful within Docker containers though, because the code will presumably never change within a container.

jdfranel:
That's what I thought. I guess that the proper way to handle this issue is to prevent the `sh: 1: git: not found` to appear in the logs.

I also noticed it happens in the phpunit tests but I would argue that it might be better to add git package to the development stage so the tests passes whether it is in a docker container or not.

If it is OK for you I will update this PR so it is properly handled.

osma:
Sounds good, please do make the changes in this PR!

.dockerignore
```diff
@@ -12,3 +12,13 @@ node_modules
 .DS_Store
 .phpunit.result.cache
 .git
+ansible
```
osma: FYI, the `ansible` directory was just removed in #1920 so this line is no longer necessary (although it does no harm)

jdfranel: I did removed the ansible directory from the .dockerignore
Is the ansible.cfg also going to be removed ?

osma: Good point. I removed ansible.cfg, along with a few other files, in PR #1923.
The project has a long history and some cruft has accumulated that wasn't actually being used. Thank you for pointing these out!

jdfranel:
git shell calls have been changed to redirect the `stderr` to `/dev/null`. I also had to adapt some tests to assert that the missing git is returning null is null so phpunit does not raise a risky test that doe not assert anything. This way, if git is installed on the host, the test swill run as it was previously and if the tests runs on the docker image where there is no git it will pass also.

There is still some SonarCloud Code Analysis that raises an error when I copy all sources recursively in the composer stage. This can be marked as safe since the final stage copy specific files only.

osma:
Sorry this didn't quite make it into the 3.1 release. There is quite a lot to test and digest here. But we will get back to it soon.

jdfranel:
Great, thank you.

dockerfiles/Dockerfile.ubuntu
osma, line 1: I'm a little worried about these substantial changes to the Dockerfile. On the one hand, this multi-stage setup does produce a cleaner final image, which is a good improvement. On the other hand, it seems to introduce new dependencies in the form of other docker images (`node:*-alpine`, `composer:latest`) and if these change, then this Dockerfile may need to be changed as well. It is also quite a bit longer and more complex than the previous one. Is this a headache to maintain?

I wonder if this could be simplified at least a little, for example having one stage for installing the dependencies (both npm and composer) and another for the final image? And could these both be based on the same Ubuntu base image so that we don't need two separate mechanisms for installing Node.js?

Also dealing with dev and production variants separately leads to some internal duplication. Could this be instead controlled with an ARG?

jdfranel: I agree the duplication between prod/dev stages adds complexity. However, I believe multi-stage builds with specialized images (`composer`, `node:*-alpine`) are Docker best practices and actually simplify maintenance compared to installing from source in Ubuntu. I use this as much as possible in all my projects.

It allows for faster image building since the composer and node images are already ready to use and cached. Also, we can choose the version we want (here I already locked node version to 24 in argument `NODE_VERSION` and we could do the same with composer if we feel like it.

The Dockerfile is indeed longer for a pure production image but when it comes to the development flavor of the image it allows for reduced code duplication, better consistency and easier maintenance of various flavors of image compared to having a full separated development image.

I could add a `ARG BUILD_ENV=production` and unify the composer and node stages like this:

```Dockerfile
FROM composer:latest AS composer-stage
ARG BUILD_ENV
WORKDIR /app
COPY composer.json composer.lock* /app/
RUN if [ "$BUILD_ENV" = "production" ]; then \
      composer install --no-dev --no-scripts --no-autoloader; \
    else \
      composer install --no-scripts --no-autoloader; \
    fi
COPY . /app/
RUN composer dump-autoload

# Stage 2: Install Node dependencies (controlled by BUILD_ENV)
ARG NODE_VERSION
FROM node:${NODE_VERSION}-alpine AS npm-stage
ARG BUILD_ENV
WORKDIR /usr/src/app
COPY package.json package-lock.json* ./
RUN if [ "$BUILD_ENV" = "production" ]; then \
      npm install --omit=dev --ignore-scripts; \
    else \
      npm install --ignore-scripts; \
    fi
```

Which would reduce the number of stages but not so much the code duplication (we still have a `composer install` and `npm install` command for each flavor). Plus this would force the docker-compose.yml for development to add a build argument like

```yaml
  skosmos:
    build:
      context: ..
      dockerfile: dockerfiles/Dockerfile.ubuntu
      target: development
      args:
        BUILD_ENV: development
```

which makes the image less "ready to use".

osma: We discussed this among Skosmos developers. We agree that it makes sense to use specialized images. So let's keep it like that after all.

We would still prefer the `ARG BUILD_ENV=production` approach, which reduces internal duplication within the Dockerfile.

As an alternative to the `if` / `else` / `fi` structures that you used in your example, it would be possible to select some flags with an inline expression like this, which is a little more compact:

```
# Add --no-dev only in production
RUN composer install \
    --no-scripts \
    --no-autoloader \
    $( [ "$BUILD_ENV" = "production" ] && echo --no-dev )  # trick here
```

Can you please modify the Dockerfile along these lines?

Thank you for your effort and persistence, this is very close to merge-ready now.

jdfranel: It is done. I added "the trick" so there is 5 stages left. While I was at it, I pinned the composer version to 2.9 to please SonarCloud and cleaned some stuff up.

osma (approved):
Excellent, thank you for your work!

PR #1917, opened 2026-01-29
Load new search results on vocab and global search pages

### LINKED ISSUE #1532: Use e.g. Waypoints to partially render the search result lists

There is probably no point in rendering all the search results when we have a lot more of them that can fit in a browser screen. Implementing a partial load would be a good idea - previously in Skosmos 2.x this was done with Waypoints.

Each rendering of a search result hit triggers calls for resolvling uri resources, translation strings, etc. At some point rendering all of the results at once will result in long page loading times. As the search strings contain an implicit wildcard character in the end, it is possible to get hundreds of hits for a short search term. These would be improved by partial rendering of the list.

## Checklist

- [x] more search loaded after first 20
- [x] special characters in the query are working (see comment below)

UnniKohonen:

No new results are loaded when scrolling on vocab and global search pages. This PR adds the functionality.

- Closes #1532

- Add `load-search-results.js` that loads new search results when the search result list is scrolled down
- Add new `search_count` and `search_results_size` keys to `SKOSMOS` object
- Add cypress tests for vocab and global search pages
- Add CSS styles for search loading and counts messages

## Checklist

### DIFF
--- a/resource/js/load-search-results.js
+++ b/resource/js/load-search-results.js
@@ -13,7 +13,7 @@ function handleScrollEvent () {
 
     // Add a spinner to end of result list
     const spinner = document.createElement('p')
-    spinner.id = 'search-loading-spinner',
+    spinner.id = 'search-loading-spinner'
     spinner.innerHTML = `${$t('Loading more items')} <i class="fa-solid fa-spinner fa-spin-pulse"></i>`
     searchResultList.append(spinner)
 
@@ -48,7 +48,7 @@ function handleScrollEvent () {
         // If all results have been loaded, display message
         if (searchResultOffset >= window.SKOSMOS.search_count) {
           const message = document.createElement('p')
-          message.id = 'search-count',
+          message.id = 'search-count'
           message.textContent = $t('All %d results displayed').replace('%d', window.SKOSMOS.search_count)
           searchResultList.append(message)
         }


resource/js/load-search-results.js
```diff
@@ -0,0 +1,62 @@
...
+    // Add a spinner to end of result list
+    const lastResult = document.querySelector('.search-result:last-of-type')
+    lastResult.innerHTML += `<p id="search-loading-spinner">${$t('Loading more items')} <i class="fa-solid fa-spinner fa-spin-pulse"></i></p>`
+
+    // Construct search URL depending on page type
+    const params = new URLSearchParams(window.location.search)
+    const searchURL =
+      window.SKOSMOS.pageType === 'vocab-search'
+        ? `${window.SKOSMOS.vocab}/${window.SKOSMOS.lang}/search?clang=${window.SKOSMOS.content_lang}&q=${params.get('q')}&offset=${searchResultOffset}`
+        : `${window.SKOSMOS.lang}/search?clang=${window.SKOSMOS.content_lang}&q=${params.get('q')}&vocabs=${params.get('vocabs')}&offset=${searchResultOffset}`
```
joelit: Maybe instead use the URLSearchParams object to handle the search parameters
```suggestion
    params.set('offset', searchResultOffset)
    const searchURL =
      window.SKOSMOS.pageType === 'vocab-search'
        ? `${window.SKOSMOS.vocab}/${window.SKOSMOS.lang}/search?${params.toString()}`
        : `${window.SKOSMOS.lang}/search?${params.toString()}`
```

resource/js/load-search-results.js
```diff
@@ -0,0 +1,62 @@
...
+
+        // Re-enable event listener after results have been loaded
+        document.addEventListener('scroll', handleScrollEvent)
+
+        // If all results have been loaded, display message
+        if (searchResultOffset >= window.SKOSMOS.search_count) {
+          const newLastResult = document.querySelector('.search-result:last-of-type')
+
+          const translatedMessage = $t('All %d results displayed').replace('%d', window.SKOSMOS.search_count)
+          newLastResult.innerHTML += `<p id="search-count">${translatedMessage}</p>`
```
joelit: Looks a bit hacky, but when I tried writing something nicer by creating a p-element and appending it to the search-results div, it didn't really come up that much better.

PR #1918, opened 2026-01-29
Concept page: special display for deprecated concepts

### LINKED ISSUE #1877: Show deprecated concepts on the concept page

osma:

This implements special display for deprecated concepts: there is a warning box indicating that the concept has been deprecated (with links to replacements if available), and the text within the concept card is shown in a greyish, muted color.

Here is how it looks:

image

- Closes #1877

* changes to the concept-card template: warning box for deprecated concepts + add some CSS classes for styling
* main CSS file: define new color variables (`--muted-color` and `--main-content-disabled`), matching existing variable naming practice
* Concept class: add getIsReplacedBy method so that the Twig template can directly query for replacements (in Skosmos 2 this was done using client-side javascript but I think this is better)
* PHPUnit tests for getDeprecated (this was missing) and getIsReplacedBy

I had to adjust the visual design because I found the original design problematic in many ways. Here are things I did:
* The warning text is not bold, unlike in the design spec; only the replacement concept labels are set in bold. I think this is better in line with the rest of the UI and makes the replacement concepts stand out better.
* The grey color used in the design spec was too light so it didn't follow accessibility requirements. I made it darker, so that it has a contrast rating slightly above 4.5:1.
* I also used the grey color for other text elements on the concept cards, not just the headline as in the visual spec. I think this makes the deprecated status more obvious and better matches the display of deprecated concepts in Skosmos 2.

## Checklist

### DIFF
--- a/src/view/concept-card.inc.twig
+++ b/src/view/concept-card.inc.twig
@@ -2,7 +2,7 @@
   <div id="main-content-top-slot"></div>
   <div id="main-content">
       {% if concept.deprecated %}
-      <div class="main-content-section py-5 alert alert-danger" id="concept-deprecated-alert">
+      <div class="main-content-section py-5 alert" id="concept-deprecated-alert">
         <i class="fa-solid fa-triangle-exclamation"></i>
         {{ "This concept has been deprecated." | trans }}
         {% set replacedByValues = concept.isReplacedBy %}


joelit:
I am able to reproduce a consistent bug where the warning box is appended after the concept info during the first full page load. On refresh it is moved back to the correct position.
image

joelit:
On further notice, I would consider dropping the bread crumb line, as it does not really offer anything useful here. Then again, this is only true in cases (like at NLF) where the deprecated concept is removed from the hierarchy. As this is not a requirement in the SKOS specification, it's probably safer to display the bread crumb section in case it makes sense for some other controlled vocabularies. Just a thought.

osma:
@joelit 

> I am able to reproduce a consistent bug where the warning box is appended after the concept info during the first full page load. On refresh it is moved back to the correct position.

Good catch! Fixed in [a91b494](https://github.com/NatLibFi/Skosmos/pull/1918/commits/a91b494b5e5eb257fa10a2dac9c13a586ca22fbe)

> On further notice, I would consider dropping the bread crumb line, as it does not really offer anything useful here. Then again, this is only true in cases (like at NLF) where the deprecated concept is removed from the hierarchy. As this is not a requirement in the SKOS specification, it's probably safer to display the bread crumb section in case it makes sense for some other controlled vocabularies. Just a thought.

Well, Skosmos 2 displays the breadcrumb and it's also shown in the Skosmos 3 visual design for deprecated concepts. So I think I'm leaving it in, even though it's a bit nonsensical for NLF vocabularies.

Added translations (had to adjust message keys, which led to some churn) and a Cypress test. I think this is now ready.

src/view/concept-card.inc.twig
```diff
@@ -1,9 +1,23 @@
 <div class="col-md-8">
   <div id="main-content-top-slot"></div>
   <div id="main-content">
+      {% if concept.deprecated %}
+      <div class="main-content-section py-5 alert alert-danger" id="concept-deprecated-alert">
```
UnniKohonen: The `.alert-danger` class adds a pink border around the element. I think the class could be removed completely.

UnniKohonen (approved):
I would remove the `.alert-danger` bootstrap class but otherwise looks good!

PR #1920, opened 2026-02-02
Remove obsolete ansible directory

osma:

The top level directory `ansible` was created 7 years ago. We have not actually used Ansible in Skosmos development, and the configuration files are no longer valid for current Skosmos. This PR removes the directory along with its contents. We can always revive it if it turns out to be useful later on.

- n/a

* remove `ansible` directory and all its contents

None

## Checklist


osma:
Waiting for tests to pass, then I will merge this.

PR #1935, opened 2026-02-13
Fix autocomplete hiding cross-language matches in "Any language" mode

### LINKED ISSUE #1934: Autocomplete: show all language matches when searching 'Any language'
## Problem

When using the autocomplete search with "Any language" selected, the `unique=true` parameter is always passed to the REST API. This causes deduplication by URI, keeping only one language match per concept.

For example, searching `restaur*` with "Any language" in a trilingual vocabulary (EN/FR/ES) returns only 5 Spanish results, hiding 4 French matches for the same concepts:

| With `unique=true` (current) | With `unique=false` (proposed) |
|---|---|
| 5 results (ES only) | 9 results (ES + FR) |

The French matches (`Restaurer 30 % des écosystèmes dégradés`, `Restaurer, préserver et renforcer...`, etc.) are silently dropped because their URIs already appeared in the Spanish results.

## Expected behavior

When "Any language" is selected, deduplication should be disabled so all language matches are shown. This is the whole point of "Any language" — to find matches across languages.

When a specific language is selected (e.g. French), `unique=true` still makes sense since there can't be cross-language duplicates.

## Proposed fix

In `resource/js/vocab-search.js`, make `unique` conditional on the selected language:

```javascript
// Current (line 63):
const skosmosSearchUrlParams = new URLSearchParams({ query: this.formatSearchTerm(), unique: true })

// Proposed:
const unique = this.selectedLanguage !== 'all'
const skosmosSearchUrlParams = new URLSearchParams({ query: this.formatSearchTerm(), unique: unique })
```

The `selectedLanguage` property is already available at this point in the code (set during `mounted()`), so no additional plumbing is needed.

## Context

Discovered while building a multilingual SKOS vocabulary (EN/FR/ES) for the [Kunming-Montreal Global Biodiversity Framework](https://www.cbd.int/gbf).

fvogel:
## Summary
- Autocomplete search always passes `unique=true` to the REST API, which deduplicates results by URI
- When "Any language" is selected, this silently drops matches from other languages (e.g. French results hidden by Spanish ones for the same concept)
- Fix: set `unique=false` when "Any language" is selected, `unique=true` otherwise

## Steps to reproduce
1. Set up a multilingual vocabulary (e.g. EN/FR/ES)
2. Select "Any language" in the search dropdown
3. Type a term that exists in multiple languages (e.g. `restaur` matching both Spanish "Restaurar..." and French "Restaurer...")
4. Only results from one language appear; the other language matches are silently dropped

## Fix
One-line change in `resource/js/vocab-search.js`: make `unique` conditional on whether a specific language is selected.

```javascript
// Before:
const skosmosSearchUrlParams = new URLSearchParams({ query: this.formatSearchTerm(), unique: true })

// After:
const unique = this.selectedLanguage !== 'all'
const skosmosSearchUrlParams = new URLSearchParams({ query: this.formatSearchTerm(), unique: unique })
```

## Before / After

**Before (v3.1):** Only Spanish (es) matches — French results silently dropped.

before

---

**After (this fix):** Both Spanish and French matches shown, as expected for "Any language" search.

after

Fixes #1934

### DIFF
--- a/resource/js/vocab-search.js
+++ b/resource/js/vocab-search.js
@@ -60,7 +60,8 @@ function startVocabSearchApp () {
         const mySearchCounter = this.searchCounter + 1 // make sure we can identify this search later in case of several ongoing searches
         this.searchCounter = mySearchCounter
         let skosmosSearchUrl = 'rest/v1/' + window.SKOSMOS.vocab + '/search?'
-        const skosmosSearchUrlParams = new URLSearchParams({ query: this.formatSearchTerm(), unique: true })
+        const unique = this.selectedLanguage !== 'all'
+        const skosmosSearchUrlParams = new URLSearchParams({ query: this.formatSearchTerm(), unique: unique })
         if (this.selectedLanguage !== 'all') skosmosSearchUrlParams.set('lang', this.selectedLanguage)
         skosmosSearchUrl += skosmosSearchUrlParams.toString()
 


osma:
Thank you for identifying the issue and proposing a fix.

There is similar functionality for the global search bar and I think it needs to be changed as well: https://github.com/NatLibFi/Skosmos/blob/3ea3d857d81b200f40b366e503d986ff51baf8bb/resource/js/global-search.js#L121

To verify that the fix works today and stays working in the future, we would like to have Cypress tests that check for this case. Would it be possible for you to provide such tests? The tests could go into tests/cypress/template/vocab-search-bar.js and tests/cypress/template/global-search-bar.js, respectively.

You probably need a test vocabulary with a suitable test case. One of the test vocabularies is a small subset of YSO that contains the `yso:p3358` concept which has the preflabels `"systems"@en` and `"system"@sv`, maybe you could use that?

PR #1939, opened 2026-02-28
Add timezone support and improve date formatting

jdfranel:

- I needed to display time along with date on concept date infos;
- Date infos in concepts and vocabularies are displayed in en_US even though the `skosmos:languages` is set to en_GB for label `en`;
- On docker image, php timezone is `UTC`;
- Timezone in dates (if defined) is not handled in date formatting.

- Closes #

This PR introduces configurable timezone support and improves date/time formatting throughout Skosmos:

1. **Timezone Support**
   - Added new `DateTimeHelper` utility class (`src/util/DateTimeHelper.php`) that handles timezone conversions
   - Configuration option `skosmos:timezone` added to `config.ttl` (e.g., `"Europe/Paris"`)
   - Alternatively, timezone can be set via `SKOSMOS_TIMEZONE` environment variable
   - Defaults to `date_default_timezone_get()` if not configured
   - All dates are now converted to the configured timezone before display

2. **Locale mapping**
   - Improved locale handling with proper mapping from language codes to Punic-compatible locales

3. **Smart Time Display in Concept Date infos**
   - Time is only shown when it's not midnight (00:00:00) in the original date (before conversion to timezone)
   - Date-only values (midnight timestamps) display only the date
   - Date+time values show both date and time in localized format

4. **Concept Date infos Layout**
   - Created and modified dates are now displayed on separate lines for better readability
   - Updated `concept-card.inc.twig` to use `nl2br` filter for multiline date display

5. **Updated tests**
   - Added unit tests in `tests/ConceptTest.php`:
     - Test date display without time component
     - Test date display with time component
     - Test timezone conversion (various timezone offsets)
     - Test interface language formatting (English vs French)
   - Updated existing unit tests in `tests/VocabularyTest.php`:
     - Updated expected date formats in vocabulary info tests
     - Added timezone conversion tests for vocabulary dates
     - Added interface language tests for vocabulary dates
   - Added test data in `tests/test-vocab-data/date.ttl`:
     - Date without time component (midnight)
     - Date with time component (non-midnight)
     - Date with timezone offset for conversion testing
   - Updated Cypress tests in `tests/cypress/template/concept.cy.js`:
     - Updated expected date formats for English and Finnish
     - Added tests to verify date format remains in interface language when switching content language
   - Updated test configuration in `tests/testconfig.ttl` to explicitly set UTC timezone for consistent test results

- The timezone conversion applies globally to all vocabs in a Skosmos instance. There's no per-vocabulary timezone configuration (this seems acceptable for most use cases)
- The PR changes some existing date display formats, which might require documentation updates for users who have customized templates
- Invalid timezone strings fall back to `date_default_timezone_get()` with a warning - this should be sufficient but could be made more explicit in documentation

## Checklist


jdfranel:
There is the same SonarCloud issue with the namespace http. I'll let you fix this form inside SonarCloud this time unless you want me to disable the warning from the code.

src/model/GlobalConfig.php
```diff
@@ -238,6 +238,23 @@ public function getDefaultEndpoint()
         }
     }
 
+    /**
+     * Returns the timezone for date formatting defined in the configuration.
+     * If not set, checks SKOSMOS_TIMEZONE environment variable, defaulting to 'UTC'.
+     * @return string timezone identifier (e.g., 'Europe/Paris', 'UTC')
+     */
+    public function getTimezone()
```
osma, line 306: I wonder if we need so many mechanisms for setting the timezone? PHP already has a configuration option [date.timezone](https://www.php.net/manual/en/datetime.configuration.php#ini.date.timezone) and the [date_default_timezone_get](https://www.php.net/manual/en/function.date-default-timezone-get.php) function (used below) returns its configured value. I understand that this PHP setting can be a bit difficult to adjust in Docker images, so the environment variable is maybe warranted, but I don't really see a use case for a new Skosmos configuration setting, we already have too many of those :)

src/util/DateTimeHelper.php
```diff
@@ -0,0 +1,169 @@
...
+        return $this->timezone;
+    }
+
+    /**
+     * Convert a language code to a Punic-compatible locale
+     *
+     * @param string $lang Language code (e.g., 'en', 'fr', 'de')
+     * @return string Locale string (e.g., 'en_GB', 'fr_FR') with encoding stripped
+     */
+    private function languageToLocale(string $lang): string
```
osma: Not sure if this kind of logic is needed? GlobalConfig::getLanguages returns an array where the keys are BCP47 style language tags and the values are locales. Using the keys instead of values should be easier?

src/util/DateTimeHelper.php
```diff
@@ -0,0 +1,169 @@
...
+        return \Punic\Calendar::formatDatetime($converted, $format, $locale);
+    }
+
+    /**
+     * Check if a DateTime has a time component (not midnight)
+     *
+     * @param DateTime $dateTime The datetime to check
+     * @return bool True if time is not 00:00:00
+     */
+    public function hasTimeComponent(DateTime $dateTime): bool
```
osma: What about actual midnight times? This would strip them.

Can't we instead assume that if the timestamp value in the RDF data is of type `xsd:dateTime` then the time component is important and needs to be shown. If it is `xsd:date`, then there is no time component to display.

osma (commented):
Thank you for the PR, this is an important feature.

I added some specific comments inline. In general I'm a bit worried that this PR introduces quite a lot of new code for what should be a rather simple new feature. Is all of this abstraction (e.g. the DateTimeHelper utility class) really needed or could it be simplified? All the extra code needs to be maintained in the future. You have added PHPUnit tests and Cypress tests, which is obviously good for future maintenance! 👍 

Also, not sure I agree with the layout changes (moving the timestamps to separate lines). It looks like this:

Image

This is of course partly a matter of taste, but we are working with a visual design where these were deliberately on the same line and this PR would change that. Perhaps this could be implemented in a way that allows these timestamps to be styled so that CSS rules could switch between inline format and separate lines?

src/model/GlobalConfig.php
sqelirus: The docker image as it is (and as it could be if/when the PR #1916 is merged) is not supporting custom timezone. I admit I went with this the approach of adding a new skosmos setting to have a quick way to setup the timezone for the application only and not having to thinker and be able to have a docker image to run and be all set from the start.
That being said, getting the timezone from `date_default_timezone_get()` is possible but would imply we add a entrypoint script to the docker image that would set the timezone from the `TZ` environment variable at runtime before the command `apache2ctl -D FOREGROUND` is executed. This would also be more future proof if we need other system configuration at runtime.
If this is fine by you, I can remove the `SKOSMOS_TIMEZONE` and `getTimezone()` logic from this PR and use `date_default_timezone_get()` move this logic to a different PR once the #1916 is merged.
This will allow for some code to be removed but I am not sure how much at that point since I don't think `Punic\Calendar` is going to convert dates in the local timezone when formating. I do think that it is preferable to have all dates displayed in the local timezone for consistency and better user experience.

jdfranel: The docker image as it is (and as it could be if/when the PR #1916 is merged) is not supporting custom timezone. I admit I went with this the approach of adding a new skosmos setting to have a quick way to setup the timezone for the application only and not having to thinker and be able to have a docker image to run and be all set from the start.
That being said, getting the timezone from `date_default_timezone_get()` is possible but would imply we add a entrypoint script to the docker image that would set the timezone from the `TZ` environment variable at runtime before the command `apache2ctl -D FOREGROUND` is executed. This would also be more future proof if we need other system configuration at runtime.
If this is fine by you, I can remove the `SKOSMOS_TIMEZONE` and `getTimezone()` logic from this PR and use `date_default_timezone_get()` move this logic to a different PR once the #1916 is merged.
This will allow for some code to be removed but I am not sure how much at that point since I don't think `Punic\Calendar` is going to convert dates in the local timezone when formating. I do think that it is preferable to have all dates displayed in the local timezone for consistency and better user experience.

src/util/DateTimeHelper.php
jdfranel: I didn't consider this option. Thank you for pointing that out. It looks like a cleaner and lighter implementation.

src/util/DateTimeHelper.php
jdfranel: This is not a easy topic. The langcode, or `GlobalConfig::getLanguages` label ("en", "de",...), is actually widely used and most of the time is used to select which Literal to use or which Symfony translation to use. The `GlobalConfig::getLanguages` value is, afaik, only used in the negociator in the `WebContoller` class.

Since I wanted to format the date in the current UI language and that it is set as the langcode I had to map the langcode to the corresponding `GlobalConfig::getLanguages` value to have the desired locale for this interface language. This was a deliberate choice to minimize the changed needed to have this feature implemented.

There is a couple of different standard used and I have not been able to find a reason in the skosmos wiki or in the code to explain what it is the case. There is the "rdfs:label" (ISO 639) `en`, the server "rdf:value" (system locale) `en_GB.utf8` and the BCP47 `en-GB` used in the rdf language-tagged literals.

IMO, the `skosmos:languages` does not need an `rdfs:label` and should only use a list of BCF47 locales and infer langcode when needed and probably force utf-8 on all locale. So a single `en-GB` would be converted in `en` or `en_GB.utf8` depending on what's needed.

The BCF47:
* should be OK to feed to the Symfony Translator
* would format the dates correctly when used in Punic\Calendar
* make useless the `DateTimeHelper->languageToLocale()` function

jdfranel:
> Also, not sure I agree with the layout changes (moving the timestamps to separate lines). It looks like this:
> 
> Image
> This is of course partly a matter of taste, but we are working with a visual design where these were deliberately on the same line and this PR would change that. Perhaps this could be implemented in a way that allows these timestamps to be styled so that CSS rules could switch between inline format and separate lines?

I totally see your point and was not aware of the design guidelines.  I went with the line split because :
1. It was a maximum of 2 lines (created and modified) so the vertical impact was limited.
2. Since I added the time in the date format some languages add a comma between the date and time in the Punic/Calendar "short" format and it was too much comma for my taste. This was not an issue when there was no time displayed. 

Screenshot from 2026-03-04 00-01-41
Screenshot from 2026-03-04 00-02-23

This is a simple rollback to single line if you rather have it on a single line. Let me know how you want it.

src/util/DateTimeHelper.php
osma: Good points. The configuration of specific locales (using rdf:value) is a remnant of how Skosmos 2 handled translations using the PHP gettext() function, which relies on system locale. Nowadays Skosmos 3 uses Symfony Translations so I think the information about locales is no longer needed at all.

But my point was that if you used something like

```php
$languages = $globalConfig->getLanguages();
$bcp47_lang_codes = array_keys($languages);
```

then you should have the BCP47 language codes you were looking for and `languageToLocale()` should not be needed anymore. Or is there something specific that Punic requires that this does not cover?

osma:
> This is a simple rollback to single line if you rather have it on a single line. Let me know how you want it.

I think it would make things easier to keep the timestamps on a single line. We can consider splitting it to two lines separately (I can discuss this with co-developers) but the display aspect is IMHO a bit separate from the core of this PR.

src/model/GlobalConfig.php
osma: How about leaving in the support for SKOSMOS_TIMEZONE but dropping `skosmos:timezone`? Then people using Docker would still have a relatively easy way of overriding the timezone set in PHP configuration.

sqelirus: Sounds like a good middle ground. I'll get to it.

src/util/DateTimeHelper.php
jdfranel:
```php
$languages = $globalConfig->getLanguages();
$bcp47_lang_codes = array_keys($languages);
// returns { [0]=> string(2) "de" [1]=> string(2) "en" [2]=> string(2) "fr" [3]=> string(2) "it" }
```

Return the list of language code and this is what is not working correctly with punic because when passing `en` as locale, Punic will format date as an en-US date. So for a `skosmos:languages ( [ rdfs:label "en" ; rdf:value "en_GB.utf8" ] ) ;` 

We have
```
"en" => "Last modified 11/30/2024, 11:06 AM"
"en_GB.utf8" => "Last modified 11/30/2024, 11:06 AM"
"en_GB" => "Last modified 30/11/2024, 11:06" (correct) 
"en-GB" => "Last modified 30/11/2024, 11:06" (correct)
```

This is why there is a `languageToLocale` function to infer the working locale from the langcode. If we use the `rdf:value` matching the `rdfs:label` would not be sufficient.

osma: Thanks, I see your point now. I'll let you know if I come up with an alternative approach, but let's stick with this for now.

osma:
> There is the same SonarCloud issue with the namespace http. I'll let you fix this form inside SonarCloud this time unless you want me to disable the warning from the code.

FYI, I tried to silence this warning in SonarCloud settings but I'm not yet sure if that worked. If you push a new commit we will find out. (It looks like SonarCloud doesn't have a way of rerunning the scan for an unchanged PR, or at least I couldn't figure it out.)

jdfranel:
> > There is the same SonarCloud issue with the namespace http. I'll let you fix this form inside SonarCloud this time unless you want me to disable the warning from the code.
> 
> FYI, I tried to silence this warning in SonarCloud settings but I'm not yet sure if that worked. If you push a new commit we will find out. (It looks like SonarCloud doesn't have a way of rerunning the scan for an unchanged PR, or at least I couldn't figure it out.)

Ok, I'll check if the tests pass the Quality Gate. If not I can add the same `// NOSONAR - RDF URI, not a web request` comment to specifically silence it.

osma:
Please don't add the NOSONAR tags, this PR is a good test case as it is without those. You can ignore the SC warnings.

src/model/Concept.php
```diff
@@ -741,45 +741,58 @@ public function getModifiedDate()
         return $this->getVocab()->getModifiedDate();
     }
 
+    /**
+     * @param string $name Resource property
+     * @return string|null The resource date formated as string or null
+     */
+    private function getFormatedDateResource($name)
```
osma: Typo nitpick: this should be `getFormattedDateResource` (two t's)

osma:
It looks like I managed to silence the SonarCloud complaint about HTTP URIs in tests. :tada: (My first attempt yesterday didn't work.)

SonarCloud still complains about duplicated code in unit tests. That is also not a very relevant complaint. I tried to configure it to ignore duplicated code in tests from now on.

tests/testconfig.ttl
```diff
@@ -31,6 +31,8 @@
     # HTTP client configuration
     skosmos:sparqlTimeout 10 ;
     skosmos:httpTimeout 20 ;
+    # timezone for date formatting (defaults to UTC)
```
osma: I think this can be dropped now (support for setting timezone in Skosmos config was removed)

src/util/DateTimeHelper.php
```diff
@@ -0,0 +1,121 @@
...
+     *
+     * @return DateTimeZone
+     */
+    public function getTimezone(): DateTimeZone
+    {
+        return $this->timezone;
+    }
+
+    /**
+     * Convert a language code to a Punic-compatible locale
```
osma: I still have reservations about what this method does. In my view, OS locale names are not necessarily easy to convert in this way. This code does not seem to handle e.g. locales with three-letter language codes (`fil_PH`, `ast_ES`, `haw_US`) or locales which specify the writing system (`sr_Latn__RS`, `zh_Hans_CN`). There are also exceptional locales like `en_001` [English (World)]. Locale names on Windows are a bit different too - e.g. they use BCP47 style `-` instead of `_` as separator. But I just think that it's fundamentally wrong to rely on the locale name/code itself for anything - it's just a name and those can be arbitrary.

How about instead of using Punic to do the date formatting, just rely on the PHP intl extension, which we already depend on in other ways? The [IntlDateFormatter](https://www.php.net/manual/en/intldateformatter.format.php) class seems to do what is needed. Then we could avoid having to convert locales. Maybe there would be no need for this DateTimeHelper class either?

We still need Punic as a dependency for other reasons (mainly language names in different languages) but that doesn't mean we should use it for everything.

osma (changes requested):
Hi, I looked through this once more and made some suggestions, mainly to consider IntlDateFormatter instead of Punic.

tests/testconfig.ttl
jdfranel: Yes, sorry I forgot to clean it earlier.

osma:
Note: We currently have some other challenges as well related to locale-aware sorting in Skosmos 3. This is apparently broken currently. We used to maintain the current locale according to the selected UI language using `setlocale()` in Skosmos 2 but that got dropped when we switched from legacy `gettext` to Symfony Translations for Skosmos 2. As a side effect, collation (ordering of items, e.g. narrower and related concepts shown on the concept page) is not following the UI language right now. This is a serious bug and needs to be fixed. (Issue will be created soon)

To fix the problem, we could reintroduce the `setlocale()` calls, but I think it would be better to modernize the collation to use Collator from the Intl extension. A mock-up for doing collation in such a way is in PR #1952.

The system locales that are set using `skosmos:languages` in config.ttl don't have much effect now that `getlocale()` is not used. I think we could change this so that the values would be BCP47 locales, which are compatible with ICU/Intl extension as well as Punic. Then there would be no need for complex conversion logic.

In any case I think it's better to use the intl extension instead of Punic for printing timezone/locale-aware dates.

osma:
The collation issue is described in #1955. Fixing it will probably affect this PR; it will remove the need for translating between system locale names and BCP47 locale names, as we will very likely switch to BCP47 locale names in the configuration.

src/util/DateTimeHelper.php
jdfranel: I put together a version using `IntlDateFormatter` and dropping `DateTimeHelper` alltogether. I need to run all test and make sure everything is fine before pushing it but it should be ready soon.

It is working fine but there is still a catch with how the languages are defines in the skosmos config.

I get sometimes confused with the different methods available `DataObject->getLang()`, `ConceptPropertyValue->getLang()` (which seems to override the `DataObject` method despite returning fundamentally the same thing),  `Model->getLocale()` (that is not named `getLang` but is what `getLang` return) that all return langage labels `en`, `de`, ... and `GlobalConfig->getLanguages()` that is a mapping of label to locale codes. That feels like some refactoring is due at some point for clarity sake but it should probable be an independent issue.

So I had to add extra steps to get the locale matching the current UI language (which was handled by `DateTimeHelper`.

```php
$localeLabel = $this->model->getLocale();
$locale = $this->model->getConfig()->getLanguages()[$localeLabel];
$dateFormatter = new IntlDateFormatter(
    $locale,
    IntlDateFormatter::FULL,
    IntlDateFormatter::FULL,
    $this->model->getConfig()->getTimezone(),
    );
// Use interface language for date formatting, not content language
$val = ucfirst($dateFormatter->format($val->getValue()));
```

Nevertheless it feels less like tinkering and reduce legacy compared to the initial proposition I made.

There is still a small issue though: When skosmos:languages value is specifying the collation IntlDateFormatter is not inferning the correct locale. So `en_GB.utf8` will make the date format to be the default english (ie: `en_US`) instead of the expected one. Removing the collation from the skosmso:languages values fixes it and I did not notice any issue somewhere else in the application.

Is this locale with collation important in some cases I did not think about ?

osma: Thanks, this is excellent progress!

You are absolutely right. The system locales currently listed as `rdf:value`s in `config.ttl` are pretty much unused in the current situation; this was also noted in the issue #1955 that I just created.

I think the fix will be to start using BCP47 locales (ICU/intl and Punic compatible) directly in config.ttl and also make these a bit easier to access so you don't have to do it like `$this->model->getConfig()->getLanguages()[$localeLabel];`. There will be a choice between the current `lang` (which is often a simple value like `en` and is used e.g. in URLs) and the corresponding, more detailed `locale` (which can be e.g. `en-US`).

Let me see if I can make a fix to #1955. It's a rather wide-ranging issue though, so may take a while to resolve. But after that is done, this PR can be simplified further.

jdfranel: That would definitely be an improvement. I am not convinced the the whole label + locale is too much because that could allow mapping of an `en` URL to a specific `en-GB` which makes sens for many cases (I'm located in Switzerland and we have our share of `de-CH` or `fr-CH` locale to handle sometimes).

I would suggest the following approach when addressing issue #1955:
- Improving consistency of method names by using either using `Locale` when returning a BCP47 string that is targeted for formatting string or selecting a matching Literal and using `LanguageLabel` when returning a shorter "URL friendly" representation of the locale.
- Store the current UI language as a `Language` object containing a "locale" and "label" property.

Anyway, I'll push the tested changes and wait for this other issue to be done and I'll adapte then.

jdfranel: All done. I pushed the all `IntlDateFormatter` implementation. Although I kept the locale manipulation to remove the collation so the formatting is correct. This will likely be removed after #1955.

**Important change notice**:
While removing Punic date formatter I noticed the handling of date properties in concepts was not coherent with how the `dc:modified` so I added support for both `xsd:date` and `xsd:dateTime`. This is tested in https://github.com/NatLibFi/Skosmos/pull/1939/changes/43c54b8c0982541b7f045cb539d6cf05db613154#diff-1254b70a889531d53ef7ad80e03747cd1ff12c0522bbe6cb0394d58e0b665d2bR56 and https://github.com/NatLibFi/Skosmos/pull/1939/changes/43c54b8c0982541b7f045cb539d6cf05db613154#diff-817d503e3f07250fd4eb858a007679149f95728cd8445886c51580585a55ce04R205

osma: @jdfranel Great! You can find the proposed fix for #1955 in #1957, which still needs to be reviewed before merging. With that in place, to get the BCP47 locale, you just have to call `$model->getLocale()`.

osma: Or alternatively, perhaps there could be a new `Model` method `getDateFormatter` that would return the properly initialized `IntlDateFormatter`. This would be in line with the `getCollator` and `getTranslator` methods that already exist in `Model`.

jdfranel: Since the date format (full, short, none, ...) and time format is set in the formatter once and for all and there is a couple of different date formatting between `Vocabulary`, `Concept` and `ConceptProperty` we would need to offer some options. I see 3 possibilities:

1. (worst) create either a bunch of `dateFormatter` (full dateTime, short dateTime, short date, ...) ready to be used (which would be a nightmare to use and maintain)
2. (less code) have a `getDateFormatter(int $dateType = IntlDateFormatter::FULL, int $timeType = IntlDateFormatter::FULL)`
3. (cleaner) have a  `getDateFormatterBuilder()` that is initialised by model with the proper locale and timezone and default to full dateTime format but we can do `$builder->setDateType(IntlDateFormatter::SHORT)->setTimeType(IntlDateFormatter::NONE)->build()->format($date)` when we need to.

osma: @jdfranel ah, I see, then it's a little bit more complicated than I thought.

I agree that 1 is a nightmare and should be avoided. Either 2 or 3 would be fine. In fact I would prefer 2 because it's so simple.

#1957 has now been merged, so I think all the infrastructure is now in place.

jdfranel: Thank you for the work on #1957. I just adapted the PR and pushed the changes. That now feels like a clean implementation and less like a workaround.

src/model/Concept.php
```diff
@@ -741,45 +741,63 @@ public function getModifiedDate()
         return $this->getVocab()->getModifiedDate();
     }
 
+    /**
+     * @param string $name Resource property
+     * @return string|null The resource date formated as string or null
+     */
+    private function getFormattedDateResource($name)
```
osma: This new method is not covered by PHPUnit tests and Codecov complains about the reduced coverage. Any chance to add a unit test or few to cover this?

osma (approved):
I tested this on our data and I think it works well enough. The new date strings are sometimes a bit verbose but I think we can live with that.

One small request: could you please add some unit tests (see comment)? Other than that, this is good for merging.

src/model/Concept.php
jdfranel: I changed the `getFormattedDateResource` method to public and added some tests.

osma:
@jdfranel Thanks a lot for your effort and persistence, I am merging this now and it will be included in the 3.3 release that will follow soon.

jdfranel:
Great. Thanks your for taking the time to consider my PR. Happy to contribute.

Issue #1940, opened 2026-03-03
Finto layout improvements


UnniKohonen:
## Description of the enhancement

Finto layout is still missing several features. These should be implemented in `finto.css`.

## What new functionalities would the enhancement make possible?

header
- [x] logo
- [x] finto ai link
- [x] border between nav links and languages

landing page
- [x] background image
- [x] logo
- [x] vocab link weight
- [x] welcome/mastodon/etc slots
  - [x] bg color
  - [x] link color
  - [x] triangle

vocab home/concept
- [x] sidebar tab border
- [x] alphabetical index bg
- [x] hierarchy buttons
- [x] property label border
- [ ] suggestion link top border

vocab/global search
- [x] result count border
- [ ] result count weight

about
- [x] bg colors
- [x] link weight
- [x] heading text/size

feedback
- [ ] vocab selector chevron
- [ ] bg colors
- [ ] text color

osma:
Next steps:

- [x] adjust yellow highlights for multiple rows
- [ ] add footer with Finto contact information etc.

PR #1948, opened 2026-03-05
Add topbar custom template

UnniKohonen:

A custom template slot is needed in the topbar to add custom navigation links. This PR adds the slot. UI language is also added as a class to the logo in the topbar.

- Part of #1940

- Add new folder in custom-templates for topbar
- Add `#topbar-slot` div to topbar
- Add class for UI language to topbar logo
- Add border between topbar links and languages

## Checklist

### DIFF
--- a/src/view/base-template.twig
+++ b/src/view/base-template.twig
@@ -64,9 +64,7 @@
                    <i class="fa-solid fa-comment"></i>&nbsp;{{ "Feedback" | trans }}
           </a>
         </li>
-        <div id="topbar-slot">
-          {{ include('custom-templates.inc.twig', {slotName: 'topbar'}) }}
-        </div>
+        {{ include('custom-templates.inc.twig', {slotName: 'topbar'}) }}
         {% if languages|length > 1 %}
         {% for langcode, langdata in languages %}
         {% if request.lang != langcode %}


src/view/base-template.twig
```diff
@@ -64,6 +64,9 @@
                    <i class="fa-solid fa-comment"></i>&nbsp;{{ "Feedback" | trans }}
           </a>
         </li>
+        <li class="nav-item" id="topbar-slot">
```
osma: This will insert an empty `<li>` element even in the case that the slot is empty.

I think it would be better to drop the `<li>` here and instead assume that the twig templates in the topbar slot will contain their own `<li>` elements. That would also allow inserting multiple independent items.

src/view/base-template.twig
```diff
@@ -64,6 +64,10 @@
                    <i class="fa-solid fa-comment"></i>&nbsp;{{ "Feedback" | trans }}
           </a>
         </li>
+        <div id="topbar-slot">
+          {{ include('custom-templates.inc.twig', {slotName: 'topbar'}) }}
+        </div>
+        <li class="nav-item ms-3 ">|</li>
```
osma: This is not very semantic HTML. The line is just a visual element, could we do it in CSS instead?

src/view/base-template.twig
```diff
@@ -64,6 +64,10 @@
                    <i class="fa-solid fa-comment"></i>&nbsp;{{ "Feedback" | trans }}
           </a>
         </li>
+        <div id="topbar-slot">
```
osma: I think it's invalid HTML to have a `div` element in between `ul` and `li`.

Maybe we can just drop this div element? That unfortunately means that there is no ID to target in CSS, though.

Issue #1420, opened 2023-03-17
Collation doesn't work in Northern Sámi in all environments


nikopartanen:
## Description of the enhancement

Currently when Northern Sámi is selected the hierarchy view displays the terms following a wrong collation. It's unclear which collation system is followed, but many characters are in unexpected places. In the example below the letter **ä** in the beginning should be in the end separately, and **a** followed by the entries starting with **á**. Comparing the hierarchy view and the list of narrower concepts [here](https://dev.finto.fi/yso/se/page/p8691) displays the difference well, as in the latter we do have the correct Northern Sámi collation.

![image](https://user-images.githubusercontent.com/6486495/225862573-c621657f-fda6-40b4-8bbd-95a4ce448300.png)

Compare with the correct:

![image](https://user-images.githubusercontent.com/6486495/225870462-24195b4e-b495-4655-bcd8-02df2d7f6392.png)
![image](https://user-images.githubusercontent.com/6486495/225870507-417681b4-23d3-47a8-9af3-90aa61a8464e.png)

This seems to happen everywhere in this part of the interface, also within the groups-view:

![image](https://user-images.githubusercontent.com/6486495/225870790-b84d4eae-a77f-4d25-8878-5b00f6412d76.png)

It seems that the search term autocompletion results are also one part that works correctly:

![image](https://user-images.githubusercontent.com/6486495/225873184-545db868-034e-43bb-a8ff-f80009afdca1.png)

The proposed enhancement would be to use the same collation rules in all environments within Skosmos. 

## Who are the users that would benefit from the enhancement and how?

All users of Skosmos in Northern Sámi and other languages that may have differently described collation rules in different programming environments. In the case of Northern Sámi the [CLDR collation](https://github.com/unicode-org/cldr/blob/main/common/collation/se.xml) rules seem to be essentially correct.

## What new functionalities would the enhancement make possible?

It would make the interface more consistent for the users. Now a given term is often in a different location in a list in different views. 

## Why is the enhancement important?

The ability to provide a consistent and uniform user experience in a wide array of languages is important when Skosmos is taken into use in different countries and regions of world.

joelit:
Looking quickly at the mechanism this bug can be replicated, it seems that the collation works correctly when the UI language is in Finnish (and the content language is Sami), but it stops working when the UI language is set to Sami:
 * Works: https://dev.finto.fi/yso/fi/page/p8691?clang=se
[...]

PR #1952, opened 2026-03-18
Create a mock-up to make collations follow the selected languages ins…

miguelvaara:
This is a mock-up intended to provide one example of how collation could be handled on the backend side (it will not work on the frontend due to the lack of Northern Sámi support in Chrome).

Previously the user interface and content could be displayed in multiple languages, but the actual sorting was in most cases determined by the runtime settings rather than by those languages.

Further investigation is required to determine how well the proposed model works in all cases and whether it can be generalized to all functionality that requires collation.

It is also worth considering whether there might be a more straightforward and reliable approach.

Attached screenshots show the situation before the mock-up. So far no errors have been dtected in Finnish, Swedish, Northern Sámi, or English when switching the content language and the user interface language across all possible permutations.

Screenshot from 2026-03-18 12-53-24
Screenshot from 2026-03-18 12-53-08

#1420

- Closes #

**Model.php**
The comparison used for narrower sorting was moved into a single backend method (_compareStringsByLanguage_) so collation no longer depends on runtime locale. It uses the following language priority (in line with already existing practise) clang -> lang -> default plus the language mapping from config.ttl.

**ConceptProperty.php**
In the _skos:narrower_ case, the previous strcoll comparison was replaced with the centralized "language-aware" comparison so concept page narrower ordering follows the selected content language correctly.

**ConceptPropertyValue.php**
Submembers sorting was changed from ksort logic to the same backend collation path (uasort + collation context), so the sublist follows the same language-based ordering as the main narrower list. Uasort() sorts an array by its values using a user-defined comparison function.

_getCollationContext()_ was added (contentLang/uiLang/defaultLang) so sorting receives explicit request language context instead of deriving/calculating it implicitly from the environment.

It is a mock-up, we cannot wait it meets all the requirements

## Checklist

### DIFF
--- a/src/model/ConceptProperty.php
+++ b/src/model/ConceptProperty.php
@@ -143,6 +143,16 @@ class ConceptProperty
             } else {
                 uasort($this->values, function ($a, $b) {
                     // assume that sort keys are unique
+                    if ($this->prop === 'skos:narrower') {
+                        $context = $a->getCollationContext();
+                        return $this->model->compareStringsByLanguage(
+                            $a->getSortKey(),
+                            $b->getSortKey(),
+                            $context['contentLang'],
+                            $context['uiLang'],
+                            $context['defaultLang']
+                        );
+                    }
                     return strcoll($a->getSortKey(), $b->getSortKey());
                 });
             }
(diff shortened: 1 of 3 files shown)


osma:
Converted to draft as this is clearly just a mock-up and not intended to be merged.

osma:
Superseded by #1957. Closing this one.

PR #1958, opened 2026-03-19
Hover color for vocab search lang list text

joelit:

- Closes #

## Checklist

### DIFF
--- a/resource/css/skosmos.css
+++ b/resource/css/skosmos.css
@@ -426,6 +426,10 @@ header .container-fluid .row {
   user-select: none;
 }
 
+#vocab-list li:hover label {
+  color: var(--headerbar-text) !important;
+}
+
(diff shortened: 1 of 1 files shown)


resource/css/skosmos.css
```diff
@@ -281,7 +281,7 @@ header .container-fluid .row {
   color: var(--headerbar-text);
 }
 
-#search-wrapper .dropdown-menu li label:hover {
+#search-wrapper .dropdown-menu li label:hover, #search-wrapper .dropdown-menu li a:hover {
   color: var(--search-dropdown-hover-text);
 }
```
UnniKohonen, line 286: Something like this is still needed for vocab list:

```
#vocab-list li:hover label {
  color: var(--headerbar-text) !important;
}
```

PR #2000, opened 2026-04-14
Fix property label tooltip underline

UnniKohonen:

Dotted underline on property labels with tooltips extends past the text in finto layout. This PR fixes the problem by using underline text decoration instead of a border. This also fixes the issue in displaying the yellow highlight described in #1940.


osma (approved):
Excellent!

Image

PR #2001, opened 2026-04-14
Adjust warning message for deprecated concepts (See: X) and improve its accessibility

### LINKED ISSUE #1983: [Accessibility] WCAG 1.1.1. information about deprecated term is quite unnoticeable with screen readers
https://test.dev.finto.fi/yso/fi/page/p20945

## Description of the problem
If some concept is derecated, it will be warned very visibly with a large red alert text and warning icon. However, this area of attention is not particularly pronounced with screen readers. The screen reader reads only the text in the section and this notification is easily missed. 

**Fix**: add alt text to the exclamation point icon, e.g. "Warning!", "Note!", or "Important!"
If the focus does not move to the main content from the activation of the term, but the Aria-Live notification is used to inform about the loading of new content, then the text in the attention area would be read in the live notification.

Severity: moderate
Priority: medium

Image

Chrome, Firefox, Edge

osma:

The warning message for deprecated concepts included the phrase "Is replaced by: X". This was widely seen as problematic by YSO maintainers. This PR replaces it with a simpler phrase "See: X" and adds translations in Finnish ("Katso: X") and Swedish ("Se: X").

Example:
image

Also, the accessibility of the warning message is improved by adding aria-role="warning" to the div element, adding aria-hidden="true" to the warning symbol icon, and adding a "Warning!" prefix that is only visible for screen readers (hidden with the class visually-hidden).

- follow-up fix to #1918
- fixes #1983 

* change phrase to "See: X"
* add role="warning"
* add aria-hidden="true" for the Font Awesome warning symbol icon
* add a "Warning!" prefix which is only visible for screen reader users
* add fi+sv translations for "See:" and "Warning!"
* (unrelated) add back some English translations in messages.en.json that mistakenly got dropped in #1959. These are all of the form `"X": "X"` (both key and message are identical) so they don't really change the translations.

n/a

## Checklist


joelit (approved):
LGTM! The only strangeness I could reproduce is a case where the disclaimer text has been translated into the UI language but the 'see' texts hasn't. Change the content language to something else and you get a situation like this where the you have a short English word sanwhiched between a Norhtern Sámi disclaimer and a Finnish concept label:

Image

Is it too confusing? Probably not, but I thought to mention it here.

osma:
@joelit You are right, but this is a more general problem with all UI translations - if they are incomplete (and they very often are), then you will get mixed up languages. Maybe it's a bit more pronounced here though since they are displayed next to each other. If we really wanted to avoid the, we could turn the whole message into a single translated string instead of assembling it from smaller pieces.

PR #2006, opened 2026-04-15
Add fontawesome as npm dependency

UnniKohonen:

Fontawesome is not currently included in the NPM dependencies of Skosmos. This PR adds it as a dependency to package.json and removes the old css and font files from resources.


joelit (approved):
LGTM! Looks like this is a bit of a rabbit hole researching whether to use the online content delivery network, or to host a font locally. Both have benefits.

Issue #2009, opened 2026-04-16
[Accessibility] Unified keyboard focus highlight for UI components


joelit:
## Description of the enhancement
- There are various visuals for showing keyboard focus (ad-hoc solutions, bootstrap defined and browser native)
- They need to be unified in a way that makes sense in the Skosmos colour palette

- [ ] Decide how the keyboard focus should look like and check the design across different contexts
- [ ] Implement the keyboard focus in CSS
- [ ] Apply the focus across the different elements

## What new functionalities would the enhancement make possible?
- having an easy way for handling focus in CSS makes it easy to keep the same logic accross old and new Skosmos UI components and plugins

PR #2012, opened 2026-04-21
Accessibility improvements for alphabetical tab letter selection

### LINKED ISSUE #1979: [Accessibility], WCAG 2.4.7, 4.1.2. There are a few accessibility problems on the A-Z tab
## Description of the problem
There is a tab structure in vocabulary pages. A-Ö tab contains letter buttons. There are a few accessibility issues:
1.	There is no visible focus indicator when the letters are focused. Keyboard users just need to guess which component I focused at the moment.
2.	Letters have the link role. When the letter is activated, the content of the column is updated, but the rest of the content of the page remains the same. The keyboard focus remains on the clicked letter. The screen reader user does not get any information about whether activating the link has done anything at all. The link role is also misleading for screen reader users. Typically, a link opens a new page or context changes, while a button triggers an action on the same page.

**Fix:** 
1.	Add visible and clearly distinguishable focus indicator around the letter buttons.
2.	A screen reader user should get some feedback about the success of the operation. The role of the element should also be changed, because the action expected from the Link role does not correspond to the actual event.
Option that would be more understandable with a screen reader is that the letters in terms of semantics are radio buttons (state selected/unselected)

Also, at the moment, no user is shown in any way which letter is selected. Sighted users can check this from the glossary list below. All users would benefit if the active letter was also highlighted in the letter listing (recommended fix.)

How to test: 
1.	Move with tab key (no screen reader).
Expected result: When the letters are focused, there are clearly distinguishable focus indicator around the letter, eg blue & white borders.

How to test: 
2. Turn on NVDA. Navigate to the letter section
Expected result: When the letters are focused, screen reader announces [letter] + radio button checked / not checked role and status.

Severity: Serious
Priority: High

Image

Chrome, Firefox, Edge

UnniKohonen:

Letter selection in alphabetical index sidebar tab is not currently accessible for screen readers and keyboard users. This PR changes the letters from links to radio buttons. Letters should now be focusable with a tab click and navigable with the arrow keys. Selecting a new letter using the arrow keys should load new concepts for that letter. Behavior should conform to expectations set [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/radio_role).

- Closes #1979

- Switch from using bootstrap's `.pagination` class to `.form-check` classes and hidden radio inputs
- Add an aria live element to tell screen readers that new concepts have been loaded
- Add an underline and a dotted outline to focused letter
- Add an underline to currently selected letter 

- No Swedish translation for aria loaded message
- If moving quickly between letters, screen reader sometimes reads messages out of order

## Checklist


joelit (approved):
LGTM! The overall look and feel of the keyboard focus should be decided in #2009 . This PR is a fine a11y improvement, overall.

PR #2013, opened 2026-04-22
Add aria message as a hidden span to sidebar concept lists

### LINKED ISSUE #1980: [Accessibility], 2.4.4. WCAG Term selection in Vocabularies is not accessible for screen readers
https://test.dev.finto.fi/yso/en/ and other vocabulary pages

## Description of the problem
Screen readers don't read the terms in the sidebar concept list correctly. For each link in the listing, aria-label="Mene käsitesivulle" has been added as a kind of supplementary text. However, the aria-label overrides the other content of the link, so screen readers only read the text “Mene käsitesivulle”/"Go to concept page" but not the visible name, e.g. "aakkostus", "AA liike", etc. The link list is therefore currently unusable for screen readers.
The problem is repeated in all tabs and in all the vocabularies studied.
This is a blocker issue for screen reader users.

**Fix**: Aria-label should not be used to add only additional information. Aria-label can be used to add the name of the link/ treeitem in its entirety (e.g. aria-label="alphabet, open concept"). However, if you only want to add additional information after the term, the recommended way to do it is to add an invisible additional information text, which can be found with screen readers https://www.w3.org/WAI/WCAG21/Techniques/css/C7 

**How to test:** Turn on  NVDA. Navigate to vocabulary tab panel. Open some tab, that contains links or treeitems. Navigate to tab panel and browse the links or treeitems:

Expected result: NVDA reads the names of the links
Actual result: NVDA reads “Go to the concept page / Mene käsitesivulle”

Severity: Critical
Priority: High

Image

Image

Chrome, Firefox, Edge

UnniKohonen:

Concept page links in sidebar are not currently accessible to screen readers. This PR moves the "Go to concept page" message to be read after the name of the concept.

- Closes #1980

- Remove aria labels that were blocking concept names from being read
- Add a hidden span element for each concept containing the message
- Move focus to prefLabel heading on concept card after clicking on concept links

- The same text is used as before, is it still OK in this use case?
- Clicking shift+tab after partial page load does not bring focus back to previously clicked concept link

## Checklist


osma (approved):
LGTM, it does what it says and the focus moves. Not 100% sure this is the best approach, but certainly better than what we had before.

PR #2014, opened 2026-04-22
feat: add custom label overrides and tooltips for vocabulary properties

jdfranel:

Adding properties to a vocabulary other than those translated in the source code results in a raw property key in the vocabulary page. If I need idot:prefix term in my vocabulary it is not translated.

*None*

I changed the vocabulary template so property label overrides defined in the skosmos.ttl vocabulary definition is applied to its properties.

*None*

## Checklist

### DIFF
--- a/tests/cypress/template/concept.cy.js
+++ b/tests/cypress/template/concept.cy.js
@@ -158,10 +158,15 @@ describe('Concept page', () => {
   it('overrides vocabulary property labels', () => {
     // Go to "Carp" concept page in vocab with property label overrides
     cy.visit('/conceptPropertyLabels/en/')
-    // Check that there is a .row .property-label with the overridden property label
-    cy.get('.row .property-label').eq(0).invoke('text').should('include', 'Vocabulary Property')
-    // Check that there is a .row .property-label with the overridden property label description as title
-    cy.get('.row .property-label').eq(0).should('have.attr', 'title').and('contain', 'Vocabulary Property description')
+    // Loop through all .row.property and if the element has a .property-value ul li with text "Overridden property value", check the label and title of the .property-label for this row
+    cy.get('.row.property').each($row => {
+      cy.wrap($row).find('.property-value ul li').each($li => {
+        if ($li.text().includes('Overridden property value')) {
+          cy.wrap($row).find('.property-label h2').eq(0).invoke('text').should('include', 'Vocabulary Property')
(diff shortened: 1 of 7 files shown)


osma:
Hi @jdfranel , thanks for the PR and sorry for the slow response.

Skosmos already supports custom properties and you can add labels to them in RDF data which will be used in the display, like this:

```
my:customProperty a rdf:Property; rdfs:label "Custom Property"@en .
```

This can be in the same RDF graph as the vocabulary, or in the default graph of the SPARQL endpoint.

Would this solve your original issue or is the functionality in this PR somehow better from your perspective?

jdfranel:
AFAIK this only applies to concept properties. Not to vocabularies, so I basically reused the same logic from the Concept to the vocabularies.
If it is supposed to also work for vocabularies, I might have missed something.

osma:
@jdfranel Ah, right. That's probably the case.

Would this PR be better than implementing the RDF property label functionality for vocabularies as well?

jdfranel:
@osma I am not sure to understand what you mean by the RDF property label functionality.

In this PR I edited the WebController so if a `skosmos:propertyLabelOverride` is set in the skosmos.ttl for a vocabulary, it also aplies to the vocabulary page by using the same `$vocab->getConfig()->getPropertyLabelOverrides();` method.

In the vocab-info.inc.twig I inserted a similar template code than the one handling `custom_labels` in the concept-card.inc.twig for the property.

tests/cypress/template/concept.cy.js
```diff
@@ -155,6 +155,19 @@ describe('Concept page', () => {
     // Check that mapping property title is overridden correctly
     cy.get('.prop-mapping .property-label').eq(0).should('have.attr', 'title').and('contain', 'Exactly matching classes in another vocabulary.')
   })
+  it('overrides vocabulary property labels', () => {
```
osma: This test is in the wrong file. Can you move it to vocab-home.cy.js instead where it belongs?

(The Cypress tests under template/ are organized mainly according to the page type / template they test. This file concept.cy.js is for the concept page.)

osma (changes requested):
Thanks for the explanation. I think this is a valid feature.

One Cypress test is in the wrong file, otherwise this looks OK.

tests/cypress/template/concept.cy.js
jdfranel: I moved the test in the vocab-home.cy.js file.

osma (approved):
LGTM, thanks! Merging this now, it will go into the 3.3 release Real Soon Now

PR #2004, opened 2026-04-14
Accessibility improvements for global & vocab search

joelit:
## Link to relevant issue(s), if any

- Closes #1974 
- Closes #1971 

## Description of the changes in this PR
 
 - [x] Add translation template for clear search button text
 - [x] Add aria label for vocab search clear search button
 - [x] Add aria label for global search clear search button
 - [x] Make sure the focus stays in the search field after pressing the clear search field button for keyboard accessibility
 - [x] Update cypress tests accordingly

## Checklist

- [x] phpUnit tests pass locally with my changes
- [x] I have added tests that show that the new code works, or tests are not relevant for this PR (e.g. only HTML/CSS changes)
- [x] The PR doesn't reduce accessibility of the front-end code (e.g. tab focus, scaling to different resolutions, use of `.sr-only` class, color contrast)
- [x] The PR doesn't introduce unintended code changes (e.g. empty lines or useless reindentation)

PR #2017, opened 2026-04-22
Issue1889 vocab search a11y fixes

### LINKED ISSUE #1889: Vocabulary Search Bar: Add keyboard support for accessibility

joelit:

- Closes #1889

This PR should be based upon PR #2004 as it improves the keyboard navigation on top of the previous accessibility improvements.

## Checklist


joelit:
Rebased to issue1889-vocab-search-a11y-fixes and force pushed.

UnniKohonen (changes requested):
Keyboard navigation seems to work correctly for the most part and screen reader messages seem to be correct.

There were a couple of issues I noticed:

Both global and vocab search have a problem in firefox where if tab is pressed when focused on dropdown options, focus moves to the next field and then immediately back. This causes the screen reader to read both aria labels. It seems like this issue was not present previously so it should be maybe fixed. In chrome, tab moves focus to the next dropdown option which is probably fine.

Dropdown chevrons have moved in the wrong position. These rules should be included for the `.chevron` class to fix the problem:
 ```
right: 1rem;
top: 50%;
transform: translate(0, -50%);
```

and for `.dropdown-toggle`:
```
padding-right: 2.5rem
```

Vocab search language dropdown hover and focus effects have also changed. If this is not intentional, they should be fixed (it seems to be at least partially because language selection list no longer has the `language-list` id).

There is also a failing cypress test in vocab search page but it probably doesn't originate from this PR.

joelit:
> Keyboard navigation seems to work correctly for the most part and screen reader messages seem to be correct.
> 
> There were a couple of issues I noticed:
> 
> Both global and vocab search have a problem in firefox where if tab is pressed when focused on dropdown options, focus moves to the next field and then immediately back. This causes the screen reader to read both aria labels. It seems like this issue was not present previously so it should be maybe fixed. In chrome, tab moves focus to the next dropdown option which is probably fine.

I cannot reproduce this behavior in Firefox or Chrome. I get the expected bahavior where tab moves the focus to the next menu, as intended.

> Dropdown chevrons have moved in the wrong position. These rules should be included for the `.chevron` class to fix the problem:
> 
> ```
> right: 1rem;
> top: 50%;
> transform: translate(0, -50%);
> ```
> 
> and for `.dropdown-toggle`:
> 
> ```
> padding-right: 2.5rem
> ```

Thanks! Fixed as suggested.
 
> Vocab search language dropdown hover and focus effects have also changed. If this is not intentional, they should be fixed (it seems to be at least partially because language selection list no longer has the `language-list` id).

That was not intentional. Thanks for spotting it (I wish we had tests for CSS too). Fixed.

> There is also a failing cypress test in vocab search page but it probably doesn't originate from this PR.

Issue #1976, opened 2026-04-09
[Accessibility], WCAG 1.1.1. The name of the service is not clear to screen readers. The structure of the page is not logical in mobile view and with screen readers, recommended to fix


Eiskore:
## URL address of the page where you encountered the problem
https://test.dev.finto.fi/en/

## Description of the problem
Screen reader users do not get the informationl from the beginning/top of the page that this is a Finto service. The Finto logo is centrally displayed to sighted users and indicates the name of the service, but there is no alternative text to the image. However, the hero image area has a hidden screen-reader-only H1 heading "Skosmos". The name Skosmos, on the other hand, is not conveyed to sighted users.

The Welcome to Finto.fi -section is available in the desktop view on the right side of the service. It pops up for sighted desktop users right at the beginning of the page and acts as an explainer/welcome text for the service. The section also provides instructions on how to use the site.
However, the section does not work as well with screen readers and in mobile view, because then the section is read and shown only at the end of the page, after a long column on the left.

**FIX**: The name of the service Finto should also be displayed on screen readers. 

The most **recommended** fix is a structural change to the front page. Especially considering mobile users and screen reader users, the structure of the front page would be much more functional if Welcome to Finto.fi! section would be placed as the first section in the main content, in which case "Welcome to Finto.fi!" would serve as the main title of the page and the section below would serve as a guiding section for the use of the page.

If such an extensive structural change to the page is not desired, at least screen reader users should be better communicated about the name of the service. Then change the hidden H1 title of the page to match the service name "Finto".

## Additional information (e.g. screenshots) about the problem

<img width="1004" height="375" alt="Image" src="https://github.com/user-attachments/assets/acbbd7eb-2995-4aac-83e7-578183289f94" />

<img width="1004" height="443" alt="Image" src="https://github.com/user-attachments/assets/e0d56400-1d3b-4b34-b53a-024c8931b4e2" />

<img width="250" height="732" alt="Image" src="https://github.com/user-attachments/assets/2230aa21-6ad4-4e7a-9888-3a5694d2daf2" />
Welcome to Finto! -section is at the end of the page on mobile view and to screen reader users.

<img width="1004" height="616" alt="Image" src="https://github.com/user-attachments/assets/bcb9b9c3-a7bd-467b-a6cf-9d91aeb5dbff" />
It is recommended to place "Welcome to finto! -section" as a first section in the main content.

Vainonen:
[...]

PR #2018, opened 2026-04-23
Replaced h1 element with service name

Vainonen:
Screen reader reads hidden h1 element which has "Skosmos" in it. Also welcome text should not be visible with mobile view.

- Partially addresses #1976

h1 element has now long service name defined in config file. If the language version of name is missing it falls back to shorter name in config file, which does not have language version. If shorter name is missing, it falls back to "Skosmos".

[Custom template](https://github.com/NatLibFi/Finto-data/blob/master/conf/test.dev.finto.fi/custom-templates/landing-end/0-welcome.twig) for landing-end has been modified with Bootstrap calls to hide the welcome text from mobile view.

According to Hahmo layout, Finto front page should have mobile navigation bar which opens a page with welcome text in it. This navigation bar is not implement yet.

## Checklist

### DIFF
--- a/src/view/base-template.twig
+++ b/src/view/base-template.twig
@@ -98,7 +98,7 @@
         <div class="d-flex flex-column flex-xl-row align-items-start align-items-xl-center my-auto{% if pageType == 'landing' %} p-0{% endif %}">
           {% if pageType == 'landing' %}
           <div id="skosmos-logo">
-            <h1 class="visually-hidden">{{ GlobalConfig.serviceNameLong(request.lang) }}</h1
+            <h1 class="visually-hidden">{{ GlobalConfig.serviceNameLong(request.lang) }}</h1>
           </div>
           {% else %}
           <h2 class="fw-bold" id="vocab-title">
--- a/src/view/base-template.twig
+++ b/src/view/base-template.twig
@@ -98,7 +98,7 @@
         <div class="d-flex flex-column flex-xl-row align-items-start align-items-xl-center my-auto{% if pageType == 'landing' %} p-0{% endif %}">
           {% if pageType == 'landing' %}
           <div id="skosmos-logo">
-            <h1 class="visually-hidden">Skosmos</h1>
+            <h1 class="visually-hidden">{{ GlobalConfig.serviceNameLong(request.lang) }}</h1
           </div>
           {% else %}
           <h2 class="fw-bold" id="vocab-title">


src/view/base-template.twig
```diff
@@ -98,7 +98,7 @@
         <div class="d-flex flex-column flex-xl-row align-items-start align-items-xl-center my-auto{% if pageType == 'landing' %} p-0{% endif %}">
           {% if pageType == 'landing' %}
           <div id="skosmos-logo">
-            <h1 class="visually-hidden">Skosmos</h1>
+            <h1 class="visually-hidden">{{ GlobalConfig.serviceNameLong(request.lang) }}</h1
```
osma:
```suggestion
            <h1 class="visually-hidden">{{ GlobalConfig.serviceNameLong(request.lang) }}</h1>
```

Part of the closing tag is missing.

osma (approved):
There is a syntax error in the closing tag (you can simply apply my suggested fix), otherwise this looks OK and works fine.

This is an improvement for sure, but I don't think it covers the whole issue. I think we should still either leave the issue open or close it and open a follow-up issue about the HTML DOM order, which is still problematic for both screen reader and mobile users. I think it should be possible to change the DOM order while preserving the visual layout for sighted users (with the welcome box in the right column) if we use CSS Grid for the layout. Here is a minimal example of how it could work:

```html
<html>

<head>
<style>
.landing-layout {
  display: grid;
  grid-template-columns: 1fr; /* mobile: single column */
  gap: 1.5rem;
}

/* Desktop layout: 2 columns (8/4 split) */
@media (min-width: 768px) {
  .landing-layout {
    grid-template-columns: 2fr 1fr; /* left: 66.7%, right: 33.3% */
    grid-template-areas:
      "vocab welcome"
      "vocab news";
  }

  .box-vocab {
    grid-area: vocab;
  }

  .box-welcome {
    grid-area: welcome;
  }

  .box-news {
    grid-area: news;
  }
}


</style>

<body>

<div class="container">
  <div class="landing-layout">
    <!-- 1. Welcome (top-right visually, first in DOM) -->
    <section class="box box-welcome">
      <h2>Welcome to Finto</h2>
      <p>...</p>
    </section>

    <!-- 2. List of vocabularies (left column, second in DOM) -->
    <section class="box box-vocab">
      <h2>Vocabularies</h2>
      <ul>...</ul>
    </section>

    <!-- 3. News (top-right visually, below welcome, third in DOM) -->
    <section class="box box-news">
      <h3>News</h3>
      <ul>...</ul>
    </section>
  </div>
</div>
</body>
</html>
```

src/view/base-template.twig
Vainonen: Hmm, I was too hasty and the tests were missing for base template. Thanks for noticing this.

PR #2020, opened 2026-04-24
Changed to asynchronous test case for loading more results

### LINKED ISSUE #2019: Failing cypress test for search results page
yso/en/search?clang=en&q=a

## Description of the problem
Steps to produce the problem, and the expected outcome after it has been fixed

Apparently, the way incremental loading of new content has changed broke a search result test. The result page is expected to load the rest of the results, but using a test case that these days loads more results twice in 5-result intervals is at best flaky.

The test case should be changed to something more stable that still shows the incremental loading functionality.

joelit:

The cypress tests are failing for a unit test reoading more search results. This fixes it by changing to a more robust test case for testing the functionality of the same thing.

- Closes #2019

Changed to a test case that returns only slightly more than 5 results

## Checklist

### DIFF
--- a/tests/cypress/template/vocab-search-page.cy.js
+++ b/tests/cypress/template/vocab-search-page.cy.js
@@ -70,19 +70,19 @@ describe('Vocabulary search page', () => {
     cy.visit(`/yso/en/search?clang=en&q=an`)
     // Check that there are 5 search results
     cy.get('#search-results').find('.search-result').should('have.length', 5)
-    // Scroll to bottom of page
+
+    // Listen to the API call
+    cy.intercept('GET', '/yso/en/search?clang=en&q=an*').as('search')
+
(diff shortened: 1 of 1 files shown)


osma (approved):
The change itself looks OK, but Cypress tests are still failing in CI for other reasons. (I restarted the tests to see if it helps)

joelit:
I tried testing for the appearance and disappearance of the spinner as an indicator that the search results list is complete.

I kept the 5 -> 6 search result list, as it removes all possibility for needing to scroll and test for the spinner twice.

Edit: the spinner approach was also flagging with one CI-cypress test passing and one failing. Instead, I changed the test to waiting for the pagination rearch result API call to finish and then counting if the amount of results have changed from 5 to 6. That seemed to work.

PR #2024, opened 2026-05-28
Sidebar hierarchy and groups tab keyboard navigation

### LINKED ISSUE #1981: [Accessibility], WCAG 4.1.2. The semantics of the tree items on the Hierarchy and Groups tabs are inadequate
## Description of the problem
There is a Hierachy and Groups tabs on vocabularies. Both tab panels contains expendable tree pattern. All the elements in a tree view are semantically links. However, the link is not the correct role in the tree view, because the page does not actually change, and the focus does not move.  Screen reader users also now do not receive any feedback on the activation of the link, or whether the activation has done anything at all.

There are also focusable arrow buttons on tree view.  The screen reader finds the arrow-buttons and reads the name “Open button”. However, the screen reader does not tell whether the structure is open or closed. When you click the button (and the tree view closes or opens), the screen reader is completely silent.

**FIX**: To make the structure and choice of term is clearer with a screen reader, the concept list should be implemented as an ARIA treeview (role="tree" with role="treeitem"), not as a list of links. Selecting a term updates content within the same page but does not change context, so links are semantically incorrect. Expand/collapse arrows shouldn’t be separate buttons; they should be purely visual indicators of state, with expansion controlled via aria-expanded and keyboard arrow keys. 
Tree item selection could be indicated with aria-selected, and changes in the main content area announced using an aria-live region if focus remains in the tree.

Keyboard interaction must follow the ARIA tree pattern: Up/Down Arrow moves between visible items, Right Arrow expands a collapsed item, Left Arrow collapses an expanded item, Enter or Space selects the current item, and Home/End moves to the first or last item. 

More information about tree pattern: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/treeitem_role 

Severity: moderate
Priority: medium

Image

Chrome, Firefox, Edge

### LINKED ISSUE #1982: [Accessibility], recommended to fix: Using the concept lists with the keyboard is heavy and unusable.
## Description of the problem
Using the concept lists of the vocabularies with the keyboard is heavy and unusable. All links in the sidebar (1710 links in Hierarchy view) are tab-focusable. If the keyboard user would like to move the focus to the main content (to the data of the selected term), the keyboard user must first move through all of the nearly 2000 links (or collapse them) before the focus moves to the main content.
This is not a WCAG violation but affect significantly to keyboard usage and is therefore recommended to fix.

**Fix**: 
**Option 1**: Keyboard use would be much easier if activating a tree item moves focus to the beginning of the new content, ideally a level one heading with content that matches the name of the tree item that was activated. Focusing on the heading informs screen reader users that navigation is complete and confirms the destination. when pressing shif+ tab, focus returns where it was on treeview. Also in this case Keyboard interaction is recommended to follow the ARIA tree pattern: Up/Down Arrow moves between visible items, Right Arrow expands a collapsed item, Left Arrow collapses an expanded item, Enter or Space selects the current item, and Home/End moves to the first or last item. 
example about similar implementation. https://dequeuniversity.com/library/aria/tree-view 

**Option 2**: Other option is that the focus stays at the treeitem when the treeitem is activated. The treeitems could only be browsed with arrow-keys (up and down arrows), and when pressing the tab- button  the focus moves to the main content.

Image

Chrome, Firefox, Edge

UnniKohonen:

Hierarchy and groups tabs in the sidebar are not currently keyboard navigable. This PR adds keyboard navigation and screen reader semantics to both tabs.

- Closes #1982 
- Closes #1981

- Add keyboard navigation to hierarchy and group trees (based on interactions defined [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tree_role))
  - Arrow up/down are used to navigate between visible list items
  - Arrow right on a closed item opens it
  - Arrow right on an open item moves focus to first child
  - Arrow left on an open item closes it
  - Arrow left on a closed item closes its parent and moves focus to the parent
  - Enter/Space opens concept page and moves focus to concept page heading
  - Home/End move focus to first/last item
- Give hierarchy/groups lists role `tree` and links role `treeitem`
- Add aria-selected and aria-expanded attributes to links
- Make hierarchy buttons non-focusable and invisible to screen readers and remove aria-labels
- Cypress tests for both Vue components

- Not all accessibility features defined [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tree_role) are implemented
  - Selection is only made with Space/Enter and not when moving between items
  - Type-ahead is not implemented
  - No aria-level, aria-setsize and aria-posinset attributes are used
  - Screen reader gets no indication when children are loaded

## Checklist

### DIFF
--- a/resource/js/tab-groups.js
+++ b/resource/js/tab-groups.js
@@ -194,8 +194,8 @@ function startGroupsApp () {
           for (const node of nodes) {
             // Assign index only if parent is open
             if (parentIsOpen) {
-              counter++
               node.index = counter
+              counter++
             } else {
               delete node.index
             }
--- a/resource/js/tab-groups.js
+++ b/resource/js/tab-groups.js
@@ -9,7 +9,8 @@ function startGroupsApp () {
         selectedGroup: '',
         loadingGroups: true,
         loadingChildren: [],
-        listStyle: {}
+        listStyle: {},
+        visibleConceptCount: 0
       }
     },
     provide () {
@@ -20,9 +21,6 @@ function startGroupsApp () {
       }
     },
     computed: {
-      openAriaMessage () {
-        return $t('Open')
-      },
       toConceptPageAriaMessage () {
         return $t('Go to the concept page')
       }
@@ -110,6 +108,7 @@ function startGroupsApp () {
                     this.setIsOpenAndAddMembers(result, this.selectedGroup, members)
 
                     this.groups = result
+                    this.addIndicesToGroups()
                     this.loadingGroups = false
                   })
               } else {
@@ -117,11 +116,13 @@ function startGroupsApp () {
                 this.setIsOpenAndAddMembers(result, this.selectedGroup, [])
 
                 this.groups = result
+                this.addIndicesToGroups()
                 this.loadingGroups = false
               }
             } else {
               // If we are on vocab home page, simply set groups to result
               this.groups = result
+              this.addIndicesToGroups()
               this.loadingGroups = false
             }
           })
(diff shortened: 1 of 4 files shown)


osma (approved):
I tested this briefly in both Firefox and Chromium and also looked at the implementation, though not in great detail.

I think that the keyboard navigation in the hierarchy and groups feels very intuitive and natural. I'm not an expert on the details of the ARIA recommendations, but this is certainly a big step forward for accessibility!

SonarCloud pointed out that the use of `counter++` within a sub-expression is bad practice, I think that could be fixed.

Other than that, LGTM!

PR #2029, opened 2026-06-04
Fix switching between sidebar tabs using the keybaord

UnniKohonen:

Switching between sidebar tabs using the arrow keys does not currently load the contents of the tabs. This PR fixes the way tab contents are loaded. Contents should now be loaded when tabs are clicked and when switching between them with arrow keys.

- Listen to [bootstrap's tab shown event](https://getbootstrap.com/docs/5.3/components/navs-tabs/#events) instead of click events in sidebar tabs to load tab content
- Add cypress test

## Checklist


osma (approved):
Fix looks sane and works perfectly (tested under Firefox and Chromium).

PR #2038, opened 2026-08-19
Refactor copy to clipboard events using a single event handler

osma:

I need to implement some concept page tweaks in the KANTO/finaf plugin that include adding new copy to clipboard buttons for identifiers like ISNI and ORCID. The existing Skosmos copy to clipboard code attaches separate event handlers for each copy button (preflabel, notation, URI). This is brittle and repetitive, and makes it hard to copy the same approach in the finaf plugin.

This PR refactors the copy to clipboard event handling to use just a single event handler attached to `#main-container`. It responds to events from every copy button on the page (event delegation). The value to copy is indicated using `data-target-id` attributes on each copy button. This makes it easy to copy the same HTML structure in the finaf plugin without having to adjust event handler. Also, this single global event handler survives partial page loads so event handlers no longer have to be reinitialized after partial page loads.

n/a

* switch to single event handler for all copy to clipboard buttons
* add `data-target-id` attributes on individual copy buttons to indicate which element value should be copied

none

## Checklist

### DIFF
--- a/resource/js/copy-to-clipboard.js
+++ b/resource/js/copy-to-clipboard.js
@@ -11,25 +11,19 @@ function copyToClipboard (id) {
     console.error('Failed to copy text to clipboard: ', err))
 }
 
-function registerCopyToClipboardEvents () {
-  const copyPrefElem = document.getElementById('copy-preflabel')
-  if (copyPrefElem) {
-    copyPrefElem.addEventListener('click', () => copyToClipboard('concept-preflabel'))
-  }
-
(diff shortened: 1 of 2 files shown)


UnniKohonen (approved):
Seems like `dataset` and `getAttribute(…)` are essentially equivalent here so I don't think the SonarCloud comment has to be addressed. Otherwise looks good to me!

osma:
I decided to change to `.dataset` anyway since it's the current recommendation and it was only a tiny change. Merging now.

PR #2041, opened 2026-08-25
Fix navigating session history on concept pages

UnniKohonen:

Clicking on browser's back and forward buttons currently does nothing on concept pages. This PR adds functionality to move between concept pages using the buttons.

- Add event listener to partial-page-load.js that listens to popstate events originating from browser's history navigation
  - Perform a partial page load when moving to a concept page
  - Load the new page fully when moving to any other type of page
- Add event listeners for popstate events to sidebar Vue components to remove selected concept highlighting
- Add cypress tests for navigating in history

I'm not sure if `${window.SKOSMOS.vocab}/${window.SKOSMOS.lang}/page/` (used to detect if the page being opened is a concept page) will work with all concepts/vocabs possible in Skosmos.

Highlighting for selected concept in sidebar is now removed when navigating between pages. Skosmos 2 works this way as well so no regression is introduced but ideally highlight should move as well.

## Checklist

### DIFF
--- a/resource/js/partial-page-load.js
+++ b/resource/js/partial-page-load.js
@@ -89,7 +89,7 @@ const moveFocus = () => {
 
 /* eslint-disable no-unused-vars */
 const partialPageLoad = (event, pageUri) => {
-  event.preventDefault()
+  event.type != 'popstate' && event.preventDefault()
 
   // fetching html content of the concept page
   fetchWithAbort(pageUri, 'concept')
@@ -97,9 +97,9 @@ const partialPageLoad = (event, pageUri) => {
       return data.text()
     })
     .then(data => {
-      // updating url and history
-      if (window.history.pushState) { window.history.pushState({ url: pageUri }, '', pageUri) }
-
+      // updating url and history when clicking on concept links
+      if (event.type != 'popstate' && window.history.pushState) { window.history.pushState({ url: pageUri }, '', pageUri) }
+      
       // removing disabled class from hierarchy tab
       if (document.querySelector('#hierarchy > a')) {
         document.querySelector('#hierarchy').classList.remove('disabled')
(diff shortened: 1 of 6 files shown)


resource/js/partial-page-load.js
```diff
@@ -128,3 +128,17 @@ const partialPageLoad = (event, pageUri) => {
     })
 }
 /* eslint-disable no-unused-vars */
+
+// Event listener for handling browser's back and forward navigation
+window.addEventListener('popstate', (e) => {
+  // Do a partial page load when moving to a concept page, otherwise load new page fully
+  if (window.location.href.includes(`${window.SKOSMOS.vocab}/${window.SKOSMOS.lang}/page/`)) {
+    if (e.state && e.state.url) {
```
osma:
```suggestion
    if (e.state?.url) {
```

as suggested by SonarCloud

resource/js/partial-page-load.js
```diff
@@ -128,3 +128,17 @@ const partialPageLoad = (event, pageUri) => {
...
+      partialPageLoad(e, e.state.url)
+    } else {
+      partialPageLoad(e, window.location.href)
+    }
+  } else {
+    window.open(window.location.href, '_self')
```
osma: Copilot had a suggestion for this

osma (approved):
I tested this briefly. It's a clear improvement over the status quo so I'm approving it as it is, but it could be improved further either now or in follow-up PRs.

Missing the highlight is indeed one problem that could be addressed either now or in the future. Ideally the previous UI state would be completely restored when navigating history back/forward and highlighting the current concept is a part of that.

There is now some duplication in the sidebar tab code due to the new event handlers. Is there an easy way to prevent that duplication?

Also, I noticed a bug of sorts when navigating back/forward in a vocabulary where the Groups tab is set as the default. To replicate:

1. Go to Metadata vocabulary front page: http://localhost/Skosmos/mts/en/
2. Click on anything in the Groups tab, e.g. "Corporate body" (does a partial load)
3. Click the back button (this reloads the front page)
4. Click the forward button

This reopens the "Corporate body" page, but for some reason now the Hierarchy tab gets selected instead of Groups. In the Hierarchy tab, a spinner spins forever. However, in the grand scheme of things this is a pretty minor bug.

SonarCloud and Copilot both suggested minor code changes.

PR #2043, opened 2026-08-26
Fix PluginRegister callback handling when there are none or multiple callbacks per plugin

osma:

The newly updated `finaf` plugin didn't work properly. It registers two callbacks (one to handle the identifiers and another to create a widget with bibliographic titles related to the person/organization) and only one of them was called by Skosmos. I traced the bug to PluginRegister, which didn't properly handle cases when a plugin defined either zero callbacks or more than one callback. If there were >1 callbacks per plugin, it would leave just one and drop the others. It could also return numbers instead of callback function names.

Analysis from the AI agent:

>The PluginRegister::getCallbacks() method had three compounding bugs:
>
>Wrong data format: Callbacks were treated as file paths. The filterPlugins() method prefixed them with plugins/{pluginName}/, so a callback like callplugin1 became plugins/test-plugin1/callplugin1.
>
>Key collision (overwrite): The code used explode('/', $callback) and stored $split[2] (the function name) keyed by $split[1] (the plugin name). When a plugin had multiple callbacks, each one overwrote the previous — only the last one survived.
>
>Numeric index leak: The code used array_replace($this->pluginOrder, $sortedCallbacks) to maintain sort order. When a plugin had no callbacks, its numeric index (0, 1, 2...) from $this->pluginOrder leaked into the final result array instead of a function name.
>
>The fix: Pass $raw=true to skip file path prefixing, and rewrite getCallbacks() to iterate plugins in order and collect all their callbacks directly without using string splitting or associative array keying.

After this fix, both finaf callbacks are called:

image

- n/a

* fix special case handling in PluginRegister.php
* update PHPUnit tests to verify

I used an AI agent (Zoo Code with Qwen3.6-35B-A3B) to trace the bug, prepare unit tests that catch it, then implement the fix. I did my best to verify the code and I have tested that it works.

## Checklist

### DIFF
--- a/src/model/PluginRegister.php
+++ b/src/model/PluginRegister.php
@@ -74,13 +74,14 @@ class PluginRegister
     /**
      * Returns the plugin configurations found from plugin folders
      * inside the plugins folder filtered by plugin name (the folder name).
-     * @param string $type filetype e.g. 'css' or 'js'
+     * @param string $type filetype e.g. 'css', 'js', or 'callback'
      * @param array $names the plugin name strings (foldernames) in an array
+     * @param bool $raw if true, returns raw values without prefixing file paths
      * @return array
      */
-    private function filterPluginsByName($type, $names)
+    private function filterPluginsByName($type, $names, $raw = false)
     {
-        $files = $this->filterPlugins($type);
+        $files = $this->filterPlugins($type, $raw);
         foreach ($files as $plugin => $filelist) {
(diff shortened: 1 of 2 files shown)


Vainonen (approved):
All seems to be working as it should. Plugin JS and CSS files are loaded in order and plugin callbacks are made in order defined in plugin JSON files.

PR #2044, opened 2026-08-26
Fix sidebar and hierarchy bugs

UnniKohonen:

There are a few bugs in the way active/disabled classes are added to sidebar tabs and how hierarchy tab is shown on different pages:

- If opening a page for a group or an array (e.g. `/mts/fi/page/m4849` or `/koko/fi/page/?uri=http%3A%2F%2Fwww.yso.fi%2Fonto%2Fyso%2Fp6386`) hierarchy tab is seemingly left loading forever because groups and arrays are not part of the hierarchy
- If clicking to a group page through the groups tab, hierarchy is enabled and the contents don't load if the tab is clicked

This PR fixes both of these issues by fixing the way hierarchy is disabled/enabled and by adding error handling to hierarchy tab so it does not load forever if no hierarchy exists. Group tab is now always opened initially on group pages if it exists and opening a group page always disables the hierarchy tab.

- Fix the way hierarchy is enabled/disabled in `partialPageLoad`
- Remove/add hierarchy tooltip text in `partialPageLoad`
- Add error handling to hierarchy tab so that it is not left loading forever if no hierarchy exists
- Fix the way active and disabled tabs are set initially in `sidebar.twig`
- Add `isGroup` to `SKOSMOS` object
- Cypress tests

The behavior of the sidebar now matches that of Skosmos 2, i.e. on a group page, hierarchy is always inaccessible. It could, however, be useful to be able to browse the hierarchy from group pages in vocabs with `showTopConcepts` set to true (e.g. `/yso/en/page/p26556`). The current tooltip message is also not very accurate in these cases.

## Checklist

### DIFF
--- a/resource/js/partial-page-load.js
+++ b/resource/js/partial-page-load.js
@@ -117,20 +117,12 @@ const partialPageLoad = (event, pageUri) => {
         if (window.SKOSMOS.isGroup) { // Add disabled class if opening a group page
           document.querySelector('#hierarchy').classList.add('disabled')
           document.querySelector('#hierarchy > a').classList.add('disabled')
-
-          const setTooltip = () => {
+          onTranslationReady(() => {
             // Prevent a race condition
-            if (!document.querySelector('#hierarchy').classList.contains('disabled')) return
-
-            document.querySelector('#hierarchy').dataset.title = $t('hierarchy-disabled-help')
-          }
-
-          // Add tooltip text
-          if (typeof $t !== 'undefined') {
-            document.querySelector('#hierarchy').dataset.title = $t('hierarchy-disabled-help')
-          } else {
-            onTranslationReady(setTooltip)
-          }
+            if (document.querySelector('#hierarchy').classList.contains('disabled')) {
+              document.querySelector('#hierarchy').dataset.title = $t('hierarchy-disabled-help')
+            }
+          })
         } else { // Otherwise remove disabled class and tooltip text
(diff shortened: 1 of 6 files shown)


resource/js/partial-page-load.js
```diff
@@ -117,6 +112,32 @@ const partialPageLoad = (event, pageUri) => {
       updateSKOSMOS(conceptHTML)
       moveFocus()
 
+      // removing/adding disabled class from hierarchy tab
+      if (document.querySelector('#hierarchy > a')) {
+        if (window.SKOSMOS.isGroup) { // Add disabled class if opening a group page
+          document.querySelector('#hierarchy').classList.add('disabled')
+          document.querySelector('#hierarchy > a').classList.add('disabled')
+
+          const setTooltip = () => {
```
osma: I think the whole block starting here can be simplified. We can skip checking $t and just always call onTranslationReady with a callback function like setTooltip because if the service has already been initialized, it will call the callback directly anyway. So something like:

```js
onTranslationReady(function () {
  if (document.querySelector('#hierarchy').classList.contains('disabled')) {
    document.querySelector('#hierarchy').dataset.title = $t('hierarchy-disabled-help')
  }
})
```

osma (approved):
I tested it and it works nicely, great work! I suggested a little simplification for the tooltip code.

SonarCloud complains about the swallowed exceptions. I don't think that's too bad but if you can find a way to avoid that warning it would be nice.

PR #2045, opened 2026-08-26
Add visible labels for search fields and dropdowns

### LINKED ISSUE #1972: [Accessibility], WCAG 3.3.2. Input fields are missing visible labels or instructions
https://test.dev.finto.fi/en/

## Description of the problem
 The "Yhteishaku sanastoista" exbandable button opens the menus “1. Valitse sanasto“ and “2. Valitse kieli” and “3. Kirjoita hakutermi”. The visible label /instructions about the element's purpose for these three elements receiving the input are disappearing. Labels/instructions disappear when a value is selected from the menus or text is entered in the search field.

**Fix:**
All elements that receive input must have visible labels at all times, so that users know what the field's meaning was even after selecting the value, and so that the field's visible label and the accessible name still match each other (important for speech control).

Icons and images can also be used as labels or instructions, if they are descriptive enough. The magnifying glass symbol is used in search fields and is understandable if you can search for anything in the field. However, the search function in question is limited to searching for terms only, so the displayed instruction " Kirjoita hakutermi " is essential information that should be displayed in the context of the field at all times.

It would work best if a visible label is added above all input elements.

Severity: moderate
Priority: medium

Image

Chrome, Firefox, Edge

osma:

To fix accessibility issues with search fields. This PR solves the problems by adding clearly visible labels on top of the search fields indicating what they are. The layout of the field content is adjusted accordingly (e.g. left align, margins) so that everything aligns nicely.

Global search:
image

Vocabulary search:
image

- Closes #1972

- accessibility is still not perfect but I hope it's an improvement
- language selector is implemented differently in global search (real radio buttons) and vocab search (li's emulating radio buttons), not sure which one is better but I think we should use the same approach in both
  - in global search, selecting language using the keyboard shows no highlights (but this is not a regression, it was already like this)
  - in vocab search, everything seems to work, but the li's emulating radio buttons solution is more suspicious IMHO

## Checklist


resource/css/skosmos.css
```diff
@@ -397,6 +413,7 @@ header .container-fluid .row {
   border-radius: 0;
   padding: 0;
   width: 44px;
```
UnniKohonen: Width of the search button could also be set to 2.8rem to make it a square

resource/css/skosmos.css
```diff
@@ -471,6 +492,10 @@ header .container-fluid .row {
   --bs-dropdown-link-active-bg: var(--search-dropdown-selected-bg);
 }
 
+#search-wrapper #language-selector .dropdown-item {
+  padding: 0.25rem 0.75rem;
```
UnniKohonen, line 497: I noticed that the color of the dropdown items is currently set to bootstrap default, it could be set to `var(--headerbar-text)` here

resource/js/global-search.js
```diff
@@ -429,188 +432,196 @@ function startGlobalSearchApp () {
...
-                  v-model="selectedVocabs"
-                  tabindex=-1
-                  @click.stop>
-                  <span class="checkmark"></span>
-                {{ value }}
-              </label>
-            </li>
-          </ul>
+        <div class="search-field-group">
+          <label id="vocab-selector-label" class="search-field-label">{{ vocabSelectorLabel }}</label>
```
UnniKohonen: Maybe a different kind of element should be used instead of label? In the suggestions widget, I used h3 elements.

resource/js/global-search.js
```diff
@@ -429,188 +432,196 @@ function startGlobalSearchApp () {
...
-                  @keydown.left.prevent
-                  @keydown.right.prevent
-                  @click.stop
-                  v-model="selectedLanguage">
-                {{ value }}
-              </label>
-            </li>
-          </ul>
+        <div class="search-field-group">
+          <label id="content-language-label" class="search-field-label">{{ langSelectorLabel }}</label>
```
UnniKohonen: Maybe a different kind of element should be used instead of label? In the suggestions widget, I used h3 elements.

resource/js/vocab-search.js
```diff
@@ -333,76 +330,65 @@ function startVocabSearchApp () {
...
-            class="btn btn-outline-secondary dropdown-toggle"
-            data-bs-toggle="dropdown"
-            @keydown="onLangMenuKeydown"
-            aria-haspopup="true"
-            :aria-label="selectSearchLanguageAriaMessage">
-            <template v-if="languageStrings">{{ languageStrings[selectedLanguage] }}</template>
-            <i class="chevron fa-solid fa-chevron-down"></i>
-          </button>
+        <div class="search-field-group">
+          <label id="content-language-label" class="search-field-label">{{ contentLanguageMessage }}</label>
```
UnniKohonen: Maybe a different kind of element should be used instead of label? In the suggestions widget, I used h3 elements.

UnniKohonen (commented):
The layout looks good and the solution seems to fulfill the requirements set in the accessibility issue. There were a few small things I noticed in CSS that I added comments for. I also commented on the labels without targets.

resource/css/skosmos.css
osma: done

resource/css/skosmos.css
```diff
@@ -296,8 +296,8 @@ header .container-fluid .row {
   color: var(--headerbar-text);
 }
 
-#search-wrapper .dropdown-menu li label:hover, #search-wrapper .dropdown-menu li:hover {
-  color: var(--search-dropdown-hover-text);
+#search-wrapper .dropdown-menu li label:hover, #search-wrapper .dropdown-menu #li:hover {
+  color: var(--search-dropdown-hover-text) !important;
```
UnniKohonen: This is overriding the rule on line 466 and makes the options in language dropdown white on hover. Removing the `!important` should fix the issue.

resource/css/skosmos.css
UnniKohonen, line 467: This is overridden by the rule on line 299. Changing to this should fix the issue:
```css
#language-list li:hover, #language-list li:active, #language-list li:hover label, #language-list li:active label {
  background-color: var(--search-dropdown-hover-bg);
  color: var(--search-dropdown-hover-text) !important;
}
```

UnniKohonen (approved):
Hover effects of dropdowns are not working quite right, I commented changes that should fix them. Otherwise looks good to me, screen reader seems to give reasonable descriptions to the buttons.

resource/css/skosmos.css
osma: Thanks, done

resource/css/skosmos.css
osma: Thanks, done

PR #2052, opened 2026-09-05
Fix owl:versionInfo warnings with language codes

### LINKED ISSUE #2051: owl:versionInfo with a language code causes PHP warnings

No public URL is available. I identified the problem on a fresh local Skosmos v3.4 container instance.

## Description of the problem

While creating a fresh thesaurus using OWL, I noticed that adding a language code to `owl:versionInfo` causes PHP warnings.

### Steps to reproduce

1. Load a vocabulary containing:

```turtle
@prefix owl: <http://www.w3.org/2002/07/owl#> .
@prefix skos: <http://www.w3.org/2004/02/skos/core#> .

<https://example.org/thesaurus> a owl:Ontology, skos:ConceptScheme;
    skos:prefLabel "Test thesaurus"@en;
    owl:versionInfo "5"@en .
```

2. Open the vocabulary page.

The page displays:

```text
Warning: Undefined array key 0 in /var/www/html/src/model/Vocabulary.php on line 225

Deprecated: explode(): Passing null to parameter #2 ($string) of type string is deprecated in /var/www/html/src/model/Vocabulary.php on line 326
```

### Expected outcome

The version information displays without PHP warnings.

Tested with Skosmos container using v3.4 and PHP 8.3.

Image

Tracing the issue shows that Skosmos stores the literal value as its array key:

https://github.com/NatLibFi/Skosmos/blob/b0367e2596946492c1b7d82b879b08ef5812bd13/src/model/Vocabulary.php#L196

In the example I showed it would create `[5 => $val]`, but the code later assumes that it is stored as key `0`:

https://github.com/NatLibFi/Skosmos/blob/b0367e2596946492c1b7d82b879b08ef5812bd13/src/model/Vocabulary.php#L224-L226

OWL supports [literal values on annotation properties](https://www.w3.org/TR/owl2-syntax/#Literals). So while little value might be gained from a language code it still is valid OWL.

Not browser-specific. The same warnings are reproducible using curl.

akeldamas:
Using a language code with `owl:versionInfo` causes PHP warnings.

- Closes #2051

I added `array_values()` because versions like `"5"@en` were stored at `[5]`, but the code tried to read `[0]`. The change renumbers the keys without changing the values. 

I also added a test covering numeric and text versions with language codes.
  
There is one failing unit test called `ModelTest::testGetResourceFromUri` from before my changes. The test I made for this PR passes correctly so I am marking the box below as complete.

## Checklist

### DIFF
--- a/src/model/Vocabulary.php
+++ b/src/model/Vocabulary.php
@@ -222,6 +222,7 @@ class Vocabulary extends DataObject implements Modifiable
             }
         }
         if (isset($ret['owl:versionInfo'])) { // if version info available for vocabulary convert it to a more readable format
+            $ret['owl:versionInfo'] = array_values($ret['owl:versionInfo']);
             $ret['owl:versionInfo'][0] = $this->parseVersionInfo($ret['owl:versionInfo'][0]);
         }
         // remove duplicate values
--- a/tests/VocabularyTest.php
+++ b/tests/VocabularyTest.php
@@ -249,6 +249,20 @@ class VocabularyTest extends \PHPUnit\Framework\TestCase
 
     /**
      * @covers Vocabulary::getInfo
+     * @covers Vocabulary::parseVersionInfo
+     */
+    public function testGetInfoWithVersionInfoWithLanguage()
+    {
+        $vocab = $this->model->getVocabulary('test');
+        $info = $vocab->getInfo('fi');
+        $this->assertEquals([
+            new EasyRdf\Literal('5', 'fi'),
+            new EasyRdf\Literal('Version 5', 'fi'),
+        ], $info['owl:versionInfo']);
+    }
+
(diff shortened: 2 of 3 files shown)


osma:
Thank you for the issue and the pull request, much appreciated!

osma (changes requested):
I see a problem with the proposed fix and the test: the code now assumes that the `owl:versionInfo` value always has a language tag. The tests only cover these cases (the language tag is either `en` or `fi` in the test data).

In practice, `owl:versionInfo` may be a plain literal without a language tag, or it may have a datatype such as `xsd:string`, which is what the old code and test assumed.

The test should cover all these cases: plain literal, literal with datatype, literal with a language tag. The code should be able to handle all of these.

PR #2070, opened 2026-09-14
Speed up Docker builds by reordering Composer operations and Cypress install

osma:

Building the Docker image for running the tests (e.g. `docker compose up -d --build` in `tests/`) takes a very long time **every time a source file changes**. Measured timings showed that a source edit re-triggered two expensive build steps that do not actually depend on the source code:

- `npx cypress install` re-downloaded the ~200 MB Cypress binary on **every** rebuild (it was placed after `COPY .`, so any source change invalidated the layer)
- `composer install` re-ran (and re-resolved dependencies without a lockfile) on every rebuild, because the `composer-installer` stage copied the entire repository (`COPY . /app/`)

**Before → after** (development target): a rebuild after a source file change drops from **~2 minutes to ~1 second**; only the file copy and autoloader generation re-run, everything else is served from the Docker layer cache.

## How to test

1. Build Docker image e.g. run init_containers.sh
2. Change something in the code (e.g. add a line to src/index.php)
3. Build Docker image again

Step 3 should only take a few seconds. Before this PR, it took several minutes.

- follow-up fix to #1916

Changes to `dockerfiles/Dockerfile.ubuntu` only:

1. **Moved `RUN npx cypress install` before `COPY . /var/www/html/`** in the `development` stage, so the cached Cypress binary layer is not invalidated when sources change.
5. **Changed `COPY . /app/` to `COPY composer.json /app/`** in the `composer-installer` stage, so `composer install` only re-runs when `composer.json` changes.
6. **Added a `composer dump-autoload` step to both the `development` and `production` stages** (with the composer binary copied in from the installer stage). This is required by change 2: `composer.json` uses a classmap over `src/`, which is not present in the installer stage, so `composer install` is now run with `--no-autoloader` and the autoloader is generated in the final stage once the sources have been copied in. The production stage uses `--no-dev` to match its `composer install` flags.

- The `dump-autoload` steps need `--ignore-platform-req=php` because the runtime image runs PHP 8.3 while the currently resolved (unlocked) Symfony dev dependencies declare `php >= 8.4.1`. Without this flag, composer generates a `platform_check.php` that fails at runtime. This preserves exactly the behavior of the current images (where composer only ever ran inside the PHP 8.4 composer stage), but it points to a pre-existing, separate problem: **without a committed `composer.lock`, dependency resolution is non-deterministic and may drift to versions requiring a newer PHP than the runtime image provides.** Committing `composer.lock` (and switching to `npm ci`) would be a follow-up.

## Checklist

- [x] phpUnit tests pass locally with my changes *(verified via `docker compose` in `tests/`: both `development` and `production` targets build, `require 'vendor/autoload.php'` works in both images, and the generated classmap contains all `src/` classes)*
- [x] I have added tests that show that the new code works, or tests are not relevant for this PR *(build-configuration change; verified by building both targets and smoke-testing the autoloader and Cypress binary in the resulting images)*
- [x] The PR doesn't reduce accessibility of the front-end code
- [x] The PR doesn't introduce unintended code changes


Vainonen (commented):
PR seems to be working: running npx cypress install takes 0 seconds.

The mentioned uncertainties coming from  ignore-platform-req option that is needed to successfully run composer dump-autoload probably should be tackled somehow in the future.

Vainonen (approved):

PR #2071, opened 2026-09-15
fix: deprecated concepts without isReplacedBy are shown with strikethrough emphasis

### LINKED ISSUE #2049: No strike-through for deprecated concepts which are not replaced by another concept
As we have replaced Skosmos 2.18 by Skosmos 3.3 and today by Skosmos 3.4 on Staging only, I cannot provide a URL address

## Description of the problem
We have three possibilities for the New and Deprecated tab: 1) new, 2) deprecated not replaced, 3) deprecated replaced by x. There is a strike-through for deprecated concepts which have been replaced by other concepts. However, there is no strike-through for deprecated concepts which have NOT been replaced by another concept. Therefore, there is no visual difference for the user between new concepts and the deprecated/not replaced concepts. This is very confusing. Could this be fixed, please.

Image

Firefox

osma:

Deprecated concepts that have no isReplacedBy relationship were not shown with the correct strikethrough emphasis (here, "No replacement" is deprecated):

image

See #2049

- Closes #2049

## How to test

1. Spin up the Docker test containers (run init_containers.sh under tests/)
2. Go to http://localhost:9090/changes/en/
3. Open the Changes tab

After this PR, it should look like this (note that "No replacement" is now shown with a strikethrough effect):

image

* expose the `deprecated` status (boolean) of concepts in the REST API method `new`
* adjust the Changes sidebar tab so it uses striketrough for all deprecated concepts, not just those that have a replacedBy
* modify PHPUnit tests to match the new REST API behaviour (added `deprecated` field in responses)
* modify Swagger schema (add the `deprecated` field)
* add Cypress test to verify

none

## Checklist


UnniKohonen (approved):
LGTM!

PR #2074, opened 2026-09-16
Fix setting searchTerm for mobile browsers

UnniKohonen:

Vocab and global search autocomplete are not currently working correctly in some mobile browsers (at least Chrome on Android). This PR fixes the way `searchTerm` is set in the Vue components when typing in the search bar. The issue is that in some mobile browsers, the v-model directive does not update the value of the variable as the user is typing. This means that no search is made until the user clicks off the input.

- Part of #1753

- Read the value of `searchTerm` directly from the input element instead of relying on v-model to update it
- Fix race condition when clearing the search field

I have not been able to test this PR natively on a mobile device but I made a similar [change](https://github.com/NatLibFi/Skosmos-widget-suggestions/commit/62271cae339189601fd11268d2b2481e0cda808e) to the suggestion widget which is visible in https://test.dev.finto.fi/yso/en/#suggestion. The fix is working currently in e.g. the *Broader term (BT)* field.

## Checklist


osma (approved):
I tested this and couldn't make it break myself, but Copilot still complains about a potential race condition. Please check that once again and then I think this can be merged.

PR #2075, opened 2026-09-17
WIP: search component keyboard focus & other fixes

### LINKED ISSUE #1856: Use getConceptURL function when forming links to concepts

Searhch components form concept links ad hoc. Instead, this nifty global function should be tested and used.

The code would be cleaner that way.

Currently, the getConceptURL does not really work in all possible use cases, e.g. when forming a search result concept link in global search context, it would give us a link like this:
`http://localhost/fi/page/?uri=http%3A%2F%2Fwww.yso.fi%2Fonto%2Fyso%2Fp19378&clang=`

Desired result would be more like this:
`http://localhost/your-skosmos-installation-folder/fi/page/?uri=http%3A%2F%2Fwww.yso.fi%2Fonto%2Fyso%2Fp19378`

Since the content language parameter is not handled correctly, it's possible that url parameters need to be tested case-by-case, with better parameter handling and cypress tests added.

Possible use cases for this function:
* vocabulary search autocomplete dropdown list
* global search autocomplete dropdown list
* vocabulary search result page
* global search result page

Conan! What's best in life?
 * To crush all bugs
 * See them tested in cypress before you push
* And hear the sweet song of good infrastructure

### LINKED ISSUE #2056: No visible keyboard focus indicator in language selector dropdown and search results (WCAG 2.4.7)
https://test.dev.finto.fi/fi/ (yhteishaku sanastoista)
https://test.dev.finto.fi/yso/fi/ (syötä haettava termi)

## Description of the problem

In the language selector dropdown (#language-list, role="radiogroup"), keyboard focus correctly moves to the active `<li>` item (roving tabindex is implemented: the active item has a tabbable index, the others tabindex="-1", and the nested native `<input type="radio">` is intentionally set to tabindex="-1" so it does not receive Tab focus itself). However, no visible focus indicator (outline, border, highlight, etc.) is rendered on the focused item, so a sighted keyboard user cannot tell which language option currently has focus.

The active item also uses tabindex="1" instead of tabindex="0", which is a secondary issue: positive tabindex values are placed ahead of all tabindex="0" elements in the browser's tab order regardless of DOM position, which can break the page's overall logical focus order.

### Testing instructions
1. Open the page and use Tab to reach the language selector button, then open the dropdown.
2. Use Tab / Arrow keys to move focus into the list of language options (#language-list).
3. Observe that no visible indicator (outline, highlight, border) appears on the item that currently has focus.
4. In DevTools, inspect the focused `<li>` element's computed styles under the :focus state to confirm whether outline: none (or equivalent) is being applied.

Language selector issue: yhteishaku
Search results issue: in every search

Note that the same focus indicator problem can be seen also in search results lists.

Expected result after the fix: the currently focused language option has a clearly visible, sufficiently contrasted focus indicator (e.g. an outline) at all times while navigating the list with the keyboard.

### Severity
Medium/High – violates a Level AA success criterion (2.4.7), affects all keyboard-only users trying to select a language, and prevents them from knowing which option is currently focused/would be activated.

### Fix instructions
- Identify and remove (or override) the CSS rule that suppresses the default focus outline on the `<li>` (or its nested `<label>`) in #language-list, and add an explicit, sufficiently contrasting focus style.
- Ensure the outline meets non-text contrast requirements (WCAG 1.4.11, ≥ 3:1 against the adjacent background).
- Change the active item's tabindex="1" to tabindex="0" so the roving-tabindex pattern doesn't disrupt the page's natural tab order.

### Browsers
Firefox, Chrome, NVDA

Image

Image

### LINKED ISSUE #2072: Search field focus outline offset is too large
Image

This could be made smaller

osma:

There are many issues with the search fields, some of them accessibility-related, others more general. This PR fixes many of the currently known problems.

WIP

- Closes #1856 
- Closes #2072
- Closes #2056

TODO, check commit messages

* no Cypress tests for new functionality introduced (e.g. keyboard navigation for autocomplete results)
* maybe the vocab & language selection fields should be changed to listbox aria-role and/or closer to native Bootstrap Dropdown components?
* the new div for vocab title in dropdown results does not sit well with Boostrap Grid CSS classes (but hey, it works)

## Checklist


osma:
I think the Codecov complaint about only 50% coverage is pretty harmless. This PR changes only 2 lines of PHP backend code; one of them is covered by tests but the other one isn't, because it's in a WebController invoke method that anyway wasn't previously targeted by PHPUnit tests. This functionality is tested via Cypress instead.

The Copilot complaint that "no-result popups cannot be dismissed with Escape" is not completely true. They can be dismissed with Enter, which also clears the search term; this is a built-in browser behaviour for `input type=search`.

resource/js/global-search.js
```diff
@@ -436,6 +497,10 @@ function startGlobalSearchApp () {
       showAutoComplete () {
         this.showDropdown = true
         this.$forceUpdate()
+      },
+      focusFirstResult () {
```
UnniKohonen: When pressing down arrow on the search field, the result list is scrolled and the first result is hidden. To fix this, `focusFirstResult` needs to call `event.preventDefault()`.
```suggestion
      focusFirstResult (event) {
        event.preventDefault()
```

resource/js/global-search.js
```diff
@@ -535,17 +600,19 @@ function startGlobalSearchApp () {
                 v-click-outside="hideAutoComplete"
                 v-model="searchTerm"
                 @input="autoComplete($event)"
+                @keydown.down="focusFirstResult()"
```
UnniKohonen: Event is needed to prevent scrolling when pressing down arrow
```suggestion
                @keydown.down="focusFirstResult($event)"
```

resource/js/global-search.js
```diff
@@ -535,17 +600,19 @@ function startGlobalSearchApp () {
...
                 @keyup.enter="gotoSearchPage()"
                 @click="showAutoComplete()">
```
UnniKohonen: This should be `@focus` to open the autocomplete when navigating to it using keyboard as well
```suggestion
                @focus="showAutoComplete()">
```

resource/js/vocab-search.js
```diff
@@ -389,17 +444,19 @@ function startVocabSearchApp () {
               v-click-outside="hideAutoComplete"
               v-model="searchTerm"
               @input="autoComplete($event)"
+              @keyup.down="focusFirstResult()"
               @keyup.enter="gotoSearchPage()"
               @click="showAutoComplete()">
```
UnniKohonen: This should be `@focus` to open the autocomplete when navigating to it using keyboard as well
```suggestion
              @focus="showAutoComplete()">
```

resource/css/skosmos.css
UnniKohonen, line 371: This rule is adding extra empty space after the last language option. A new rule for overriding it for the radio inputs in the language list should be added.

UnniKohonen (changes requested):
This looks good! Most of the issues seem to be fixed and the changes work correctly. Screen reader output seems reasonable. There were still some issues, especially with keyboard usage.

Autocomplete results:
- It's not currently possible to open or close the autocomplete results with keyboard navigation in vocab or global search
  - Arrow up, esc and tab/shift-tab on the search field should close autocomplete results
  - Arrow down on the search field and focus moving back to the search field should open it
- When using arrow down to navigate the vocab search autocomplete list, the list does not scroll when moving focus to the last visible result (scroll only happens on the next down press). The scroll should happen when focus moves to the last result. I'm not sure what is causing the issue or how to fix it, though.
- When pressing down arrow on the search field in global search, the result list is scrolled down. I added a comment in the code for the (likely) fix
- I would expect tab/shift-tab on autocomplete results list to move focus to the search button/language selection but I'm not sure if it is actually needed

Styling:
- Focus effect is not visible in Chrome for global or vocab search autocomplete
- There was also a small older issue with global search language list that I commented to the code.

osma:
> * It's not currently possible to open or close the autocomplete results with keyboard navigation in vocab or global search      
>   * Arrow up, esc and tab/shift-tab on the search field should close autocomplete results
>   * Arrow down on the search field and focus moving back to the search field should open it

These should now be implemented!

> * When using arrow down to navigate the vocab search autocomplete list, the list does not scroll when moving focus to the last visible result (scroll only happens on the next down press). The scroll should happen when focus moves to the last result. I'm not sure what is causing the issue or how to fix it, though.

The problem is that native scrolling behaviour does not ensure that the active item is fully visible; it is enough if it is partly visible. Added some code to improve this.

> * When pressing down arrow on the search field in global search, the result list is scrolled down. I added a comment in the code for the (likely) fix

Now fixed (not quite in the suggested way).

> * I would expect tab/shift-tab on autocomplete results list to move focus to the search button/language selection but I'm not sure if it is actually needed

Agree, done.

> * Focus effect is not visible in Chrome for global or vocab search autocomplete

This was tricky, the effect had to be implemented another way. Should work now in both Firefox and Chromium.

> * There was also a small older issue with global search language list that I commented to the code.

Fixed by narrower scoping of CSS selectors (match search field by ID).

UnniKohonen (approved):
LGTM!

